PR #3, opened 2019-06-20
Makefile updates

sjalander:
Hopefully addresses all previous comments

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,6 @@ include $(JIVE_ROOT)/Makefile.sub
 
 LLVMCONFIG ?= llvm-config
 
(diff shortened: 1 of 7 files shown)


phate:
Found issues:

1. The help message for the jlc target should NOT be "Compiles the jlm compiler", but "Compiles the jlc compiler"

2. When perform "make jive" from the jlm root directory, it compiles jive without -Wall and -Wpedantic

3. In the make help text, you have the targets libjlm.a and libjlc.a listed. Would it be possible to rename them to libjlm and libjlc (without the ending)?

4. Change the help message "Compiles jlm and runs unit and c tests" to "Compiles jlm, and runs unit and C tests"

5. The *.la files in the tests directory are never removed.

That is all I found for now.

sjalander:
Those should be fixed now

PR #4, opened 2019-07-01
Added checks if submodules been initiated

sjalander:
The setup depend on Makefile.sub from the jive submodules but it
does not exist until 'make submodule' has been done so the include
needs to be guarded


sjalander:
Added submodule target

PR #12, opened 2020-10-18
jive: Bumped jive to version that use .hpp and .cpp

sjalander:
The file endings in jive has been renamed to .hpp and .cpp from
.h and .c, which required all includes in jlm to be updated and a
make target for %.c to be changed to %.cpp


phate:
You did not change the commit of the jive submodule (at least I do not see it in the commit), so how can this work with you just renaming the files?

phate:
Nevermind, my bad, it is there.

PR #13, opened 2020-10-18
make: Added release, debug and deep-clean as targets

sjalander:
release is a simple alias for jlm-release while debug calls
jlm-debug followed by check.
clean has been modified to not call jive-clean and instead
deep-clean has been added, which calls both jlm-clean and
jive-clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,12 @@ ifneq ("$(wildcard $(JIVE_ROOT)/Makefile.sub)","")
 include $(JIVE_ROOT)/Makefile.sub
 endif
 
+
+# LLVM related variables
 LLVMCONFIG ?= llvm-config
+CLANG_BIN=$(shell $(LLVMCONFIG) --bindir)
+CC=$(CLANG)
+CXX=$(CLANG_BIN)/clang++
 
 .PHONY: all
 all: jlm-release check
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -12,8 +12,11 @@ help:
 	@$(HELP_TEXT_JLM)
 	@echo ""
 	@echo "submodule              Initializes all the dependent git submodules"
-	@echo "all                    Compile jlm in debug mode, and run unit and C tests"
-	@echo "clean                  Calls clean for jive and jlm"
+	@echo "all                    Compiles jlm in release mode, and runs unit and C tests"
+	@echo "release                Alias for jlm-release"
+	@echo "debug                  Alias for jlm-debug and check"
+	@echo "clean                  Alias for jlm-clean"
+	@echo "deep-clean             Alias jlm-clean and jive-clean"
```
phate, line 19: Missing "for"

Makefile
```diff
@@ -12,8 +12,11 @@ help:
 	@$(HELP_TEXT_JLM)
 	@echo ""
 	@echo "submodule              Initializes all the dependent git submodules"
-	@echo "all                    Compile jlm in debug mode, and run unit and C tests"
-	@echo "clean                  Calls clean for jive and jlm"
+	@echo "all                    Compiles jlm in release mode, and runs unit and C tests"
```
phate, line 15: Compile jlm in release mode, and run unit and C tests

Makefile
phate: Would it be possible to call it "purge" instead of deep-clean?

sjalander:
Commit has been updated

PR #14, opened 2020-10-20
LLVM warnings

sjalander:
These changes address most warnings from clang but maybe should be solved differently. So this is to start the discussion on how the warnings should be addressed.


phate:
What clang version did you use? Clang 7?

sjalander:
It's clang-10

PR #16, opened 2020-10-24
Reduced the number of warnings to 3, all being 'unused function'

sjalander:


phate:
I used your branch as a base for also fixing the leftover issues. I pushed the changes already to master. No need to merge this any longer.

PR #22, opened 2020-12-05
Loop unrolling

sjalander:


tests/libjlm/opt/test-unroll.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include "test-operation.hpp"
 #include "test-registry.hpp"
+#include "test-types.hpp"
```
phate, line 8: Do you actually need this include? I do not see that you defined any of the types in the new test case.

tests/libjlm/opt/test-unroll.cpp
```diff
@@ -157,9 +158,9 @@ test_known_boundaries()
 		auto end = jive::create_bitconstant(graph.root(), 32, 4);
 
 		auto theta = create_theta(ult, add, init, step, end);
-//		jive::view(graph, stdout);
```
phate, line 160: This is just noise, please add the comments before the jive::view calls again.

tests/libjlm/opt/test-unroll.cpp
```diff
@@ -270,11 +271,95 @@ test_unknown_boundaries()
...
+	auto inner2_compare = jive::bitult_op::create(32, inner2_add, lvi2_end->argument());
+	auto inner2_match = jive::match(1, {{1, 1}}, 0, 2, inner2_compare);
+	inner2_theta->set_predicate(inner2_match);
+	lvi2_init->result()->divert_to(inner2_add);
+
+	
+	jive::view(graph, stdout);
+	jlm::loopunroll loopunroll(4);
+	loopunroll.run(rm, sd);
+	/* FIX ME! Need a way to check that only the inner loop gets unrolled */
```
phate, line 346: Can we please add these inserts (and remove the FIXMEs)?

libjlm/include/jlm/opt/unroll.hpp
```diff
@@ -34,6 +34,14 @@ class loopunroll final : public optimization {
 	: factor_(factor)
 	{}
 
+	/**
+	* Given a module all inner most loops (thetas) are found and unrolled if possible.
+        * All nodes in the module are traversed and if a theta is found and is the inner most theta
+        * then an attempt is made to unroll it.
+	*
+	* \param module The module to attempt loop unrolling on.
```
phate, line 42: \param module Module where the innermost loops are unrolled

libjlm/include/jlm/opt/unroll.hpp
```diff
@@ -225,6 +233,13 @@ class unrollinfo final {
 	jive::argument * idv_;
 };
 
+
+/**
+* Try to unroll the given theta.
+*
+* \param theta The theta to attempt the unrolling on.
+* \param factor The number of times to unroll the loop.
```
phate, line 247: Can we add an example here to specify what factor means? Like if the factor is 2, then the loop body is duplicated in the unrolled loop (if that is actually true).

libjlm/src/opt/unroll.cpp
```diff
@@ -484,51 +458,47 @@ unroll(jive::theta_node * otheta, size_t factor)
...
 		if (auto structnode = dynamic_cast<jive::structural_node*>(node)) {
 			for (size_t n = 0; n < structnode->nsubregions(); n++)
-				unroll(structnode->subregion(n), factor);
-
-			if (auto theta = dynamic_cast<jive::theta_node*>(node))
-				unroll(theta, factor);
+				unrolled = unroll(structnode->subregion(n), factor);
+
+			/* Try to unroll if an inner loop hasn't already been found */
+			if (!unrolled)
```
phate, line 510: Can we use {} for this if-statement here?

libjlm/src/opt/unroll.cpp
```diff
@@ -205,105 +205,145 @@ unroll_body(
 	theta->subregion()->copy(target, smap, false, false);
 }
 
+/*
+	Copy the body of the thet theta and unroll it ntimes.
```
phate, line 209: remove thet

phate: Also, you say here ntimes, but the parameter is called factor.

libjlm/src/opt/unroll.cpp
```diff
@@ -432,7 +472,6 @@ unroll_unknown_theta(const unrollinfo & ui, size_t factor)
 			smap.insert(*olv, xv);
 		}
 	}
-
```
phate, line 461: Noise, please reinsert.

sjalander: So you like meaningless whitespaces :)

tests/libjlm/opt/test-unroll.cpp
sjalander: Not sure how to check for this.
Possibly by counting the number of nodes in the thetas, but it's not really checking for correctness.
Maybe could do check that the number of nodes is greater than the number of nodes of the original body.

phate: The checking for (complete) correctness will be messy. I would not advice it. Simply counting the number of nodes is sufficient for now.

libjlm/src/opt/unroll.cpp
phate: Fair enough. Leave it as it is now. Thanks!

PR #23, opened 2021-01-20
jlc: Added the option to specify jlm-opt optimization -J

sjalander:
It's now possible to specify jlm-opt optimizations using -J.
There's also a default set of optimization specified if -J is not
used.


phate:
Looks good!

PR #25, opened 2021-01-29
Jlc opt flags

sjalander:
Must have deleted a line by mistake when adding the previous changes as to commits.
The opts.jlmopts was never set in cmdline.cpp so didn't matter if -J was specified for jlc.


phate:
Could you rebase this branch and base it on master, please. Currently, there are 3 commits in this pull request. It should only be one if I understand it correctly.

sjalander:
I thought you already had merged the two first commits.
You are too slow :)
I've squashed them into a single commit

PR #49, opened 2021-07-10
JLC: Fixed problem with extensions

sjalander:
The check for object files used the first occuring '.' in the file
thus names such as test.c.o was not detected as an object file name..
This fix instead looks for the last occuring '.' in the file name.


phate:
I already pushed a fix for this issue.

PR #55, opened 2021-07-17
test CI

phate:


phate:
Magnus incorporated the llvm-test-suite into the CI.

PR #82, opened 2021-12-01
Added gcc to CI

sjalander:


Makefile
```diff
@@ -41,6 +41,11 @@ all: jlm-release check
 .PHONY: release
 release: jlm-release
 
+.PHONY: release-gcc
+release-gcc: CXX=g++ 
+release-gcc: jlm-release
+release-gcc: check
+
```
phate, line 48: There is no need to introduce a new make target. Just use make CXX=g++ release -j2

.github/workflows/tests.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     - name: Run unit and C tests
       run: make check -j 2
 
-  ubuntu-latest:
+  ubuntu-latest-gcc:
     runs-on: ubuntu-latest
```
phate, line 38: Why did you just change the ubuntu-latest and not introduced a new CI run?

phate (changes requested):

phate:
I just merged a commit that will resolve the error for g++. Please rebase your branch with master.

.github/workflows/tests.yml
sjalander: Because I felt it wasteful to run three compilations in parallel that basically does the same thing.
It should be easy to determine if it is due to g++ or OS version that cause the issue, in case it fails.

Makefile
sjalander: This commit has been removed.

phate (approved):

PR #136, opened 2022-05-09
backend/hls: Initial commit of high-level-synthesis backend

sjalander:


libjlm/Makefile.sub
```diff
@@ -63,6 +64,26 @@ LIBJLM_SRC = \
...
+	libjlm/src/backend/hls/rvsdg2rhls/add-prints.cpp \
+	libjlm/src/backend/hls/rvsdg2rhls/add-buffers.cpp \
+	\
+	libjlm/src/backend/hls/rhls2firrtl/base-hls.cpp \
+	libjlm/src/backend/hls/rhls2firrtl/dot-hls.cpp \
+	libjlm/src/backend/hls/rhls2firrtl/verilator-harness-hls.cpp \
+	libjlm/src/backend/hls/rhls2firrtl/firrtl-hls.cpp \
+	libjlm/src/backend/hls/rhls2firrtl/mlirgen.cpp \
+	\
+	$(LLVM_VER_SRC) \
```
phate, line 86: This seems to be nowhere defined!?

libjlm/include/jlm/opt/inlining.hpp
```diff
@@ -7,6 +7,8 @@
 #define JLM_OPT_INLINE_HPP
 
 #include <jlm/opt/optimization.hpp>
+#include <jive/rvsdg.hpp>
+#include <jlm/ir/operators/lambda.hpp>
```
phate, line 11: I believe that both headers are not necessary here. Forward declaring jive::input, jive::simple_node and lambda::node should be enough

libjlm/include/jlm/ir/hls/hls.hpp
```diff
@@ -0,0 +1,560 @@
+//
+// Created by david on 7/2/21.
+//
```
phate, line 3: Could we bring the headers for all newly added files into the same format as all the other headers, please?

libjlm/Makefile.sub
sjalander: FIXED: This should have been removed.

libjlm/include/jlm/opt/inlining.hpp
sjalander: Seems like you assume you are talking to somebody that have a complete overview of the code.
If you meant that you think that jive/rvsdg.hpp is not needed, it has been removed.

libjlm/include/jlm/ir/hls/hls.hpp
sjalander: The headers have been replaced

PR #139, opened 2022-05-09
make: Improved jlm-clean

sjalander:
Removed -r from 'rm' command when only deleting files.
Now also deletes the llvmpaths.hpp that is autogenerated for libjlc.

### DIFF
--- a/Makefile.sub
+++ b/Makefile.sub
@@ -71,6 +71,7 @@ jlm-release: libjlm-release libjlc-release jlm-print-release jlm-opt-release jlc
 jlm-clean:
 	@rm -rf $(JLM_BUILD)
 	@rm -rf $(JLM_BIN)
-	@rm -rf $(JLM_ROOT)/utests.log
-	@rm -rf $(JLM_ROOT)/ctests.log
-	@rm -rf $(JLM_ROOT)/check.log
+	@rm -f $(JLM_ROOT)/utests.log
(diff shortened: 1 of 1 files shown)


Makefile.sub
```diff
@@ -71,6 +71,7 @@ jlm-release: libjlm-release libjlc-release jlm-print-release jlm-opt-release jlc
 jlm-clean:
 	@rm -rf $(JLM_BUILD)
 	@rm -rf $(JLM_BIN)
-	@rm -rf $(JLM_ROOT)/utests.log
-	@rm -rf $(JLM_ROOT)/ctests.log
-	@rm -rf $(JLM_ROOT)/check.log
+	@rm -f $(JLM_ROOT)/utests.log
+	@rm -f $(JLM_ROOT)/ctests.log
+	@rm -f $(JLM_ROOT)/check.log
+	@rm -f $(LLVMPATHSFILE)
```
phate, line 77: LLVMPATHSFILE should be clean up in libjlc/Makefile.sub, no?

sjalander: Since the compilation now happens in build, I didn't view it as necessary to call the clean for each individual component. But there are a few other files that get generates, like the test logs and the llvmpaths.hpp.
So jlm-clean, cleans everything by deleting build, bin, and all other temp files.

We could change this, so it calls all the individual parts and then simply does a rmdir build bin
Another option could be to remove the individual clean commands

Issue #190, opened 2023-01-24
Alias analyses segfaults for LLVM test suite


sjalander:
cd jlm-eval-suite/llvm-test-suite/build-aa/tools
jlc -DNDEBUG  -JAASteensgaardRegionAware   -w -Werror=date-time -MD -MT tools/CMakeFiles/fpcmp-target.dir/fpcmp.c.o -MF CMakeFiles/fpcmp-target.dir/fpcmp.c.o.d -o CMakeFiles/fpcmp-target.dir/fpcmp.c.o -c ../../llvm-test-suite.git/tools/fpcmp.c2

sjalander:
#include <stdio.h>

char *load_file(const char *path) {
  FILE *fp;

  fp = fopen(path, "rb");
}

phate:
This snippet

`
#include <stdio.h>

void
load_file(const char *path, long *size_out) 
{
  fopen(path, "rb");
}
`
compiled with

`jlc -JAASteensgaardRegionAware -c example.c`

results in a Segmentation fault.

phate:
I believe this [line](https://github.com/phate/jlm/blob/d6200c2bac76c4dfd66824d8228b2cec291ae8f7/libjlm/src/opt/alias-analyses/Steensgaard.cpp#L1032) should not invoke FindOrInsertRegisterLocation, but [LookupRegisterLocation](https://github.com/phate/jlm/blob/d6200c2bac76c4dfd66824d8228b2cec291ae8f7/libjlm/include/jlm/opt/alias-analyses/Steensgaard.hpp#L151) instead.

PR #197, opened 2023-03-12
Miscellaneous

phate:
This is a collection of commits that are "prerequisites" from the alias analysis branch. I need some of these commits merged as they would otherwise create conflicts when I fix this [bug](https://github.com/phate/jlm/issues/190).

The best to review this branch is commit by commit.


tests/TestRvsdgs.hpp
```diff
@@ -9,9 +9,9 @@
 #include <jlm/ir/operators.hpp>
 
 /**
-* \brief AliasAnalysisTest class
+* \brief RvsdgTest class
```
sjalander, line 12: The file is called TestRvsdgs, but the class is RvsdgTest.
I guess there's a logic that the file is intended as the collection of tests used to test Rvsdgs while RvsdgTest is one specific test.

sjalander (approved):

PR #200, opened 2023-03-23
Fix bug in PointsTo graph construction of Steensgaard analysis

phate:
1. Some clean up
2. Overhauls StructType
3. Fixes bug in points-to graph construction of Steensgaard analysis

Best to be reviewed commit by commit.

### DIFF
--- a/libjlm/include/jlm/ir/types.hpp
+++ b/libjlm/include/jlm/ir/types.hpp
@@ -12,6 +12,7 @@
 #include <jlm/common.hpp>
 #include <jlm/util/iterator_range.hpp>
 
+#include <memory>
 #include <vector>
 
 namespace jlm {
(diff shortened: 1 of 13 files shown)


libjlm/include/jlm/ir/types.hpp
```diff
@@ -358,80 +359,96 @@ create_varargtype()
...
-		const jive::rcddeclaration * dcl)
-	: jive::valuetype()
-	, packed_(packed)
-	, declaration_(dcl)
-	{}
+  StructType(
+    bool isPacked,
+    const jive::rcddeclaration & declaration)
+    : jive::valuetype()
+    , IsPacked_(isPacked)
```
sjalander, line 374: Is the _ at the end of the variables names necessary?
They are distinguished by the upper case at the start.

phate: Yes. The rule is to have an _ at the end of class attributes. It will otherwise collide with method  ames.

PR #203, opened 2023-03-29
Porting jlm to LLLVM 15 part 1

phate:
The biggest change when porting to LLVM 15 is the transition to [opaque pointers](https://releases.llvm.org/14.0.0/docs/OpaquePointers.html). This means that we have to get rid off the element type in the PointerType class. A prerequisite for this is to eliminate all usages of the GetElementType() method, which in turn means that we have to add types explicitly to certain operations.

This PR removes some usages of the GetElementType() method of the PointerType class. These were the low hanging fruits. The other usages will be removed in a later PR.

### DIFF
--- a/libjlm/include/jlm/ir/operators/alloca.hpp
+++ b/libjlm/include/jlm/ir/operators/alloca.hpp
@@ -23,14 +23,27 @@ public:
 	virtual
 	~alloca_op() noexcept;
 
-	inline
-	alloca_op(
-		const PointerType & atype,
-		const jive::bittype & btype,
-		size_t alignment)
-	: simple_op({btype}, {atype, {MemoryStateType::Create()}})
-	, alignment_(alignment)
-	{}
+  inline
+  alloca_op(
+    const jive::valuetype & allocatedType,
+    const jive::bittype & btype,
+    size_t alignment)
+    : simple_op({btype}, {PointerType(allocatedType), {MemoryStateType::Create()}})
+    , alignment_(alignment)
+    , AllocatedType_(allocatedType.copy())
+  {}
+
+  alloca_op(const alloca_op & other)
+    : simple_op(other)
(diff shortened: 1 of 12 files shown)


libjlm/include/jlm/ir/operators/alloca.hpp
```diff
@@ -61,38 +74,33 @@ class alloca_op final : public jive::simple_op {
...
 		if (!bt) throw jlm::error("expected bits type.");
 
-		alloca_op op(PointerType(*vt), *bt, alignment);
+		alloca_op op(allocatedType, *bt, alignment);
 		return jive::simple_node::create_normalized(size->region(), op, {size});
 	}
 
 private:
 	size_t alignment_;
+  std::unique_ptr<jive::type> AllocatedType_;
```
sjalander, line 103: Why isn't this a jive::valuetype?

libjlm/include/jlm/ir/operators/store.hpp
```diff
@@ -88,12 +88,12 @@ class StoreOperation final : public jive::simple_op {
   ~StoreOperation() noexcept override;
 
   StoreOperation(
-    const PointerType & pointerType,
+    const jive::valuetype & valueType,
```
sjalander, line 91: For the LoadOperation it was called loadedType.
Why is the more generic valueType used here?

phate: Renamed it to storedType.

libjlm/include/jlm/ir/operators/alloca.hpp
phate: The copy method used in the constructor is returning std::unique_ptr<jive::type>, and there it is not very readable to change the type to a jive::valuetype there. So I decided to rather change it in the value_type() method.

sjalander (approved):

PR #212, opened 2023-04-30
Create rvsdg library

phate:
This PR renames the libjive library to librvsdg. It moves the files to the jlm/rvsdg subfolder, adjusts the header guards, adjusts all the include files, and move the unit tests to tests/jlm/rvsdg. It does not change any namespaces or other code. This will happen in a separate PR.


jlm/rvsdg/rvsdg.hpp
```diff
@@ -0,0 +1,18 @@
+/*
```
sjalander, line 1: Why do you have this header if you never use it?

sjalander (approved):

jlm/rvsdg/rvsdg.hpp
phate: Removed.

PR #228, opened 2023-05-05
Move jhls, jlc, jlm-hls, and jlm-opt tools into tools subdirectory

phate:


sjalander (approved):
Each tool is a single file, so the subdirectories are kind of pointless. Could have everything living in tools. The only potential benefit of this structure that I can see is if there is a tool in the future that will be more complex, containing multiple files.

PR #239, opened 2023-06-01
Rework build system

caleridas:
Rework all makefiles and the build system to avoid repetition of rules. Introduce macros instead that cover the generation of rules for libraries, executables etc. Make a "regular" structure that identifies all sources, headers and tests for each library (or executable) and that could also be used to convert to a different build system, eventually.

Change how tests are operated by making multiple test binaries. This makes test execution more consistent with other testing frameworks (e.g. gtest) and may allow to transition to it, eventually.

Add a "test code coverage" computation target that uses the library macro infrastructure.


caleridas:
Note that this change of build files most likely breakes some expectations on "build targets commonly used" as well as "build configurations" (debug vs. release). So, most certainly you do not want to have it merged this way, but this is probably the best way to figure out what is missing/required.

Makefile.rules
```diff
@@ -0,0 +1,162 @@
+CXXFLAGS += -O2 -fPIC -Wall
```
phate, line 1: Several flags got lost along the way: `--std=c++17 -Wpedantic - Wextra -Wno-unused-parameter -Werror -Wfatal-errors -gdwarf-4 -DJLM_DEBUG -DJLM_ENABLE_ASSERTS`

Makefile.rules
```diff
@@ -0,0 +1,162 @@
...
+$(TESTS): % : %.la
+$(TESTS): LDFLAGS += -pthread
+
+$(RUNTESTS): run-% : %
+	$^
+
+$(VALGRINDTESTS): valgrind-% : %
+	valgrind --leak-check=full --error-exitcode=1 $^
+
+check: $(RUNTESTS)
```
phate, line 134: This compiles each test individually and immediately runs it afterwards. The test-runner executable does not seem to be used at all, even you modified it in this PR.

tests/Makefile.sub
```diff
@@ -1,58 +1,8 @@
...
-	mkdir -p ${dir $@}
-	$(CXX) -o $@ $(filter %.la, $^) $(LDFLAGS)
-
-$(patsubst %, %(JLM_BUILD)/tests/%.la, $(TESTS)): CPPFLAGS += -Itests -I$(shell $(LLVMCONFIG) --includedir)
-
-TESTLOG = true
-
-jlm-check: jlm-check-utests jlm-check-ctests
-
-jlm-check-ctests: jlc-debug jlm-opt-debug
```
phate, line 27: The C tests seem to have gotten lost completely.

phate (changes requested):
The CI needs adjustment with the new rules.

Makefile.rules
```diff
@@ -0,0 +1,162 @@
...
+
+################################################################################
+# Documentation rules
+
+docclean:
+	rm -rf doc
+
+doc: .PHONY
+	mkdir -p doc
+	doxygen doxygen.conf
```
phate, line 160: This command creates the doc folder, but then writes the documentation in the docs folder. Either the rule name, or the `OUTPUT_DIRECTORY` variable in the doxygen.conf needs to be adjusted.

phate (changes requested):
The CI needs adjustment with the new rules.

Makefile
```diff
@@ -1,58 +1,25 @@
...
-	@echo "release                Alias for jlm-release"
-	@echo "debug                  Alias for jlm-debug and check"
-	@echo "docs                   Generate doxygen documentation."
-	@echo "clean                  Alias for jlm-clean"
+LLVMCONFIG = llvm-config-15
 
-JLM_ROOT ?= .
+CPPFLAGS = -I. -Itests $(shell $(LLVMCONFIG) --cflags)
 
-include $(JLM_ROOT)/Makefile.sub
```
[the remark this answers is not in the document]
sjalander: The Makefile.sub was constructed such that it could be included from the jlm-eval-suite
https://github.com/phate/jlm-eval-suite
such that one could be in the root of jlm-eval-suite and basically perform all the compilation of jlm from there.

For example, jlm-eval-suite has the option to compile the CIRCT project and then compile jlm with the HLS backend enabled. This has been achieved by when standing in jlm-eval-suite and the build of CIRCT is found then the CIRCT_PATH is set, which cause -DCIRCT to be set when building jlm.

How can this be achieved with the new setup?

Makefile.rules
```diff
@@ -0,0 +1,162 @@
...
+	$(INSTALL) -m 755 -v $(BASENAME).so.$($(BASENAME)_VERSION) $(DESTDIR)$(libdir)/
+	$(RM) -rf $(DESTDIR)$(libdir)/$(BASENAME).so $(DESTDIR)$(libdir)/$(BASENAME).so.$($(BASENAME)_SOVERSION)
+	ln -sf $(BASENAME).so.$($(BASENAME)_VERSION) $(DESTDIR)$(libdir)/$(BASENAME).so
+	ln -sf $(BASENAME).so.$($(BASENAME)_VERSION) $(DESTDIR)$(libdir)/$(BASENAME).so.$($(BASENAME)_SOVERSION)
+
+################################################################################
+# Pseudo-rules for maintaining development build tree
+
+clean:
+	find $(BUILD_OUT_PREFIX). -name "*.[ao]" -o -name "*.l[ao]" -o -name "*.so" -o -name "*.so.*" -o -name "*.gcda" -o -name "*.gcno" | xargs rm -f
```
sjalander, line 86: Why searching for all files instead of simply deleting the build directory?

tests/test-registry.hpp
```diff
@@ -19,6 +19,9 @@ register_unit_test(const std::string & name, int(*verify)());
 int
 run_unit_test(const std::string & name);
 
+int
+run_all_unit_tests();
```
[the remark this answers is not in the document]
caleridas: Can change it, was just trying to be consistent with what is in this file at present.

Makefile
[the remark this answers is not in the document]
caleridas: then I would propose to do it this way:
- have a separate "target selection mechanism" (could be "configure"-like script, but can also have default target; could also have "make" targets that switch current config)
- use "build-<config>" directories; e.g. build-debug, build-release, build-hls-release
so you would call to set up build config, then build (and when not changing build config, every "make" will just use presently set config). Dependency tracking etc all works relative to currently set build target. (Most build systems actually handle build different targets in this way).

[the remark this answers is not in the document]
sjalander: That sounds good, with the added benefit that one can have multiple versions in parallel without having to rebuild everything.

How would we handle the final executables that currently are placed in jlm/bin?
Do we then also have multiple versions placed here, e.g., jlc-release, jlc-debug, and then a way of switching between which one is default (e.g., jlc symlink to either of them)?

Makefile.rules
sjalander: The CommandPaths specify local paths to tools that, e.g., jlc and jhls depends on, such as clang, llc, and firtool.
I was thinking that we maybe could remove it and depend on that these tools are in the user's $PATH instead.
But it's possible to have multiple versions of clang and llc installed, and I'm not sure if the version specific binary, e.g., clang-15 and llc-15, is always in the $PATH by default if only a single version of llvm is installed.

.github/workflows/tests.yml
```diff
@@ -37,19 +39,23 @@ jobs:
...
+    - name: Configure jlm
+      run: CXX=clang++-16 ./configure.sh release
     - name: Compile jlm
-      run: make jlm-release -j `nproc`
+      run: make -O -j `nproc`
     - name: Run unit and C tests
-      run: make check -j `nproc`
+      run: make check -O -j `nproc`
 
   gcc:
```
sjalander, line 49: clang used to be the default compiler, and this test was to assure that there were no errors compiling with g++.
If g++ is used as the default compiler, then this test should be changed to run with clang.

.github/workflows/hls.yml
```diff
@@ -41,10 +41,12 @@ jobs:
     # Compile MLIR version of jlm
     # --------
 
+    - name: Configure jlm
```
sjalander, line 44: jlm needs to be compiled with -DCIRCT for the HLS backend to be enabled.

tools/Makefile.sub
```diff
@@ -1,4 +1,90 @@
...
+  CIRCT = 1
+  CIRCT_INC = $(shell $(LLVMCONFIG) --includedir)
+  CIRCT_LIB = $(shell $(LLVMCONFIG) --libdir)
+
+  jlm-hls_EXTRA_CPPFLAGS += \
+        -DCIRCT=1 \
+        -I$(CIRCT_PATH)/include \
+
+  jlm-hls_LDFLAGS += \
+        -L$(CIRCT_PATH)/lib \
```
sjalander, line 60: These flags should also be needed for libhls when -DCIRCT is used.

tools/Makefile.sub
```diff
@@ -1,4 +1,90 @@
...
+	librvsdg \
+	libutil \
+
+ifdef CIRCT_PATH
+  CIRCT = 1
+  CIRCT_INC = $(shell $(LLVMCONFIG) --includedir)
+  CIRCT_LIB = $(shell $(LLVMCONFIG) --libdir)
+
+  jlm-hls_EXTRA_CPPFLAGS += \
+        -DCIRCT=1 \
```
sjalander, line 56: This doesn't seem to be applied when configuring with --circt-path

CXX=clang++-16 ./configure.sh release --circt-path $(pwd)/jlm-eval-suite/circt/local

Related to comment in .github/workflows/hls.yml

tools/Makefile.sub
```diff
@@ -1,4 +1,90 @@
...
+	librvsdg \
+	libutil \
+
+jlc_EXTRA_CPPFLAGS += -I$(shell $(LLVMCONFIG) --includedir)
+
+jlc_LDFLAGS += $(shell $(LLVMCONFIG) --libs core irReader) $(shell $(LLVMCONFIG) --ldflags) $(shell $(LLVMCONFIG) --system-libs)
+
+$(eval $(call common_executable,jlc))
+
+jlm-hls_SOURCES = \
```
sjalander, line 36: jhls and jlm-hls is only useful if the libhls has been compiled with -DCIRCT.
Currently, these tools print a statement if CIRCT wasn't enabled, notifying the user.
An alternative would be to not build them at all if CIRCT isn't enabled.

sjalander (commented):
The updates look good.

For the HLS github workflow to work, the libhls, jhls, and jlm-hls need to be compiled with -DCIRCT, which doesn't seem to be the case.

I left comments related to this in
.github/workflows/hls.yml
tools/Makfile.sub

tools/Makefile.sub
caleridas: same as other comment

tools/Makefile.sub
caleridas: same as other comment regarding test failure fixes

Makefile.rules
caleridas: changed target name

sjalander:
I've tried to run the HLS test case locally, and I get a different behavior than for the github workflow.
I build jlm with
```
CXX=clang++-16 ./configure.sh debug --circt-path $(CIRCT_INSTALL) 
make -j `nproc` -O
```
But then `jhls` fails to find the `firrtol` command
Using `jhls` with `--###` shows that the `firtool` is expected to be in `/bin`
```
/bin/firtool  -format=fir --verilog /tmp/test-return.c_1229439998//jlm_hls.fir > /tmp/test-return.c_1229439998/jlm_hls.v
```
I don't see how the github workflow actually works, as `/bin` doesn't seem to be setup to contain `firtool`.
In the old make, this was handled by the `commandpath.hpp` that was generated by the Makefile and used during compilation.
Ideally, jhls shouldn't depend on a hard-coded path to firtool. Instead, it could maybe be expected to be part of the user's PATH.



make distclean seems to delete the Makefile in the root directory.
There doesn't seem to exist a version of clean that deletes the build directories.
The current versions leave a tree with empty directories. 

Makefile.sub in the root directory has no purpose any more.

It might be useful to be able to build jlm with and without CIRCT support without having to rebuild everything.
I guess this could be done by creating specific build directories, e.g., build-release-circt if the --circt-path is given to the configure script.

sjalander:
The targets for clang format is missing
```
make format
make format-dry-run
```

sjalander:
The README.md should be updated.

jlm/llvm/Makefile.sub
```diff
@@ -59,18 +59,126 @@ LIBLLVM_SRC = \
...
+	jlm/llvm/backend/jlm2llvm/context.hpp \
 
-.PHONY: libllvm-release
-libllvm-release: CXXFLAGS += -O3
-libllvm-release: $(JLM_BUILD)/libllvm.a
+libllvm_TESTS += \
+	tests/jlm/llvm/backend/llvm/r2j/test-empty-gamma \
+	tests/jlm/llvm/backend/llvm/r2j/test-partial-gamma \
+	tests/jlm/llvm/backend/llvm/r2j/test-recursive-data \
+    tests/jlm/llvm/backend/llvm/jlm-llvm/TestAttributeConversion \
```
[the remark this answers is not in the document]
caleridas: fixed

Makefile.rules
caleridas: The test-runner is still used as the "main" program for each test. But now it is for each test file individually (and each test file may still register multiple tests inside it).

tests/Makefile.sub
caleridas: yes indeed, they are restored now

tests/test-registry.cpp
```diff
@@ -21,20 +24,43 @@ class unit_test {
 	int (*verify)();
 };
 
-static std::unordered_map<std::string, std::unique_ptr<unit_test>> unit_test_map;
+using unit_test_map_t = std::map<std::string, std::unique_ptr<unit_test>>;
```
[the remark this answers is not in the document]
caleridas: std::unordered_map has the unpleasant property that iteration order can be different between different program executions and/or on different compile targets. I would favor determinism in execution and output, I think "performance" is not of concorn with test registration...

tests/test-registry.cpp
```diff
@@ -6,8 +6,11 @@
 #include "test-registry.hpp"
 
 #include <assert.h>
```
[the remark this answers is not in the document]
caleridas: sure can change it

Makefile
```diff
@@ -1,58 +1,25 @@
 # Copyright 2014 Nico Reißmann <nico.reissmann@gmail.com>
 # See COPYING for terms of redistribution.
 
-define HELP_TEXT
```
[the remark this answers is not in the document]
caleridas: These are documented in README.md as well as the configure script now.

Makefile.rules
caleridas: Fixed cleaning up the generated file -- but I did not change behavior in relying on $PATH, leaving that for later if desired.

tests/test-registry.hpp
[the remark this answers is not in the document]
caleridas: renamed

caleridas:
Went through all comments again to verify that I have addressed them, added documentation and remaining clean-ups, rebased on latest master (without issues).

tests/Makefile.sub
```diff
@@ -1,58 +1,63 @@
...
+GENERATED_FILES+=$(patsubst %, $(BUILD_OUT_PREFIX)%.jlm, $(JLC_COMPILE_TESTS))
+GENERATED_FILES+=$(patsubst %, $(BUILD_OUT_PREFIX)%.o, $(JLC_COMPILE_TESTS))
+
+$(patsubst %, run-%, $(JLC_COMPILE_TESTS)): run-%: $(BUILD_OUT_PREFIX)%.jlm
+	$<
+
+# Run the jlc-compiled binaries as part of tests. This may
+# be problematic when cross-compiling for foreign targets,
+# but leave this to be addressed another day.
+check: $(patsubst %, run-%, $(JLC_COMPILE_TESTS))
```
phate, line 63: Is there any way to print a little summary at the end. I see that this runs the C tests and the unit tests, but I had to manually redirect the output into a file and inspect the file to see that "everything" is running. Before, we had a little message saying that all tests passed. It was very convenient.

The other "problem" now is that `make check` fails immediately if one test fails without running the rest of the tests at all. Before, it ran all tests and printed me the ones that failed, and I could choose which one to attack. I do not see this as a huge issue, but it basically gave me a choice from which "angle to attack" a problem. If this is not easily to do, then ignore it, please.

jlm/llvm/Makefile.sub
```diff
@@ -61,18 +61,126 @@ LIBLLVM_SRC = \
...
+	tests/jlm/llvm/ir/test-cfg-structure \
+	tests/jlm/llvm/ir/test-cfg-validity \
+	tests/jlm/llvm/ir/test-domtree \
+	tests/jlm/llvm/ir/test-ssa-destruction \
+	tests/jlm/llvm/ir/TestAnnotation \
+	tests/jlm/llvm/opt/alias-analyses/TestAgnosticMemoryNodeProvider \
+	tests/jlm/llvm/opt/alias-analyses/TestMemoryStateEncoder \
+	tests/jlm/llvm/opt/alias-analyses/TestRegionAwareMemoryNodeProvider \
+	tests/jlm/llvm/opt/alias-analyses/TestSteensgaard \
+	tests/jlm/llvm/opt/test-cne \
```
phate, line 168: There are tests missing. The `TestAndersen`, `TestPointerObjectSet`, and `TestPointsToGraph` tests are missing. I suspect that the list of tests is still the initial list from when you started this PR. You did not add the additional tests that were added since then to the tests list when updating the branch. Thus, there might be other tests missing as well.

phate (changes requested):
It is mostly small issues left, in my opinion. I was running out of time for reviewing now. I will have another round later or tomorrow. I am really excited to get this one in.

README.md
```diff
@@ -27,14 +46,33 @@ make docs
...
+```
+
+The tests can also be run instrumented under valgrind to validate absence
+of detectable memory errors:
+```
+make valgrind-check
+```
+
+Lastly, when the build has been configured with coverage support, then
+the following build target will compute unit test coverage for all files:
```
phate, line 62: Could you be more precise here, please? I just stumbled over the problem that I did not have it build with coverage enabled and invoked `make coverage`. The output `No rule to make target coverage` is not very helpful. Could you clarify that one would need to configure (and build) with `--enable-coverage`, please?

README.md
```diff
@@ -27,14 +46,33 @@ make docs
...
+```
+make coverage
```
phate, line 64: The code coverage creates a dependency to `gcovr`. It should be added to the dependency list in the README.md.

Makefile.rules
```diff
@@ -0,0 +1,181 @@
+CXXFLAGS += -O2 -fPIC -Wall
+
+TEST_EXECUTABLES += $(TESTS)
+COVERAGE_EXECUTABLES += $(COVERAGE_TESTS)
+
+EXECUTABLES += $(TARGET_EXECUTABLES) $(TEST_EXECUTABLES) $(COVERAGE_EXECUTABLES)
+
+GENERATED_FILES += $(EXECUTABLES)
+
+all: $(SHARED_LIBRARIES) $(STATIC_LIBRARIES) $(EXECUTABLES) $(TEST_EXECUTABLES)
```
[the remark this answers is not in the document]
caleridas: no that was an oversight, fixed

.gitignore
```diff
@@ -4,6 +4,10 @@ check.log
 /bin
 /build
 .Makefile.override
+Makefile.config
 jlm/tooling/CommandPaths.hpp
```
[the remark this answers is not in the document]
caleridas: right, now goes into build directory

.gitignore
```diff
@@ -4,6 +4,10 @@ check.log
 /bin
 /build
 .Makefile.override
```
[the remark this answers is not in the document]
caleridas: removed

.gitignore
```diff
@@ -4,6 +4,10 @@ check.log
 /bin
 /build
```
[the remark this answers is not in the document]
caleridas: removed

Makefile.rules
```diff
@@ -0,0 +1,181 @@
+CXXFLAGS += -O2 -fPIC -Wall
```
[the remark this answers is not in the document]
caleridas: that was an oversight again -- removed

tests/Makefile.sub
caleridas: You can always run "make -i check". I generally like to also parallelize unit test execution. What is going to be useful later is that unit tests should provide output which assertion failed where, such that vim and various IDEs understand it and can correlate the source line. I modified the output of the unit test running command in the makefile so you will have at least easily searchable logs telling you which unit test fails. NB that "make check" exits with an error code if any test fails, is that insufficient?

jlm/llvm/Makefile.sub
caleridas: fixed this as well. The problem is that I forgot to delete the Makefile.sub to which these were added, so I did not get a merge conflict that would highlighted addition of the tests. Fixed now, also cleaned up orphaned makefiles so any other addition is going to cause a merge conflict in case I have to rebase again.

tests/test-runner.cpp
```diff
@@ -10,6 +10,20 @@
...
+    int fail = 0;
+    for (int n = 1; n < argc; ++n)
+    {
+      if (jlm::tests::run_unit_test(argv[n]))
+      {
+        fail = 1;
+      }
+    }
+    return fail;
+  }
```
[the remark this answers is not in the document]
caleridas: Well you can still run them manually and may want to execute just a single failing test case (e.g. under gdb).

README.md
caleridas: documented the flag

README.md
caleridas: added explanation to README.md

tests/Makefile.sub
phate: The failure or success is still not as easily visible as before (I think), but good enough. The modification improves it significantly. It should be sufficient that "make check" exits with an error code.

tests/test-runner.cpp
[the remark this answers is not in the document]
phate: I still do not seem to fully understand. As it currently is, we create a single executable for each unit test file:

```
g++ -o build/tests/jlm/rvsdg/test-nodes build/tests/jlm/rvsdg/test-nodes.la   build/librvsdg.a  build/libutil.a  build/libjlmtest.a build/librvsdg.a  -pthread
```

(I just noticed that build/librvsdg.a is provided two times in the above line, but that is aside the point right now)

Such an executable is build by taking a single test file, linking this with libjlmtest.a, which contains the test-runner main() function (and some other dependencies) from above. Now, since we only have a single test file in this executable, the only test that gets registered is the one from the single test file. In the example above that would be `jlm/rvsdg/test-nodes`.

I can take this executable now and simply invoke to run the test as follows:

```
./test-nodes
```

This will go in the if case of the above if-statement and run all unit tests. Since only a single unit test file is registered as explained above, we would only run the `jlm/rvsdg/test-nodes` unit tests.

I could also go ahead and invoke the executable as follows:

```
./test-nodes jlm/rvsdg/test-nodes
```

This will go in the else case of the above if-statement and run this particular unit test. The effect is the same as invoking `./test-nodes` without argument. Since only a single unit test file is compiled into an executable, the two cases of the if-statement effectively are the same for every compiled unit test file executable. We currently do not build an executable that contains all unit test files, where it would make sense to be able to select a unit test. In other words, my point is that a selection mechanism where we can only select a single option feels redundant.

Maybe I am horribly misunderstanding something here?

phate (changes requested):
I am sorry, but I have to ask for another round. Your last revision broke something when enabling coverage:

```
./configure.sh --target debug --enable-coverage
make
```

These two commands will result in the following output:

```
Makefile.rules:45: target 'build/tests/jlm/rvsdg/test-binary.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/rvsdg/test-bottomup.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/rvsdg/test-cse.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/rvsdg/test-gamma.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/rvsdg/test-graph.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/rvsdg/test-nodes.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/rvsdg/test-statemux.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/rvsdg/test-theta.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/rvsdg/test-topdown.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/rvsdg/test-typemismatch.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/rvsdg/TestRegion.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/rvsdg/TestStructuralNode.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/util/test-disjointset.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/util/test-intrusive-hash.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/util/test-intrusive-list.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/util/TestBijectiveMap.coverage' given more than once in the same rule
Makefile.rules:45: target 'build/tests/jlm/util/TestFile.coverage' given more than once in the same rule
....
```

Moreover, the formatting in test-util.hpp is broken as well. The CI complains.

tests/test-runner.cpp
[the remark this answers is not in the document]
caleridas: There are some files that register multiple tests, e.g. tests/jlm/rvsdg/test-graph.cpp. That is also the norm in e.g. gtest setups to bundle "related" tests.

jlm/tooling/Makefile.sub
```diff
@@ -32,24 +25,50 @@ namespace jlm::tooling {\n \
...
+$(BUILD_OUT_PREFIX)jlm/tooling/Command.la: $(COMMANDPATHSFILE)
+$(BUILD_OUT_PREFIX)jlm/tooling/Command.coverage.la: $(COMMANDPATHSFILE)
+$(BUILD_OUT_PREFIX)jlm/tooling/Command.lo: $(COMMANDPATHSFILE)
+$(BUILD_OUT_PREFIX)jlm/tooling/Command.la: CPPFLAGS+=-I$(BUILD_OUT_PREFIX)
+$(BUILD_OUT_PREFIX)jlm/tooling/Command.coverage.la: CPPFLAGS+=-I$(BUILD_OUT_PREFIX)
+$(BUILD_OUT_PREFIX)jlm/tooling/Command.lo: CPPFLAGS+=-I$(BUILD_OUT_PREFIX)
 
 $(COMMANDPATHSFILE):
+	@mkdir -p $(dir $@)
+	echo creating $(COMMANDPATHSFILE)
```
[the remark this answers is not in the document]
caleridas: added missing output suppressor

Makefile.macros
```diff
@@ -0,0 +1,121 @@
...
+
+# list of test binaries within the target build directory
+$1_TESTBINARIES = $$(patsubst %, $$(BUILD_OUT_PREFIX)%, $$($1_TESTS))
+
+# add to global list of tests
+TESTS += $$(patsubst %, $$(BUILD_OUT_PREFIX)%, $$($1_TESTS))
+
+# linking of test binaries
+$$($1_TESTBINARIES): $$(patsubst %, $$(BUILD_OUT_PREFIX)%.a, $$($1_TEST_LIBS)) $$(BUILD_OUT_PREFIX)$1.a
+$$($1_TESTBINARIES): LDFLAGS+=$$(patsubst %, $$(BUILD_OUT_PREFIX)%.a, $$($1_TEST_LIBS)) $$(BUILD_OUT_PREFIX)$1.a $$($1_TEST_EXTRA_LDFLAGS)
```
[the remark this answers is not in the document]
caleridas: This is annoying to resolve. The problem is that you would "normally" want the library under test to come first, such that it can resolve everything against the additional libraries required in the test. The problem is that jlmtest depends on libjlm, so there the ordering is wrong and has to be the other way round.

The only way to resolve is to remove implicit dependency on the library-under-test and need it mentioned explicitly as testlibs, so that is what I changed now.

tests/test-runner.cpp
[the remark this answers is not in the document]
phate: I was not aware of this. Thanks for the clarification.

phate (approved):
Thank you very much for all the effort. Really appreciated!

PR #241, opened 2023-06-08
Replace optimize function

phate:
Replace optimize function with the SequentialApplication class. The SequentialApplication class fits neatly into the entire optimization architecture.


sjalander:
The class applies the provided optimizations one after each other sequentially, so that part of the name I get.
But why is it called an Application?
ApplySequentially would be a better name as far as I can tell.
More generally, it's a simple passmanager.

phate:
Well, application is the noun of apply. I would not call it a passmanager, as it does not manage anything. How about OptimizationSequence as name?

sjalander:
That's a more descriptive name.

PR #242, opened 2023-06-18
Move jlm-opt logic into JlmOptCommand class

phate:
The JlmOptCommand class represents the jlm-opt command line tool. This PR moves the entire logic of the jlm-opt command line tool from the jlm-opt.c file into the JlmOptCommand class. This permits for better testing in the future and cleans up quite a few code smells. Specifically, this PR makes the following changes:

1. It privatizes all attributes of the JlmOptCommandLineOptions class.
2. Establishes the JlmOptCommandLineOptions::OptimizationIds as the sole truth for jlm-opt optimizations.
3. Ensures that the jlm-opt optimizations performed under the hood by jlc are set in the command line parser and not later in the pipeline.
4. Moves the jlm-opt logic into the JlmOptCommand class

### DIFF
--- a/jlm/tooling/CommandGraphGenerator.cpp
+++ b/jlm/tooling/CommandGraphGenerator.cpp
@@ -114,11 +114,16 @@ JlcCommandGraphGenerator::GenerateCommandGraph(const JlcCommandLineOptions & com
 
     if (compilation.RequiresOptimization())
     {
-      auto & jlmOptCommandNode = JlmOptCommand::Create(
-        *commandGraph,
+      JlmOptCommandLineOptions jlmOptCommandLineOptions(
         CreateParserCommandOutputFile(compilation.InputFile()),
         CreateJlmOptCommandOutputFile(compilation.InputFile()),
+        JlmOptCommandLineOptions::OutputFormat::Llvm,
+        util::StatisticsCollectorSettings(),
         commandLineOptions.JlmOptOptimizations_);
+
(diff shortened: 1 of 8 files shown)


phate:
It might be best to review this PR by looking at each commit individually.

jlm/tooling/CommandGraphGenerator.cpp
```diff
@@ -114,99 +114,17 @@ JlcCommandGraphGenerator::GenerateCommandGraph(const JlcCommandLineOptions & com
...
-         * If a default optimization level has been specified (-O) and no specific jlm options
-         * have been specified (-J) then use a default set of optimizations.
-         */
-      } else if (commandLineOptions.JlmOptOptimizations_.empty()
-          && commandLineOptions.OptimizationLevel_ == JlcCommandLineOptions::OptimizationLevel::O3)
-      {
-        /*
-         * Only -O3 sets default optimizations
-         */
-        optimizations = {
```
sjalander, line 183: I realize that the definition of the OptimizationIds are not nor part of this PR, but the list of optimizations have gone from more descriptive names to the shorthand names, e.g., from DeadNodeElimination to dne

phate: Yes, the OptimizationIds of the jlm-opt command line tool have not changed, and yes, they are currently very undescriptive. However, making them more descriptive is for another PR.

PR #247, opened 2023-09-20
Simplify Steensgaard alias analysis statistics

phate:
Simplifies the Steensgaard alias analysis statistics. The PR makes the following changes:

1. Merge the alias analysis and points-to graph construction statistics into a single statistics
2. Remove the dead steensgaard points-to graph construction statistics option from jlm-opt command line interface
3. Add a simple unit test to ensure the statistics is actually gathered

### DIFF
--- a/jlm/llvm/opt/alias-analyses/Steensgaard.cpp
+++ b/jlm/llvm/opt/alias-analyses/Steensgaard.cpp
@@ -19,7 +19,7 @@
 namespace jlm::llvm::aa
 {
 
-/** \brief Steensgaard analysis statistics class
+/** \brief Collect statistics about Steensgaard alias analysis pass
  *
  */
 class Steensgaard::Statistics final : public jlm::util::Statistics {
--- a/jlm/llvm/opt/alias-analyses/Steensgaard.cpp
+++ b/jlm/llvm/opt/alias-analyses/Steensgaard.cpp
@@ -46,14 +46,14 @@ public:
   {}
 
   void
-  StartAliasAnalysisStatistics(const jlm::rvsdg::graph & graph) noexcept
+  StartSteensgaardStatistics(const jlm::rvsdg::graph & graph) noexcept
   {
     NumRvsdgNodes_ = jlm::rvsdg::nnodes(graph.root());
     AnalysisTimer_.start();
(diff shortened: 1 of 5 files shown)


jlm/llvm/opt/alias-analyses/Steensgaard.cpp
```diff
@@ -93,18 +46,31 @@ class SteensgaardPointsToGraphConstructionStatistics final : public jlm::util::S
   {}
 
   void
-  Start(const LocationSet & locationSet)
+  StartAliasAnalysisStatistics(const jlm::rvsdg::graph & graph) noexcept
```
sjalander, line 49: Would StartSteensgardsStatistics be more informative

jlm/llvm/opt/alias-analyses/Steensgaard.cpp
```diff
@@ -93,18 +46,31 @@ class SteensgaardPointsToGraphConstructionStatistics final : public jlm::util::S
...
+  {
+    NumRvsdgNodes_ = jlm::rvsdg::nnodes(graph.root());
+    AnalysisTimer_.start();
+  }
+
+  void
+  StopAliasAnalysisStatistics() noexcept
```
sjalander, line 56: Same as above

jlm/llvm/opt/alias-analyses/Steensgaard.cpp
sjalander: Or is this one step in the pipeline that all together is Steensgard?

sjalander (approved):
Do one have to start twice now if one want to collect statistics for the whole Steensgard pass?

Feels like a start() and stop() that calls Analsysis and Construction under the hood is missing.

jlm/llvm/opt/alias-analyses/Steensgaard.cpp
phate: Renamed.

jlm/llvm/opt/alias-analyses/Steensgaard.cpp
phate: Renamed.

PR #248, opened 2023-09-20
Add statistics collection support to jlc

phate:
Adds support for collecting jlm-opt statistics via jlc. Usage:

`jlc --JlmOptPassStatistics=print-aggregation-time inputFile.c`

It utilizes the same command line arguments for gathering statistics as jlm-opt.

### DIFF
--- a/jlm/tooling/CommandLine.cpp
+++ b/jlm/tooling/CommandLine.cpp
@@ -113,7 +113,7 @@ JlmOptCommandLineOptions::GetOptimizations() const noexcept
 }
 
 JlmOptCommandLineOptions::OptimizationId
-JlmOptCommandLineOptions::FromCommandLineArgument(const std::string& commandLineArgument)
+JlmOptCommandLineOptions::FromCommandLineArgumentToOptimizationId(const std::string& commandLineArgument)
 {
   static std::unordered_map<std::string, OptimizationId> map(
     {
@@ -300,7 +300,7 @@ JlcCommandLineParser::ParseCommandLineArguments(int argc, char **argv)
       JlmOptCommandLineOptions::OptimizationId optimizationId;
       try
       {
-        optimizationId = JlmOptCommandLineOptions::FromCommandLineArgument(optimization);
+        optimizationId = JlmOptCommandLineOptions::FromCommandLineArgumentToOptimizationId(optimization);
       }
       catch (util::error&)
       {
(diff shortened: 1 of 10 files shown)


jlm/tooling/CommandLine.cpp
```diff
@@ -176,6 +176,72 @@ JlmOptCommandLineOptions::ToCommandLineArgument(OptimizationId optimizationId)
...
+  if (map.find(commandLineArgument) != map.end())
+    return map[commandLineArgument];
+
+  throw util::error("Unknown command line argument: " + commandLineArgument);
+}
+
+const char*
+JlmOptCommandLineOptions::ToCommandLineArgument(jlm::util::Statistics::Id statisticsId)
+{
+  static std::unordered_map<util::Statistics::Id, const char*> map(
```
sjalander, line 215: This feels like duplicated code. Wouldn't it be possible to generate one map from the other by simply reverting the <key,value> pair:
Pseudo code:
for (key in map) {
  new_map.add(map[key].value(), key);
}

jlm/util/Statistics.hpp
```diff
@@ -66,10 +71,19 @@ class Statistics {
 /**
  * Determines the settings of a StatisticsCollector.
  */
-class StatisticsCollectorSettings final {
+class StatisticsCollectorSettings final
+{
+  inline static const char* DefaultFilePath_ = "/tmp/jlm-stats.log";
```
sjalander, line 76: This breaks when running jlc in parallel, would it be possible to add a random string to the filename?

sjalander (approved):

jlm/tooling/CommandLine.cpp
phate: You already mentioned this the last time I did this with the optimization flags. There is no way doing this without recurring a runtime overhead and losing the amortized constant lookup time. What could work is to "define both maps in one go", since it is all static. However, I consider this a minor thing since both maps are nicely encapsulated in the class and are not leaking. It is not a big issue.

jlm/util/Statistics.hpp
phate: yes, /tmp is where the resulting files ends up if nothing was specified. I do not like the working directory solution, as I would first need to find the working directory somehow, secondly it pollutes the working dir with the files (which might be unwanted by the user), and thirdly there is no automatic cleaning  up (which happens in /tmp for each reboot). I think /tmp is a reasonable default that does not interfere with the user if no one cares about the statistics.

Yes, running jlc in parallel will result in gibberish in the statistics files. I will address it in another PR.

PR #251, opened 2023-09-25
LLVM 16

sjalander:
These are the commands needed to install LLVM-16 on Ubunut Jammy Jellyfish (aka 22.04)

# Get LLVM apt key
wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc

# Update apt sources
sudo add-apt-repository deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-16 main

# Update apt database
sudo apt-get update

# Install dependencies (LLVM)
sudo apt-get install llvm-16-dev clang-16


phate:
@sjalander : I would like to merge  this, but it is out-of-date with master. Could update it, please?

.github/workflows/tests.yml
```diff
@@ -30,8 +30,12 @@ jobs:
       with:
         path: ${{ github.workspace }}/*
         key: ${{ runner.os }}-${{ github.sha }}-jlm
+    - name: Get LLVM apt key
```
phate, line 33: So you have this workflow for the HLS stuff, where you install it and then cache it such that you could reuse it. Do you think it would be possible to do the same here for the LLVM such that we can avoid in the future to change the LLVM version in a lot of places and instead have just one place?

phate: I do not expect you to do this as part of this PR.

jlm/tooling/Command.cpp
```diff
@@ -449,7 +449,7 @@ LlvmOptCommand::ToString(const Optimization & optimization)
 {
   static std::unordered_map<Optimization, const char*>
     map({
-          {Optimization::Mem2Reg, "-mem2reg"},
+          {Optimization::Mem2Reg, "-passes=mem2reg"},
```
phate, line 452: No, this is not the right place to insert the "-passes=". It should happen in the ToString() method.

.github/workflows/tests.yml
phate: Yes, it seems to be possible.

jlm/tooling/Command.cpp
sjalander: Moved to ToString().

PR #255, opened 2023-10-04
Move points-to graph unknown memory node sources redirection to Steensgaard alias analysis

phate:
This PR makes does the following:
1. Moves the redirection of the points-to graph unknown memory sources to the Steensgaard alias analysis pass. This avoids code duplication and enables the gathering of statistics.
2. Adds some documentation about this pass.
3. Cleans up the Bits2PtrTest class
4. Cleans up the Steensgaard TestBits2Ptr test

### DIFF
--- a/tests/TestRvsdgs.cpp
+++ b/tests/TestRvsdgs.cpp
@@ -334,7 +334,7 @@ Bits2PtrTest::SetupRvsdg()
   auto nf = graph->node_normal_form(typeid(jlm::rvsdg::operation));
   nf->set_mutable(false);
 
-  auto SetupBit2PtrFunction = [&]()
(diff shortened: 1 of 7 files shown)


tests/TestRvsdgs.cpp
```diff
@@ -334,7 +334,7 @@ Bits2PtrTest::SetupRvsdg()
   auto nf = graph->node_normal_form(typeid(jlm::rvsdg::operation));
   nf->set_mutable(false);
 
-  auto SetupBit2PtrFunction = [&]()
+  auto setupBit2PtrFunction = [&]()
```
sjalander, line 337: Since we had the topic about noise earlier.

sjalander (approved):

PR #260, opened 2023-10-21
Initial code structure for Andersen Alias analysis

haved:
It contains minimal optimizations for constraint solving, but tries to avoid excessive heap usage.

It might be better to move more of the struct definitions into headers, to make it possible to write unit tests for the set constraint solving itself, without an RVSDG graph in between.

There are also many statistics that it would make sense to track, such as the number of each constraint, how long solving takes, how many iterations and how long building the PtG takes.


jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+  // Mapping from functions declared with lambda nodes to PointerObject
+  std::unordered_map<jlm::llvm::lambda::node *, PointerObjectIndex> FunctionMap_;
+  // Mapping from symbols imported into the module to PointerObject
+  std::unordered_map<jlm::rvsdg::argument *, PointerObjectIndex> ImportMap_;
+
+  /**
+   * Internal helper function for adding nodes, use the Create* methods instead
+   */
+  PointerObjectIndex
+  AddNode(PointerObjectKind kind)
```
phate, line 147: `AddPointerObject()`?

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+  {
+    Nodes_.push_back(PointerObject(kind));
+    PointsToSets_.emplace_back(); // Add empty points-to-set
+    return Nodes_.size() - 1;
+  }
+
+public:
+  PointerObject &
+  GetNode(PointerObjectIndex index)
```
phate, line 156: `GetPointerObject()`?

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+    auto& P_sub = PointsToSets_[subset];
+
+    auto sizeBefore = P_super.size();
+    P_super.insert(P_sub.begin(), P_sub.end());
+
+    return P_super.size() != sizeBefore;
+  }
+
+  PointerObjectIndex
+  CreateRegisterNode(jlm::rvsdg::output *rvsdgOutput)
```
phate, line 208: `CreateRegisterNode(rvsdg::output & rsvdgOutput)`?

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+  {
+    JLM_ASSERT(RegisterMap_.count(rvsdgOutput) == 0);
+    return RegisterMap_[rvsdgOutput] = AddNode(PointerObjectKind::Register);
+  }
+
+  void
+  MapRegisterToExistingNode(jlm::rvsdg::output *rvsdgOutput, PointerObjectIndex node)
```
phate, line 215: `MapRegisterToExistingNode(rvsdg::output & rvsdgOutput, ....)`

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+  {
+    JLM_ASSERT(RegisterMap_.count(rvsdgOutput) == 0);
+    JLM_ASSERT(node < NodeCount());
+    RegisterMap_[rvsdgOutput] = node;
+  }
+
+  PointerObjectIndex
+  CreateAllocaMemoryObject(jlm::rvsdg::node *allocaNode)
```
phate, line 223: `CreateAllocaMemoryObject(rvsdg::node & allocaNode)`

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+  {
+    JLM_ASSERT(AllocaMap_.count(allocaNode) == 0);
+    return AllocaMap_[allocaNode] = AddNode(PointerObjectKind::AllocaMemoryObject);
+  }
+
+  PointerObjectIndex
+  CreateMallocMemoryObject(jlm::rvsdg::node *mallocNode)
```
phate, line 230: `CreateMallocMemoryObject(rvsdg::node & mallocNode)`

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+  {
+    JLM_ASSERT(MallocMap_.count(mallocNode) == 0);
+    return MallocMap_[mallocNode] = AddNode(PointerObjectKind::MallocMemoryObject);
+  }
+
+  PointerObjectIndex
+  CreateGlobalMemoryObject(jlm::llvm::delta::node *deltaNode)
```
phate, line 237: `CreateGlobalMemoryObject(delta::node & deltaNode)`

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+  {
+    JLM_ASSERT(GlobalMap_.count(deltaNode) == 0);
+    return GlobalMap_[deltaNode] = AddNode(PointerObjectKind::GlobalMemoryObject);
+  }
+
+  PointerObjectIndex
+  CreateFunction(jlm::llvm::lambda::node *lambdaNode)
```
phate, line 244: `CreateFunction(lambda::node & lambdaNode)`

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+  {
+    JLM_ASSERT(FunctionMap_.count(lambdaNode) == 0);
+    return FunctionMap_[lambdaNode] = AddNode(PointerObjectKind::Function);
+  }
+
+  PointerObjectIndex
+  CreateImportMemoryObject(jlm::rvsdg::argument *importNode)
```
phate, line 251: `CreateImportMemoryObject(rvsdg::argument & importNode)`

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+   * The *escaped* node is not included.
+   * Instead implicit edges through escaped+external, are added as explicit edges.
+   * @return
+   */
+  std::unique_ptr<PointsToGraph>
+  ConstructPointsToGraph()
+  {
+    auto pointsToGraph = PointsToGraph::Create();
+
+    // memory nodes are the nodes that can be pointed to in the PtG.
```
phate, line 272: `PtG` spell it out to `points-to graph`. I do not think this abbreviation is used somewhere else in the code.

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+    for (auto [lambdaNode, pointerObjectIndex] : FunctionMap_) {
+      auto& node = PointsToGraph::LambdaNode::Create(*pointsToGraph, *lambdaNode);
+      memoryNodes[pointerObjectIndex] = &node;
+    }
+    for (auto [argument, pointerObjectIndex] : ImportMap_) {
+      auto& node = PointsToGraph::ImportNode::Create(*pointsToGraph, *argument);
+      memoryNodes[pointerObjectIndex] = &node;
+    }
+
+    auto applyPointsToSet = [&](PointsToGraph::Node* node, PointerObjectIndex index) {
```
phate, line 305: `(PointsGraph::Node & node, ....)`

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+ * Example of application is a store, e.g. when *pointer1 = pointer2
+ */
+class AllPointeesPointToSupersetConstraint final {
+  PointerObjectIndex Pointer1_;
+  PointerObjectIndex Pointer2_;
+
+public:
+  AllPointeesPointToSupersetConstraint(PointerObjectIndex pointer1, PointerObjectIndex pointer2)
+          : Pointer1_(pointer1), Pointer2_(pointer2) {}
+
```
phate, line 362: Indentation

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+    return pointsToGraph;
+  }
+};
+
+/**
+ * A constraint of the form:
+ * for all x in P(pointer1), make P(x) a superset of P(pointer2)
+ * Example of application is a store, e.g. when *pointer1 = pointer2
+ */
+class AllPointeesPointToSupersetConstraint final {
```
phate, line 355: Curly braces on new line, please

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+    return modified;
+  }
+};
+
+/**
+ * A constraint of the form:
+ * P(loaded) = union of P(x) for all x in P(pointer)
+ * Example of application is a load, e.g. when loaded = *pointer
+ */
+class SupersetOfAllPointeesConstraint final {
```
phate, line 383: Curly braces on new line, please

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+  PointerObjectIndex Loaded_;
+  PointerObjectIndex Pointer_;
+
+public:
+  SupersetOfAllPointeesConstraint(PointerObjectIndex loaded, PointerObjectIndex pointer)
+          : Loaded_(loaded), Pointer_(pointer) {}
```
phate, line 389: Indentation?

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+      modified |= set.MakePointsToSetSuperset(Loaded_, x);
+    }
+
+    // TODO: external and escaped flag stuff
+
+    return modified;
+  }
+};
+
+using ConstraintVariant = std::variant<AllPointeesPointToSupersetConstraint, SupersetOfAllPointeesConstraint>;
```
phate, line 406: Move it in the PionterObjectConstraintSet class?

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+    for (PointerObjectIndex pointee : Set_.GetPointsToSet(idx)) {
+      auto& pointeeNode = Set_.GetNode(pointee);
+      modified |= !pointeeNode.HasEscaped();
+      pointeeNode.MarkAsEscaped();
+    }
+  }
+
+  return modified;
+}
```
phate, line 501: Why not defined inlined?

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+
+std::unique_ptr<PointsToGraph>
+Andersen::Analyze(const RvsdgModule &module, jlm::util::StatisticsCollector &statisticsCollector)
+{
+  Set_ = std::make_unique<PointerObjectSet>();
+  Constraints_ = std::make_unique<PointerObjectConstraintSet>(*Set_);
```
phate, line 507: Please call `reset()` on the `Set_` and `Constraints_` unique pointers before you exit the function.

jlm/llvm/opt/alias-analyses/Andersen.hpp
```diff
@@ -0,0 +1,120 @@
+#ifndef JLM_LLVM_OPT_ALIAS_ANALYSES_ANDERSEN_HPP
+#define JLM_LLVM_OPT_ALIAS_ANALYSES_ANDERSEN_HPP
+
+#include <jlm/llvm/ir/RvsdgModule.hpp>
+#include <jlm/llvm/ir/operators.hpp>
+#include <jlm/llvm/opt/alias-analyses/AliasAnalysis.hpp>
+
+namespace jlm::llvm::aa {
```
phate, line 8: Curly braces on new line, please.

jlm/llvm/opt/alias-analyses/Andersen.hpp
```diff
@@ -0,0 +1,120 @@
+#ifndef JLM_LLVM_OPT_ALIAS_ANALYSES_ANDERSEN_HPP
+#define JLM_LLVM_OPT_ALIAS_ANALYSES_ANDERSEN_HPP
+
+#include <jlm/llvm/ir/RvsdgModule.hpp>
+#include <jlm/llvm/ir/operators.hpp>
```
phate, line 5: Alphabetical order?

jlm/llvm/opt/alias-analyses/Andersen.hpp
```diff
@@ -0,0 +1,120 @@
+#ifndef JLM_LLVM_OPT_ALIAS_ANALYSES_ANDERSEN_HPP
+#define JLM_LLVM_OPT_ALIAS_ANALYSES_ANDERSEN_HPP
```
phate, line 2: The header comment is missing.

jlm/llvm/opt/alias-analyses/Andersen.hpp
```diff
@@ -0,0 +1,120 @@
...
+  Andersen &
+  operator=(const Andersen &) = delete;
+
+  Andersen &
+  operator=(Andersen &&) = delete;
+
+  std::unique_ptr<PointsToGraph>
+	Analyze(
+    const RvsdgModule & module,
+    jlm::util::StatisticsCollector & statisticsCollector) override;
```
phate, line 39: Indentation?

jlm/util/math.hpp
```diff
@@ -0,0 +1,35 @@
...
+ * Takes a sentinel end value as parameter, requires all other enum values to have a lower integer value.
+ */
+template<class T>
+static inline constexpr int BitWidthOfEnum(T sentinelEndValue) {
+  static_assert(std::is_enum<T>::value, "BitWidthOfEnum only takes enums");
+
+  using UnderlyingT = std::underlying_type_t<T>;
+  return Log2i(static_cast<UnderlyingT>(sentinelEndValue));
+}
+
```
phate, line 32: I would like to see some unit  tests on those, please.

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+   * and points-to-graph set memberships into edges.
+   *
+   * Note that registers sharing PointerObject, become separate PointsToGraph nodes.
+   *
+   * The *escaped* node is not included.
+   * Instead implicit edges through escaped+external, are added as explicit edges.
+   * @return
+   */
+  std::unique_ptr<PointsToGraph>
+  ConstructPointsToGraph()
```
phate, line 268: The points-to graph construction method as part of the `PointerObjectSet` class feels "wrong". Should it not rather be a static private method of the Andersen class and you hand the set in as parameter? This would require that you define access methods to the private attributes though....

jlm/util/math.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+#ifndef JLM_UTIL_MATH_HPP
```
phate, line 3: Two new lines and no header comment.

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -0,0 +1,514 @@
...
+ */
+class PointerObjectConstraintSet final
+{
+public:
+  explicit PointerObjectConstraintSet(PointerObjectSet& set) : Set_(set) {}
+
+  PointerObjectConstraintSet(const PointerObjectConstraintSet& other) = delete;
+
+  PointerObjectConstraintSet&
+  operator =(const PointerObjectConstraintSet& other) = delete;
```
phate, line 423: What about deleting move constructor and move assignment?

phate (commented):
Please do not misunderstand the number of comments as an indication about the code quality. The code is very readable and I like it a lot. Great work! I just went all out on it and commented on everything minor that caught my eye. Some of it is certainly a matter of personal preference.

I would move the PointerObject stuff into its own file though. This way you can easily unit test it.

haved:
Thank you for the great reviewing, your suggestions all make prefect sense, so I'll apply them, make a separate file for PointerObject stuff, and write some tests. I have also changed references to "Node" to use "PointerObject" instead.

I'm wondering what your preferred way of referring to symbols from "cousin" namespaces is. If I'm in the `jlm::llvm::aa` namespace, and want to access `jlm::llvm::delta::node`, do you prefer just `delta::node`? What about accessing `jlm::rvsdg::output`, do you include the `jlm` part? In e.g. `PointsToGraph.hpp` it seems the jlm is more often included in the second case. 

I will remove the jlm part for both, since your review indicated a preference for `rvsdg::output` etc.

The use of `std::set` was a mistake, I thought ordered sets had faster union operations, but that does not seem to be the case. If all the `P[x].insert(P[y].begin(), P[y].end())` calls end up taking a lot of time, it is probably better to implement some "dirty" tracking, and only make the calls when sets have changed. (A much cooler optimization would be to let each P-set be generational, with newly added members in a separate set until they have been new for long enough that all constraints have seen them). Then again it might be the case that 99% of sets are tiny, so I'll not implement any such things for now.

I'm a bit surprised that
```
PointerObjectKind Kind_ : jlm::util::BitWidthOfEnum(PointerObjectKind::COUNT);
```
Is a legal field, but it saves us from doing `static_cast< ... >()`, so I'm glad it is.

jlm/util/math.hpp
haved: You have a very good point, because that Log2i is absolutely not rounding up :upside_down_face:

jlm/llvm/opt/alias-analyses/Andersen.cpp
phate: I did not know that. Learned something. Thanks! Personally, I prefer to have it explicitly instead of remembering such rules.

phate:
> I'm wondering what your preferred way of referring to symbols from "cousin" namespaces is. If I'm in the jlm::llvm::aa namespace, and want to access jlm::llvm::delta::node, do you prefer just delta::node? What about accessing jlm::rvsdg::output, do you include the jlm part? In e.g. PointsToGraph.hpp it seems the jlm is more often included in the second case.

As short as possible. The `jlm::` part is mostly there because of something automatic namespace refactoring.

haved:
@phate I tried adding two new classes to `TestRVSDG.cpp` to fully test the `PointerObjectSet.cpp`. Please comment if you want anything different. Also, I didn't find a clean way of ensuring subclasses of `TestRVSDG` call `SetupModule()`, other than calling `module()` and discarding the result, so I added a method. If you have some other suggestion, please share.

tests/TestRvsdgs.hpp
```diff
@@ -18,9 +18,7 @@ class RvsdgTest {
 	jlm::llvm::RvsdgModule &
```
phate, line 18: Let's fix the indentation for the entire function, please.

tests/TestRvsdgs.hpp
```diff
@@ -30,6 +28,13 @@ class RvsdgTest {
 		return module().Rvsdg();
 	}
 
+  void
+  EnsureInitialized()
```
phate, line 32: Could we call this function `Initialize()` (or something else) and put some documentation above it that it always needs to be called to setup the test.

tests/TestRvsdgs.hpp
```diff
@@ -1807,4 +1812,146 @@ class LinkedListTest final : public RvsdgTest
   jlm::rvsdg::node * Alloca_;
 };
 
+/** \brief RVSDG module with one of each memory node type.
+ *
+ * The class sets up an RVSDG module corresponding to the code:
+ *
+ * \code{.c}
+ *   static int global = 1;
```
phate, line 1820: This is static and nowhere used in the code, and therefore the delta is not exported. This means DNE will just remove it if it is ever used on it. I am not sure that is what you want!?

tests/TestRvsdgs.hpp
```diff
@@ -1807,4 +1812,146 @@ class LinkedListTest final : public RvsdgTest
...
+  {
+    JLM_ASSERT(index < AllocaNodes_.size());
+    return *AllocaNodes_[index]->output(0);
+  }
+
+private:
+  std::unique_ptr<jlm::llvm::RvsdgModule>
+  SetupRvsdg() override;
+
+  size_t NumAllocaNodes_;
```
phate, line 1952: You might want to add a Getter for the `NumAllocaNodes_` as well.

tests/TestRvsdgs.hpp
```diff
@@ -1807,4 +1812,146 @@ class LinkedListTest final : public RvsdgTest
...
+ *     ...
+ *   }
+ * \endcode
+ *
+ * It provides getters for the alloca nodes themselves, and for their outputs.
+ */
+class NAllocaNodesTest final : public RvsdgTest
+{
+public:
+  NAllocaNodesTest(size_t numAllocaNodes) : NumAllocaNodes_(numAllocaNodes) {}
```
phate, line 1932:
```
NAllocNodesTest(size_t numAllocaNodes)
  : NumAllocaNodes_(numAllocaNodes)
{}
```
It is more readable, one piece of information per line.

phate:
> @phate I tried adding two new classes to `TestRVSDG.cpp` to fully test the `PointerObjectSet.cpp`. Please comment if you want anything different. Also, I didn't find a clean way of ensuring subclasses of `TestRVSDG` call `SetupModule()`, other than calling `module()` and discarding the result, so I added a method. If you have some other suggestion, please share.

I only looked through the TestRvsdg.hpp/*.cpp files. I sadly do not have another suggestion. I was aware of this shortcoming in the TestRvsdg class, but had no better idea either. The introduction of your method would make it the initialization at least explicit and add some documentation. I like that.

tests/TestRvsdgs.hpp
haved: This TestRVSDG subclass doesn't really represent a useful program, so I agree that any DNE pass applied to it would remove unused things (same applies to the import and alloca). I can try to make the program a bit more realistic, to perhaps make the test class more generally useful.

tests/TestRvsdgs.hpp
haved: I changed all tabs to spaces in the RvsdgTest class, but left the rest of the file alone. Is that what you meant? I hadn't noticed before that this file contains both tabs and spaces, but it seems recent code prefers spaces?

phate: Yes, we switched from tabs to spaces at some point. Now, we have a happy mix. I think it is time to get a clang format going and enforce such behavior at some point.

tests/TestRvsdgs.hpp
phate: It is fine as it is, I just wanted to point it out in  case it was not a wanted side-effect.

phate: Btw, did you make any changes after my last round of comments? If so, did you forget to push them as I see no differences?

haved:
@phate I removed the draft tag on this PR, as it should be ready to be merged as a standalone unit.
From last time you looked at it:
 - I removed the Andersen.[ch]pp files from this PR (including the PointerObjectSet-> PointsToGraph conversion code)
 - Fixed the `AllMemoryNodesTest` graph a bit, to not have any obviously dead nodes.
 - Added tests of the methods on `PointerObjectConstraintSet`.
 - Added the implication Escaped -> PointsToExternal, as we established at the meeting.

Each constraint was already tested on its own, but to test the constraint solver itself, I had to combine multiple constraints on a somewhat larger graph. The result is some long and dense testing functions, with `TestPointerObjectConstraintSetSolve()` trying to use pretty much every type of constraint.

I hope that no more constraints will have to be added to handle all of Andersen, so it should hopefully stay quite static from here on out.

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+  GetPointerObject(PointerObject::Index index);
+
+  const PointerObject &
+  GetPointerObject(PointerObject::Index index) const;
+
+  const std::unordered_set<PointerObject::Index> &
+  GetPointsToSet(PointerObject::Index idx) const;
+
+  /**
+   * Adds pointee to P(pointer)
```
phate, line 209: You have nowhere defined what `P(pointer)` is. Maybe: "Adds \p pointee to pointer object set of \p pointer"

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+   * @param pointer the index of the PointerObject that points
```
phate, line 210: "...that points" where to? The sentence is incomplete

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+   * @param pointee the index of the PointerObject that is pointed at, can not be a register.
+   * @return true if P(pointer) was changed by this operation
+   */
+  bool
+  AddToPointsToSet(PointerObject::Index pointer, PointerObject::Index pointee);
+
+  /**
+   * Makes P(superset) a superset of P(subset) by adding any elements in the set difference
```
phate, line 218: Similarly here, P(...) was nowhere defined.

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+   * @param superset the index of the PointerObject that shall point to everything subset points to
+   * @param subset the index of the PointerObject whose pointees shall all be pointed to by superset as well
+   * @return true if P(superset) was modified by this operation
+   */
+  bool
+  MakePointsToSetSuperset(PointerObject::Index superset, PointerObject::Index subset);
+
+  /**
+   * Adds the Escaped flag to all PointerObjects in the P(pointer) set
```
phate, line 227: The same here.

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+ * Example of application is when a register has multiple source values
+ */
+class SupersetConstraint final
+{
+  PointerObject::Index Superset_;
+  PointerObject::Index Subset_;
+
+public:
+  SupersetConstraint(PointerObject::Index superset, PointerObject::Index subset)
+          : Superset_(superset), Subset_(subset) {}
```
phate, line 247: Indentation, one at a time, please.

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+  bool
+  MarkAllPointeesAsEscaped(PointerObject::Index pointer);
+};
+
+/**
+ * A constraint of the form:
+ * P(superset) is a superset of P(subset)
+ * Example of application is when a register has multiple source values
+ */
+class SupersetConstraint final
```
phate, line 240: Okay, by now I would suggest to you rather introduce P(...) somewhere instead of rewriting comments.

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+{
+  PointerObject::Index Superset_;
+  PointerObject::Index Subset_;
+
+public:
+  SupersetConstraint(PointerObject::Index superset, PointerObject::Index subset)
+          : Superset_(superset), Subset_(subset) {}
+
+  bool
+  Apply(PointerObjectSet& set);
```
phate, line 250: What does this function do? Apply is a rather generic name? Some documentation might help.

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+ */
+class AllPointeesPointToSupersetConstraint final
+{
+  PointerObject::Index Pointer1_;
+  PointerObject::Index Pointer2_;
+
+public:
+  AllPointeesPointToSupersetConstraint(PointerObject::Index pointer1, PointerObject::Index pointer2)
+                                      : Pointer1_(pointer1), Pointer2_(pointer2) {}
+
```
phate, line 266: Indentation, one at a time, please.

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+  bool
+  Apply(PointerObjectSet& set);
```
phate, line 268: Same here. very generic name. Documentation?

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+ * Example of application is a load, e.g. when loaded = *pointer
+ */
+class SupersetOfAllPointeesConstraint final
+{
+  PointerObject::Index Loaded_;
+  PointerObject::Index Pointer_;
+
+public:
+  SupersetOfAllPointeesConstraint(PointerObject::Index loaded, PointerObject::Index pointer)
+                                 : Loaded_(loaded), Pointer_(pointer) {}
```
phate, line 283: Indentation, one at a time.

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+
+  bool
+  Apply(PointerObjectSet& set);
```
phate, line 286: Same here

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+};
+
+/**
+ * A class for adding and applying constraints to the points-to-sets of the PointerObjectSet.
+ * Unlike the set modification methods on PointerObjectSet, constraints can be added in any order, with the same result.
+ * Use Solve() to calculate the final points-to-sets.
+ *
+ * Constraints on the special nodes, external and escaped, are built in.
```
phate, line 294: External and Escaped are special nodes? What nodes?

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+ */
+class PointerObjectConstraintSet final
+{
+public:
+  using ConstraintVariant = std::variant<
+          SupersetConstraint,
+          AllPointeesPointToSupersetConstraint,
+          SupersetOfAllPointeesConstraint>;
+
+  explicit PointerObjectConstraintSet(PointerObjectSet& set) : Set_(set) {}
```
phate, line 304: Indentation

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+
+  /**
+   * Generic add function for all struct based constraints
+   * @param c an instance of a constraint struct, passed as a ConstraintVariant
+   */
+  void
+  AddConstraint(ConstraintVariant c);
+
+  /**
+   * Iterates over and applies constraints until the all points-to-sets satisfy them
```
phate, line 346: "...until the all..." -> "...until all..."

This function sounds "expensive". Can you say anything about its runtime performance?

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+   */
+  void
+  Solve();
+
+private:
+
+  /**
+   * Makes sure any PointerObject marked as escaped, also makes all its pointees get the HasEscaped flag set.
```
phate, line 354: "Ensures that the escaped flag is set for all pointees of any pointer object that is marked as escaped, "

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+PointerObject::Index
+PointerObjectSet::CreateRegisterPointerObject(const rvsdg::output & rvsdgOutput)
+{
+  JLM_ASSERT(RegisterMap_.count(&rvsdgOutput) == 0);
+  return RegisterMap_[&rvsdgOutput] = AddPointerObject(PointerObjectKind::Register);
+}
+
+void
+PointerObjectSet::MapRegisterToExistingPointerObject(const rvsdg::output & rvsdgOutput, PointerObject::Index pointerObject)
```
phate, line 29: each argument on new line, please

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -0,0 +1,369 @@
...
+  const std::unordered_map<const delta::node *, PointerObject::Index>&
+  GetGlobalMap() const noexcept;
+
+  const std::unordered_map<const lambda::node *, PointerObject::Index>&
+  GetFunctionMap() const noexcept;
+
+  const std::unordered_map<const rvsdg::argument *, PointerObject::Index>&
+  GetImportMap() const noexcept;
+
+  PointerObject::Index
```
phate, line 196: I would have expected a `size_t` as return type, not a `PointerObject::Index`

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -0,0 +1,303 @@
...
+{
+  JLM_ASSERT(pointer < NumPointerObjects());
+  JLM_ASSERT(pointee < NumPointerObjects());
+  // Registers can not be pointed to
+  JLM_ASSERT(GetPointerObject(pointee).GetKind() != PointerObjectKind::Register);
+
+  auto sizeBefore = PointsToSets_[pointer].size();
+  PointsToSets_[pointer].insert(pointee);
+
+  return PointsToSets_[pointer].size() != sizeBefore;
```
phate, line 147: Does the[ insert() method from std::unordered_set ](https://en.cppreference.com/w/cpp/container/unordered_set/insert) not already give you that information?

Could you not replace this code with:

```
return PointsToSets_[pioniter].insert(pointee).second
```

I am not saying it is more readable....

Maybe you want to consider to use [HashSet](https://github.com/phate/jlm/blob/master/jlm/util/HashSet.hpp)?

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -0,0 +1,303 @@
...
+SupersetOfAllPointeesConstraint::Apply(PointerObjectSet& set)
+{
+  bool modified = false;
+
+  for (PointerObject::Index x : set.GetPointsToSet(Pointer_)) {
+    modified |= set.MakePointsToSetSuperset(Loaded_, x);
+  }
+
+  // Handling pointing to external is done by MakePointsToSetSuperset,
+  // Propagating escaped status is handled by different constraints
```
phate, line 220: Thanks for the comments.

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+  return modified;
+}
+
+
+void
+PointerObjectConstraintSet::AddPointerPointeeConstraint(PointerObject::Index pointer, PointerObject::Index pointee)
```
phate, line 227: Both on new line, please.

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -0,0 +1,303 @@
...
+}
+
+void
+PointerObjectConstraintSet::AddConstraint(ConstraintVariant c)
+{
+  Constraints_.push_back(c);
+}
+
+void
+PointerObjectConstraintSet::Solve() {
```
phate, line 257: Parenthesis on new line, please.

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -0,0 +1,303 @@
...
+  // Keep applying constraints until no sets are modified
+  bool modified = true;
+
+  while (modified) {
+    modified = false;
+
+    for (auto& constraint : Constraints_)
+      std::visit([&](auto constraint) {
+        modified |= constraint.Apply(Set_);
+      }, constraint);
```
phate, line 267:
```
for (auto& constraint : Constraints_)
{
  std::visit(
    [&](auto  & constraint) { modified |= constraint.Apply(Set_); },
    constraint);
}
```
Please mind the `auto & constraint` instead of the `auto constraint` from the original code. None of those constraints has a deleted copy constructor. So not using an `&` should copy this stuff happily around, no? Is this what you want here?

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+    modified |= PropagateEscapedFlag();
+  }
+}
+
+bool
+PointerObjectConstraintSet::PropagateEscapedFlag()
+{
+  bool modified = false;
```
phate, line 276: Could we move this line of code between the while and for loop. It is not used in the first part of the code, i.e., the while loop.

tests/jlm/llvm/opt/alias-analyses/TestPointerObjectSet.cpp
```diff
@@ -0,0 +1,516 @@
...
+  TestSupersetOfAllPointeesConstraint();
+  TestAddPointsToExternalConstraint();
+  TestAddRegisterContentEscapedConstraint();
+  TestPointerObjectConstraintSetSolve();
+  return 0;
+}
+
+JLM_UNIT_TEST_REGISTER(
+"jlm/llvm/opt/alias-analyses/TestPointerObjectSet",
+TestPointerObjectSet)
```
phate, line 516: Indent, please.

tests/jlm/llvm/opt/alias-analyses/TestPointerObjectSet.cpp
```diff
@@ -0,0 +1,516 @@
...
+   * Line in the test corresponds to lines of pseudo-SSA, with branching removed.
+   * First, the function definition and local variables.
+   *
+   * function f(%0):
+   * %1 = alloca 8 (variable v1)
+   * %2 = alloca 8 (variable v2)
+   * %3 = alloca 8 (variable v3)
+   * %4 = alloca 8 (variable v4)
+   */
+
```
phate, line 385: Could we use `//` instead of `/* ... */`

phate: I do not fully understand what this comment means?

phate (changes requested):
This was a pleasure to read. Thanks a lot. Great work!

I am really looking forward to the rest of the analysis code. I am so excited to get this in jlm. 

I went all out on commenting again. Most of it is absolutely minor stuff. I would still like for you to fix at least the 1-2 more "serious" comments.

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
phate: Meh, after reading it a second time, it obviously applies the constraint. Maybe still some documentation that highlights this for sleepy readers....

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
haved: It most definitely is expensive, as in O(n^3). In practice it should be a bit better if the constraint order is mostly the same as data flow order.

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
haved: `std::unordered_set::insert()` absolutely does, I will use that.

I can use HashSet, I didn't know in what ways it differed from unordered_set, so I went with what I already (kind of) knew.

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
haved: I didn't intend to use a reference only for the loop and not the call. I know most constraints are just two 64-byte words, so they should fit in registers, but the call to Apply passes `this` as a pointer, so it doesn't matter. If we are lucky, the entire call to Apply gets inlined.

haved:
I think I fixed all your review comments, I'm compiling my "responses" here:

I can replace `std::unordered_set` with `HashSet`, if that is preferred. At the moment I have no idea when you would prefer one or the other. Do they have different features, or different performance or memory characteristics?

`PointerObjectConstraintSet::Solve()` can absolutely be slow. In the literature, Andersen is frequently referred to as a O(n^3) algorithm. The current solver is not at all trying to be clever, just blindly applying constraints, which can lead to the following performance pitfalls:
  - A chain of superset constraints of length N, if added in the opposite order of the "set membership flow", will need N full iterations though all the constraints, to pass a set member all the way through.
  - When applying a superset constraint, all members in the subset M will be iterated over and attempted inserted, even if they are already in the superset N. This gives the subset constraint a Theta(|M| log |N|) runtime, even when no actual "progress" is being made.

I have some thoughts about improving the performance, such as creating a graph of constraints and trying to *mostly* topologically sort it, and keeping a set membership history to only process set members that have been added since the last time the constraint was processed. I have also found several papers discussing optimizations, which should help. Before any of that, though, I want to get the baseline in.

Finally, regarding the use of `[&](auto & constraint)`, I think passing by reference makes the most sense, since the Apply call takes `this` by pointer. We can also hope everything gets inlined through the `std::visit`, in which case it doesn't matter.

phate:
> I can replace std::unordered_set with HashSet, if that is preferred. At the moment I have no idea when you would prefer one or the other. Do they have different features, or different performance or memory characteristics?

I prefer HashSet. HashSet is nothing else than a wrapper around std::unordered_set with a sane interface. I got tired of sometimes deciphering the interface from std::unordered_set. It is not very readable. Best example is the change I suggested yesterday: `....Insert(...).second`.

> PointerObjectConstraintSet::Solve() can absolutely be slow. In the literature, Andersen is frequently referred to as a O(n^3) algorithm.

I was aware of the O(n^3) runtime. The point of my comment was to maybe add this piece of information to the documentation. I should have written it more clearly.

> I have some thoughts about improving the performance, such as creating a graph of constraints and trying to mostly topologically sort it, and keeping a set membership history to only process set members that have been added since the last time the constraint was processed. I have also found several papers discussing optimizations, which should help. Before any of that, though, I want to get the baseline in.

You just gave a TLDR of your master thesis. The seven papers I sent you try to do exactly this: decrease the constant of this algorithm such it becomes usable for practical programs.

> Before any of that, though, I want to get the baseline in.

Yes, please. This is the most important part now.

> Finally, regarding the use of [&](auto & constraint), I think passing by reference makes the most sense, since the Apply call takes this by pointer. We can also hope everything gets inlined through the std::visit, in which case it doesn't matter.

I see no benefit in passing by value here. Regarding inlining: So I had to look this up, but as I understand it, the compiler will most likely inline std::visit, which under the hood will be nothing else than a switch statement on the underlying type(s), and then depending on the type, the right `Apply` will be called, which might be inlined as well. So if I understand this correctly, then this is nothing else than a dispatch table that is created for a virtual method. Is this correct? Why did you go this way instead of using a class with an abstract `Apply` method?

Okay, after writing the above paragraph, I googled a little bit to get some of my questions answered. It is a poor man's sum types. My next follow up question: Do we expect some other kind of constraints to pop up at some point (such that we have to change all the visitors)?

phate (approved):

haved:
I hadn't realized it was a nicer interface around unordered_set, I will switch to it.

I will add documentation about the runtime of Solve().

I don't imagine any other constraints will need to be added, based on my current Andersen class, and even if it is needed, it should not affect any other code, since there is only one std::visit, and no way of extracting constraints from the ConstraintSet.

I will make these changes, rebase to get the clang-format, then I'll be very happy to get this merged.

haved:
I did what I described above, feel free to merge once the tests have passed. I only applied clang-format to files added by this PR.

I also added a bool return to HashSet's `UnionWith` returning true if modified, hope that's OK.

Feel free to merge once the tests have passed. I'm not sure what's going on with hls-test-suite at the moment, but it seems to affect the master branch as well.

PR #262, opened 2023-10-28
Add missing forward declarations and headers to jlm/rvsdg/tracker.hpp

phate:
The PR title says it all.

### DIFF
--- a/jlm/rvsdg/traverser.hpp
+++ b/jlm/rvsdg/traverser.hpp
@@ -7,11 +7,18 @@
 #ifndef JLM_RVSDG_TRAVERSER_HPP
 #define JLM_RVSDG_TRAVERSER_HPP
 
+#include <jlm/rvsdg/tracker.hpp>
+
 #include <stdbool.h>
 #include <stdlib.h>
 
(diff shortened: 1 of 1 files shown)


jlm/rvsdg/traverser.hpp
```diff
@@ -7,11 +7,18 @@
 #ifndef JLM_RVSDG_TRAVERSER_HPP
 #define JLM_RVSDG_TRAVERSER_HPP
 
+#include <jlm/rvsdg/tracker.hpp>
```
caleridas, line 10: this is strange -- the file now includes itself?

caleridas (changes requested):

jlm/rvsdg/traverser.hpp
phate: The file is called `jlm/rvsdg/traverser.hpp` and the file I include is called `jlm/rvsdg/tracker.hpp`. It is not the same file.

caleridas: sorry yes you are right I misread!

caleridas (approved):

PR #263, opened 2023-10-28
Add call summary computation

phate:
This PR adds a call summary computation for lambda nodes. A call summary summarizes the call usages of a lambda. It currently distinguishes between three different usages:
1. The export of the lambda
2. The direct calls of the lambda
3. All other usages

The call summary computation supersedes the direct_calls() method of the lambda node.


jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -663,7 +669,197 @@ class node::fctresconstiterator final : public jlm::rvsdg::input::constiterator<
...
+  CallSummary(
+    rvsdg::result * rvsdgExport,
+    std::vector<CallNode*> directCalls,
+    std::vector<rvsdg::simple_input*> otherUsers)
+    : RvsdgExport_(rvsdgExport)
+    , DirectCalls_(std::move(directCalls))
+    , OtherUsers_(std::move(otherUsers))
+  {}
+
+  CallSummary(const CallSummary&) = delete;
```
caleridas, line 694: copy & move of the class would be "safe", so it would not be necessary to be so strict here

jlm/llvm/opt/inlining.cpp
```diff
@@ -142,22 +142,21 @@ inlineCall(jlm::rvsdg::simple_node * call, const lambda::node * lambda)
...
-		bool onlyDirectCalls = lambda->direct_calls(&calls);
-
-		if (onlyDirectCalls && calls.size() == 1)
-			inlineCall(calls[0], lambda);
-	}
+  for (auto node : rvsdg::topdown_traverser(rvsdg.root()))
+  {
+    if (auto lambda = dynamic_cast<const lambda::node*>(node))
+    {
+      auto callSummary = lambda->ComputeCallSummary();
```
caleridas, line 151: this is already traversing the graph, and "ComputeCallSummary" performs some inner traversal. I guess it "could" become quadratic for pathological cases of deeply nested lambdas, but probably it is not worth doing anything about this.

caleridas (approved):

jlm/llvm/ir/operators/lambda.hpp
phate: Removed explicit deletion.

jlm/llvm/opt/inlining.cpp
phate: Good point, but I also agree with your assessment that it is not worth doing something about it (at least so far).

PR #264, opened 2023-10-28
Add Analyze() method to Steensgaard class

phate:
Add second Analyze() method to Steensgaard class for convenience. It does not take a StatisticsCollector parameter.


phate:
@sjalander @caleridas @haved : Could I get a review, please?

haved (approved):
Looks good to me.

I guess in theory this function could be implemented as part of the AliasAnalysis base class, giving all subclasses the option for ignoring statistics, but there is currently only one subclass, so that can be changed later if desired.

PR #266, opened 2023-10-28
Add IntersectWith() method to util::HashSet class

phate:
The PR title says it all.

### DIFF
--- a/jlm/util/HashSet.hpp
+++ b/jlm/util/HashSet.hpp
@@ -222,6 +222,22 @@ public:
   }
 
   /**
+   * Modifies this HashSet object to contain only elements that are present in itself and \p other.
+   *
+   * @param other A HashSet to intersect with.
+   */
+  void
+  IntersectWith(const HashSet<ItemType> & other)
+  {
+    auto isContained = [&](const ItemType item)
+    {
+      return !other.Contains(item);
+    };
+
+    RemoveWhere(isContained);
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


jlm/util/HashSet.hpp
```diff
@@ -221,6 +221,22 @@ class HashSet
...
+   */
+  void
+  IntersectWith(const HashSet<ItemType> & other)
+  {
+    auto isContained = [&](const ItemType item)
+    {
+      return !other.Contains(item);
+    };
+
+    RemoveWhere(isContained);
```
caleridas, line 237: This will get converted to a functional object, with implicit memory allocation. You could change RemoveWhere to use template parameter (and const& qualify the parameter), and then this works without malloc.

caleridas (approved):
LGTM

jlm/util/HashSet.hpp
phate: I am not sure that I fully follow about the RemoveWhere part of your comment. Could you elaborate, please?

caleridas: Proposal:

`template<typename F>
size_t
RemoveWhere(const F& match)
{
  ...
}`

and then if you pass a lambda type it will just use it "as-is" without converting it to a std::function

phate: Addressed here: https://github.com/phate/jlm/pull/269

PR #269, opened 2023-10-31
Change signature of RemoveWhere method in HashSet class

phate:
Avoid usage of std::function<> as it leads to an implicit dynamic allocation.


phate:
@caleridas : Could I get a review, please?

haved (approved):

PR #270, opened 2023-11-04
Move RegionAwareMemoryNodeProvider class constructor to cpp file

phate:
The PR title says it all.


haved (approved):

haved:
Is there a reason you prefer to declare the default definition in the cpp file?

phate:
No preference. It was necessary in this case. The Context class is only defined in the cpp file and when an instance of the RegionAwareMemoryNodeProvider class is created using the constructor, the size of Context needs to be known for unique_ptr. The only way for it to know the size is to move the implementation of the constructor in the cpp file.

PR #272, opened 2023-11-04
Add MemoryNodeEliminator interface

phate:
The interface for future memory node eliminators. This is going to be used by the upcoming lifetime memory node eliminator.


haved (approved):
Looks good to me. I'm not entirely sure what kind of functionality an implementation would provide, but the interface makes sense.

phate:
The "eliminating" refers to the elimination of the memory nodes from the MemoryNodeProvisioning for regions where they are not defined yet. It takes a seed provisioning, removes the unnecessary memory nodes from it, and returns a new provisioning with less memory nodes (and ultimately less states that need to be routed through regions). I will soon create a PR with the first subclass using this interface.

PR #273, opened 2023-11-04
Add methods for removing multiple inputs and outputs

phate:
The current interface for removing inputs/outputs from nodes/regions has the following downsides:

1. It exposes the way the inputs/outputs are removed to the caller. The caller needs to know how to iterate through the inputs/outputs while removing (some of) them in order to avoid the invalidation of indices:

```
// iterate backwards to avoid the invalidation of 'n' by RemoveOutput()
for (size_t n = noutputs()-1; n != static_cast<size_t>(-1); n--)
{
  RemoveOutput(n);
}
```
This is not great for a public interface.

2. The sweep phase of dead node elimination experiences a performance issue when removing multiple inputs/outputs from structural nodes. The removal of an output/input requires the adjustment of the index of (potentially) all other inputs/outputs. In other words, every time we remove a single input/output, we have (in the worst case) to touch all other inputs/outputs to adjust their index. This leads to noticeable quadratic behavior in the sweep phase:

```
// iterate backwards to avoid the invalidation of 'n' by RemoveOutput()
for (size_t n = noutputs()-1; n != static_cast<size_t>(-1); n--)
{
 auto & output = ...
  if (Context_.IsDead(output))
  {
    RemoveOutput(n); // Needs to iterate through all outputs with index > n to adjust their indices
  }
}
```

This PR is the first step towards fixing these two issues. It makes the following contributions:

1. It renames the `remove_input()`, `remove_output()`, `remove_argument()`, and `remove_result()` methods such that they align with the naming convention.
2. It adds documentation for the above methods.
3. It introduces `RemoveOutputsWhere()`, `RemoveInputsWhere()`, `RemoveArgumentsWhere()`, and `RemoveResultsWhere()` methods. These methods hide the required way of iterating through the inputs/outputs from the public.

Future PRs will then replace the usages of `RemoveInput()`, `RemoveOutput()`, `RemoveArgument()`, and `RemoveResult()` methods with usages of `RemoveInputsWhere()`, `RemoveOutputsWhere()`, `RemoveArgumentsWhere()`, and `RemoveResultsWhere()`, respectively. This will enable us to privatize the `RemoveInput()`, `RemoveOutput()`, `RemoveArgument()`, and `RemoveResult()` methods such that they cannot be used publicly any longer. Once all usage sides rely on the `RemoveInputsWhere()`, `RemoveOutputsWhere()`, `RemoveArgumentsWhere()`, and `RemoveResultsWhere()` methods, we can start to efficiently implement the removal of multiple inputs/outputs internally in the respective classes.

### DIFF
--- a/jlm/rvsdg/gamma.hpp
+++ b/jlm/rvsdg/gamma.hpp
@@ -287,13 +287,14 @@ public:
 	add_exitvar(const std::vector<jlm::rvsdg::output*> & values);
 
   /**
-   * Removes all outputs that have no users and match the condition specified by \p match.
+   * Removes all gamma outputs and their respective results. The outputs must have no users and match the condition
+   * specified by \p match.
    *
    * @tparam F A type that supports the function call operator: bool operator(const gamma_output&)
    * @param match Defines the condition of the elements to remove.
    */
   template<typename F> void
-  RemoveOutputsWhere(const F& match);
+  RemoveGammaOutputsWhere(const F& match);
 
   /**
    * Removes all outputs that have no users.
@@ -306,7 +307,7 @@ public:
       return true;
     };
 
-    RemoveOutputsWhere(match);
+    RemoveGammaOutputsWhere(match);
   }
 
 	virtual jlm::rvsdg::gamma_node *
@@ -508,7 +509,7 @@ gamma_node::add_exitvar(const std::vector<jlm::rvsdg::output*> & values)
 }
 
 template<typename F> void
-gamma_node::RemoveOutputsWhere(const F& match)
+gamma_node::RemoveGammaOutputsWhere(const F& match)
 {
   // iterate backwards to avoid the invalidation of 'n' by RemoveOutput()
   for (size_t n = noutputs()-1; n != static_cast<size_t>(-1); n--)
--- a/jlm/rvsdg/gamma.hpp
+++ b/jlm/rvsdg/gamma.hpp
@@ -518,7 +518,7 @@ gamma_node::RemoveOutputsWhere(const F& match)
     {
       for (size_t r = 0; r < nsubregions(); r++)
       {
-        subregion(r)->remove_result(n);
+        subregion(r)->RemoveResult(n);
       }
 
       RemoveOutput(n);
--- a/jlm/rvsdg/gamma.hpp
+++ b/jlm/rvsdg/gamma.hpp
@@ -295,6 +295,20 @@ public:
   template<typename F> void
   RemoveOutputsWhere(const F& match);
 
+  /**
+   * Removes all outputs that have no users.
+   */
+  void
+  PruneOutputs()
+  {
+    auto match = [](const gamma_output&)
+    {
+      return true;
+    };
+
+    RemoveOutputsWhere(match);
+  }
+
 	virtual jlm::rvsdg::gamma_node *
 	copy(jlm::rvsdg::region * region, jlm::rvsdg::substitution_map & smap) const override;
 };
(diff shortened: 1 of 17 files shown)


jlm/rvsdg/region.hpp
```diff
@@ -203,8 +203,43 @@ class region {
 	void
 	append_argument(jlm::rvsdg::argument * argument);
 
-	void
-	remove_argument(size_t index);
+  /**
+   * Removes an argument from the region given a arguments' index.
```
sjalander, line 207: an arguments

jlm/rvsdg/gamma.hpp
```diff
@@ -286,6 +286,29 @@ class gamma_node : public jlm::rvsdg::structural_node {
 	jlm::rvsdg::gamma_output *
 	add_exitvar(const std::vector<jlm::rvsdg::output*> & values);
 
+  /**
```
sjalander, line 289: This seems to be a generic functionality, so why isn't it part of the structural node instead?

sjalander (commented):
Why is gamma treated separately?
Isn't it possible to have a generic Remove*Where for the structural node?
Else, why doesn't the theta have the same functionality?

jlm/rvsdg/region.hpp
phate: I am not sure I follow. What do you mean?

jlm/rvsdg/gamma.hpp
phate: The function is specific to the semantics of the gamma node. I guess the comment could be more specific saying that it removes all outputs with the corresponding results. The function is a little bit the odd one out in the PR, but I did not want to remove it again as I will need it anyway later.

phate:
It is not treated separately, but I agree that it is a little bit the odd one out in this PR. I wanted to see whether the interface also works for concrete structural nodes (gamma, theta, etc.) and what I could do with it. Ultimately, the RemoveOutputsWhere of the gamma node should not be part of this PR, but I will need this function soon anyway (along with similar functions for theta, etc.), so I left it in the PR.

jlm/rvsdg/region.hpp
sjalander: Removes an argument from the region given a arguments' index.
 Removes an argument from the region given an arguments' index.

sjalander:
I did not think of gamma's special case that you need to adjust the result for each region.
But I guess this functionality should in the end exist for all structural nodes, so will the functions be defined as virtual once all the derived structural nodes hava this functionality?

haved:
Making the `RemoveOutputsWhere()` function virtual would remove the option of templating F, which would force us back to `std::function`. Additionally, the function would have to have the same signature for all types, so it would not allow predicates of type `bool (gamma_output& c)`, just `output&`.

I think perhaps the naming becomes a bit confusing, now that calling `RemoveOutput` on a `gamma_node` only removes the output, and not the regions' results, while `RemoveOutputWhere` removes both.

Since the behavior is doing more than removing the output, and also does gamma specific things, renaming it on a per implementation basis might be a better approach than trying to make all node types have an identical interface. If the latter is preferred, then I think `RemoveOutput()` should be overridden by `gamma_node` etc.

jlm/rvsdg/region.hpp
phate: Fixed.

phate:
Renamed the gamma_node::RemovesOutputsWhere to gamma_node::RemovesGammaOutputsWhere, and improved the methods documentation. Please ignore the RemoveOutput methods. As mentioned in the PR description, they are going to disappear from the interface. Moreover, as Håvard pointed out above:
1. Virtual is not an option due to templating.
2. The plan is to do a per implementation abstraction. If it turns out that we can wrap this up in the structural node in the end, then we can push it up one level in the class hierarchy. I do not fully see this yet though.

I hope that resolves the issue with the gamma nodes methods.

sjalander (approved):

PR #282, opened 2023-11-15
Add an attempt at a .clang-format file

haved:
There are two reasons I attempted to make a `.clang-format` file for the project.

First and foremost, I want to free up the time Nico spends correcting my curly bracket placement.
I'm also sometimes a bit unsure how to format certain things.

Secondly, I have started to use the [CLion Nova](https://blog.jetbrains.com/clion/2023/11/clion-nova/) preview, in hopes its RAM usage is more modest and stable than CLion Classic.
Reading their blog, it seems it will be merged into CLion in the future, and my experience so far has been ok.
It is better at informing about unnecessarily precise namespaces, inline and stuff.

Formatting (and completion / indenting as you type) is now done by either ReSharper or clang-format, so instead of mucking about in the ReSharper settings dialog, I thought it would be better to make a proper `.clang-format`.
Do however note, that the built in `clang-format` binary is a bit older, because it does some silly stuff. Passing in my system's `/usr/bin/clang-format` worked nicely.

**I'm not suggesting running clang-format on the whole project, or forcing formatting on new PRs, only as an aid to people who want it / IDEs inserting indentation when you press enter**

Without further ado, here is some example output, representing the `.clang-format` from the latest commit:

### Example
```cpp
#include <jlm/llvm/opt/Something.hpp>
#include <jlm/rvsdg/SomethingElse.hpp>

#include <vector>

namespace jlm::test
{

class MyClass final : public ParentClass
{
  class ForwardDeclared;

public:
  explicit MyClass(const char * name)
      : ParentClass(name),
        Member1_(param)
  {}

  explicit MyClass(int & param)
      : ParentClass("MyClass"),
        Member1_(param),
        Member2_(2),
        Member3_(3),
        Member4_(4),
        AVeryLongMember5ThatNeedsToBeSuperLong_(5)
  {}

  virtual ~MyClass() override = default;

  MyClass &
  MyClass(const MyClass & other) = delete;

  template<typename T>
  [[nodiscard]] int
  GetSum(util::HashSet<const PointsToGraph::MemoryNode *> hashSet)
  {
    if (Member1_ == 0)
      return Memeber1_;

    if (Member1_ > 10)
    {
      int & tmpVar = Member2_;
      tmpVar += Member3_;
    }
    else
    {
      std::swap(Member1_, Member2_);
    }

    return Member1_ + Member2_ + Member3_;
  }

  int
  VeryLongFunctionNameWithALotOfParameters(
      const SomeClass & param1,
      const OtherClass & param2,
      int x,
      int y,
      int z,
      Quaterion rotation,
      Vector3 scale)
  {
    switch (Member1_)
    {
    case 1:
      return 2;
    case 2:
    {
      Member2_++;
      return VeryLongFunctionNameWithALotOfParameters(
          param1,
          param2,
          x + 500,
          y - 400,
          z * 50 + 2,
          rotation,
          scale);
    }
    default:
      break;
    }

    auto myFancyLambda = augmentLambda(
        [&](int arg)
        {
          return arg;
        },
        arg2);

    return myFancyLambda(Member3_);
  }

private:
  /**
   * \brief a documentation comment
   */
  int Member1_;
  int Member2_, Member3_;
  int Member4_, AVeryLongMember5ThatNeedsToBeSuperLong_;
};

enum class Xyz : uint8_t
{
  Enum1,
  Enum2,
  Enum3
};

}
```

It is based on the LLVM format, with most unchanged rules removed for brevity.
Some unchanged rules are still included, in cases where I'm uncertain, or feel like it should be present.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -65,7 +65,6 @@ SpacesInAngles: Never
 # public:
 #   int field;
 AccessModifierOffset: -2
-AlignAfterOpenBracket: Align
 
 AlignEscapedNewlines: Left
 AlignOperands: AlignAfterOperator
@@ -76,6 +75,8 @@ BinPackArguments: false
 BinPackParameters: false
 AllowAllArgumentsOnNextLine: false
 AllowAllParametersOfDeclarationOnNextLine: false
+# If parameters/arguments don't fit on a single line, put the first parameter on the next line
+AlignAfterOpenBracket: AlwaysBreak
 
 # Always give constructor initializers their own line
 PackConstructorInitializers: Never
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,119 @@
+---
+Language:      Cpp
+Standard:      Latest
+BasedOnStyle:  LLVM
+
+# int * a and int & a
+PointerAlignment: Middle
+ReferenceAlignment: Pointer
+SpaceAroundPointerQualifiers: Both
+# Don't touch position of "const int *" vs "int const *"
+QualifierAlignment: Leave
+
+ColumnLimit: 100
+
+# Indentation width related flags
+IndentWidth: 2
+TabWidth:    2
+UseTab:      Never
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+IndentCaseBlocks: false
+IndentCaseLabels: false
+IndentExternBlock: NoIndent
+NamespaceIndentation: None
+
+# Newline rules
+InsertNewlineAtEOF: true
+LineEnding: LF
+MaxEmptyLinesToKeep: 1
+
+# Collection of all brace wrapping rules
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterCaseLabel:  false
+  AfterClass:      true
+  AfterControlStatement: Never
+  AfterEnum:       true
+  AfterExternBlock: true
+  AfterFunction:   true
+  AfterNamespace:  true
+  AfterStruct:     true
+  AfterUnion:      true
+  BeforeCatch:     true
+  BeforeElse:      false
+  BeforeLambdaBody: true
+  BeforeWhile:     false
+  IndentBraces:    false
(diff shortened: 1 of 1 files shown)


phate (changes requested):
I really appreciate that you look into this. It was on our TODO list since a long time. Let's start with these values, it looks like a decent starting point. If necessary, we can adjust as we go.

phate:
In addition, we should also have:

BinPackArguments: false
BinPackParameters: false
AllowAllParametersOfDeclarationOnNextLine: false

phate:
It seems there is no way to force arguments/parameters to one per line be default. It only seems to kick in when it overflows. https://discourse.llvm.org/t/clang-format-force-parameters-to-one-line-each-even-if-the-statement-does-not-overflow/67903

haved:
I added your changes, I'm a bit surprised you want `BraceWrapping.AfterControlStatement: Always`, as we tend to write
```
if (conditional) {
  statement1;
  statement2;
}
```
but I can see that some `if` and `for` statements in the repository do have the bracket on a separate line.

I see that you don't want `PackConstructorInitializers: NextLine`, but are you sure you want to go directly to `Never`, and not stop at `CurrentLine`? For ctors with just a couple of members it would be more compact, but I guess it's better to stick to keeping one "statement" per line?

I updated the example in the PR description.

phate:
> I see that you don't want PackConstructorInitializers: NextLine, but are you sure you want to go directly to Never, and not stop at CurrentLine? For ctors with just a couple of members it would be more compact, but I guess it's better to stick to keeping one "statement" per line?

Yes, I would like to have `Never` for `PackConstructorInitializers`. `CurrentLine` would be more compact, but that is not really what I am interested in. I do not care about a few code lines more, but rather the mental overhead of reading a line. Since my current job, I started to realize how much easier it is if one only has to worry about "one expression/statement/thing" per line.

> This also applies to function calls.
Am I correct in assuming this is what you prefer?

You assumed correctly. :-) Thanks a lot for doing this.

phate (approved):

PR #283, opened 2023-11-18
Reformat all files and add clang format workflow

phate:
This PR does the following:

1. Reformats ALL files according to the formatting rules given in .clang-format. The following command was used: `find . -name "*.cpp" -or -name "*.hpp" | xargs clang-format-16 -i -style="file:.clang-format"`
2. Adds a workflow to ensure that no code is merged which violates the formatting rules.


phate:
@sjalander and @haved : Could I get this PR reviewed, please? I would like to get it merged and out of the way as it creates a lot of conflicts.

.github/workflows/tests.yml
```diff
@@ -14,6 +14,21 @@ jobs:
     - name: Generate documentation
       run: make docs
 
+  ClangFormat:
```
haved, line 17: For some reason this check was running on my PR before it got merged, and it was just stalling. Looking at my PR now it is no longer referenced, so it might have just been a GirHub bug? The action itself looks good.

haved (approved):

.github/workflows/tests.yml
phate: It was not a github bug. It was a user bug. Sleepy me put it in the required list before it was merged. I removed it before I merged your PR.

PR #285, opened 2023-11-19
Create LLVM composite action

phate:
Creates a composite action for installing all the required LLVM dependencies and uses it in the various workflows.

### DIFF
--- a/.github/actions/InstallLlvmDependencies/action.yml
+++ b/.github/actions/InstallLlvmDependencies/action.yml
@@ -13,5 +13,5 @@ runs:
       shell: bash
 
     - name: "Install LLVM dependencies"
-      run: sudo apt-get update && sudo apt-get install llvm-16-dev clang-16 clang-format-16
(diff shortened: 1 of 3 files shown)


phate:
I just realized that we are compiling jlm in release, i.e., without JLM_DEBUG and JLM_ENABLE_ASSERTS, when running all the tests. I am not sure how this screw up happened, but I will fix it in another PR.

haved:
Oh wow, I see. I had noticed that tests use `assert` and not `JLM_ASSERT`, but I didn't think it really made a difference.

Will you fix this by changing the actions yaml, or modifying the Makefile to make separate build directories for the different targets?

haved (approved):
Looks good, nice deduplication. The action still has to run for every job, I tried looking for an official docker image with llvm binaries installed, but couldn't find one, so this is probably the next best thing.

sjalander (approved):

PR #286, opened 2023-11-23
Add delta input removal methods

phate:
This PR adds an API for delta input removal. It contains the following:

1. Add RemoveDeltaInputsWhere and PruneDeltaInputs method to delta node class
2. These implementation hide nasty implementation details, such as iteration order, from outside the class hierarchy
3. Removes the usage of RemoveInput and RemoveArgument from outside the structural node class hierarchy.
4. Adds a test for delta nodes to the TestDeadNodeElimination.cpp. This node was untested.

The implementation of RemoveDeltaInputsWhere will be simplified further down the road. The implementation details, such as iteration order, will be pushed further up the class hierarchy at a later point in time.

### DIFF
--- a/jlm/llvm/ir/operators/delta.cpp
+++ b/jlm/llvm/ir/operators/delta.cpp
@@ -110,6 +110,12 @@ node::input(size_t n) const noexcept
   return static_cast<cvinput *>(structural_node::input(n));
 }
 
+cvargument *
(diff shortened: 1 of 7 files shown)


haved (approved):
Looks great, i like the function names, documentation comments, and renamed variables a lot

PR #287, opened 2023-11-23
Add lambda input removal methods

phate:
This PR adds an API for lambda input removal. It contains the following:

1. Add RemoveLambdaInputsWhere and PruneLambdaInputs method to lambda node class
2. These implementation hide nasty implementation details, such as iteration order, from outside the class hierarchy
3. Removes the usage of RemoveInput and RemoveArgument from outside the structural node class hierarchy.

The implementation of RemoveLambdaInputsWhere will be simplified further down the road. The implementation details, such as iteration order, will be pushed further up the class hierarchy at a later point in time.

### DIFF
--- a/jlm/llvm/ir/operators/lambda.hpp
+++ b/jlm/llvm/ir/operators/lambda.hpp
@@ -276,8 +276,9 @@ public:
    *
    * \see cvargument#IsDead()
    */
-  template <typename F> size_t
-  RemoveLambdaInputsWhere(const F& match);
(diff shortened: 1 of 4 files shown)


haved (approved):
Looks good.

I'm seeing that `output::IsDead()` is being added by this PR as well, with a different commit hash despite the commits being identical across the PRs.

Hopefully git figures it all out :)

phate:
It did figure it out. :-)

PR #290, opened 2023-11-28
Clean up HLS theta conversion

phate:
I have quite some problems with refactoring the HLS code. Start to clean it up such that I can familiarize myself better with the code and gain a better starting position for refactoring. This PR overhauls the theta conversion pass.

### DIFF
--- a/jlm/hls/backend/rvsdg2rhls/theta-conv.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/theta-conv.cpp
@@ -11,56 +11,66 @@ namespace jlm::hls
 {
 
 static void
-theta_conv(jlm::rvsdg::theta_node * theta)
+ConvertTheta(jlm::rvsdg::theta_node & theta)
 {
   jlm::rvsdg::substitution_map smap;
 
-  auto loop = hls::loop_node::create(theta->region());
+  auto loop = hls::loop_node::create(theta.region());
   std::vector<jlm::rvsdg::input *> branches;
 
   // add loopvars and populate the smap
-  for (size_t i = 0; i < theta->ninputs(); i++)
+  for (size_t i = 0; i < theta.ninputs(); i++)
   {
     jlm::rvsdg::output * buffer;
-    loop->add_loopvar(theta->input(i)->origin(), &buffer);
-    smap.insert(theta->input(i)->argument(), buffer);
+    loop->add_loopvar(theta.input(i)->origin(), &buffer);
+    smap.insert(theta.input(i)->argument(), buffer);
     // buffer out is only used by branch
     branches.push_back(*buffer->begin());
     // divert theta outputs
-    theta->output(i)->divert_users(loop->output(i));
+    theta.output(i)->divert_users(loop->output(i));
   }
 
   // copy contents of theta
-  theta->subregion()->copy(loop->subregion(), smap, false, false);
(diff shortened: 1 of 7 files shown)


jlm/hls/backend/rvsdg2rhls/theta-conv.cpp
```diff
@@ -10,67 +10,72 @@
...
-          theta_conv(structnode->subregion(n));
-      }
+      ConvertThetaStructuralNode(*structuralNode);
     }
   }
 }
 
 void
-theta_conv(jlm::llvm::RvsdgModule & rm)
+theta_conv(jlm::llvm::RvsdgModule & rvsdgModule)
```
sjalander, line 76: Why not update the function name?

sjalander: Maybe call it ConvertAllTheta()

jlm/hls/backend/rvsdg2rhls/theta-conv.cpp
```diff
@@ -10,67 +10,72 @@
...
+  }
+
+  if (auto thetaNode = dynamic_cast<jlm::rvsdg::theta_node *>(&structuralNode))
+  {
+    ConvertTheta(*thetaNode);
+  }
+}
+
+static void
+ConvertThetaRegion(jlm::rvsdg::region & region)
```
sjalander, line 64: This is the one that traverses the graphs and converts thetas, so it would make sense to call this one something with convert all thetas.

sjalander (commented):
I think the naming of the functions are missleading
ConvertThetaRegion is not taking a theta region and converting it but traverses the given region and finds all thetas that then get converted.
conv_theta() is not following the naming convention.
Previously, I've seen that you changed to explicitly name the function based on the argument type that is taken.

So maybe 
ConvertAllThetaInRegion()
ConvertAllThetaInModule()

jlm/hls/backend/rvsdg2rhls/theta-conv.cpp
```diff
@@ -10,67 +10,72 @@
...
 
-void
-theta_conv(jlm::rvsdg::region * region)
+static void
+ConvertThetaRegion(jlm::rvsdg::region & region);
+
+static void
+ConvertThetaStructuralNode(jlm::rvsdg::structural_node & structuralNode)
+{
+  for (size_t n = 0; n < structuralNode.nsubregions(); n++)
```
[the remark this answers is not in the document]
phate: Renamed the functions to hopefully be clearer.

jlm/hls/backend/rvsdg2rhls/theta-conv.cpp
phate: Renamed to ConvertThetaNodes.

jlm/hls/backend/rvsdg2rhls/theta-conv.cpp
phate: Renamed to ConvertThetaNodesInRegion

phate:
Changes since last review:
1. Renamed the functions
2. Renamed the files from theta-conv.hpp/*.cpp to ThetaConversion.hpp/*.cpp
3. Improved the documentation of ConvertThetaNodes function

PR #292, opened 2023-11-29
CI: Building CIRCT using an action

sjalander:
This simplifies the HLS test flow.

### DIFF
--- a/.github/actions/BuildCirct/action.yml
+++ b/.github/actions/BuildCirct/action.yml
@@ -1,0 +1,42 @@
+name: "Build CIRCT"
+description: "Builds CIRCT, which is used for the HLS backend"
+
+runs:
+  using: "composite"
+  steps:
+    # Clone jlm-eval-suite
+    - name: Clone jlm-test-suite
+      run: git clone https://github.com/phate/jlm-eval-suite.git
+      shell: bash
+
+    # Extract the CIRCT submodule hash for use in the cache key
+    - name: Get CIRCT hash
+      id: get-circt-hash
+      run: |
+        cd jlm-eval-suite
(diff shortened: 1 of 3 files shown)


.github/workflows/hls.yml
```diff
@@ -71,46 +31,11 @@ jobs:
       with:
         fetch-depth: '0'
 
-    # --------
-    # Install dependencies
-    # --------
-
     - name: "Install LLVM, Clang, MLIR, and Ninja"
       uses: ./.github/actions/InstallLlvmDependencies
```
phate, line 35: Do these lines still need to be here? Are the dependencies not installed in the BuildCirct composite action?

.github/actions/BuildCirct/action.yml
```diff
@@ -0,0 +1,42 @@
...
+      id: cache-circt
+      uses: actions/cache@v3
+      with:
+        path: |
+          jlm-eval-suite/circt/local
+        key: ${{ runner.os }}-circt-${{ steps.get-circt-hash.outputs.hash }}
+
+    # Install dependencies if we didn't hit in the cache
+    - name: "Install LLVM, Clang, MLIR, and Ninja"
+      if: steps.cache-circt.outputs.cache-hit != 'true'
```
sjalander, line 31: The installation of dependencies are conditional

.github/workflows/hls.yml
sjalander: The dependencies are only installed if you miss in the cache for the build CIRCT action.
See comment below.

phate: I missed that the first time. Thanks.

phate:
If you update the branch, I will merge it.

phate (approved):

PR #297, opened 2023-12-05
Add overloads for rvsdg::view and PointsToGraph::ToDot to allow giving the same unique names to outputs in both

haved:
In practice this means the `PointsToGraph`'s dot output now adds an extra part to `RegisterNode`s, as seen here:
image

The names in parentheses correspond to the names given by rvsdg::view in the following output:
```
[]{
  o0 := DELTA[MyList] 
    []{
      o1 := ConstantPointerNull 
    }[o1]
  o2 := LAMBDA[next] o0 
    [a3, a4, a5, a6 <= o0]{
      o7 := BITS32(0) 
      o8 := BITS32(4) 
      o9 o10 := ALLOCA[ptr] o8 
      o11 := MemStateMerge o10 a4 
      o12 o13 := LOAD a6 o11 
      o14 := STORE o9 o12 o13 
      o15 o16 := LOAD o9 o14 
      o17 := GetElementPtr o15 o7 o7 
      o18 o19 := LOAD o17 o16 
      o20 := STORE o9 o18 o19 
      o21 o22 := LOAD o9 o20 
    }[o21, a3, o22, a5]
}[o0, o2]
```

The example is generated by adding the following lines to the top of `TestAndersen.cpp`'s `TestLinkedList()`:
```cpp
  std::unordered_map<const jlm::rvsdg::output*, std::string> outputMap;
  std::cout << jlm::rvsdg::view(test.graph().root(), outputMap) << std::endl;
  std::cout << jlm::llvm::aa::PointsToGraph::ToDot(*ptg, outputMap) << std::endl;
```

The functionality is added through overloads that allow specifying the `unordered _map<const output*, std::string>` used to give outputs unique names. The `ToDot` method has been modified to be able to take in such a map, and use it while making output. It is not the cleanest approach ever, but it was very helpful during debugging.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
+++ b/jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
@@ -216,7 +216,7 @@ PointsToGraph::ToDot(
         "\"]; }\n");
   };
 
-  auto edgeString = [&](const PointsToGraph::Node & node, const PointsToGraph::Node & target)
+  auto edgeString = [](const PointsToGraph::Node & node, const PointsToGraph::Node & target)
   {
     return util::strfmt(
         reinterpret_cast<uintptr_t>(&node),
--- a/jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
+++ b/jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
@@ -159,8 +159,19 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
 }
 
 std::string
-PointsToGraph::ToDot(const PointsToGraph & pointsToGraph)
+PointsToGraph::ToDot(
+    const PointsToGraph & pointsToGraph,
+    const std::unordered_map<const rvsdg::output *, std::string> & outputMap)
 {
+  auto nodeFill = [&](const PointsToGraph::Node & node)
+  {
+    // Nodes that are marked as having escaped the module get a background color
+    if (const auto memoryNode = dynamic_cast<const MemoryNode *>(&node))
+      if (pointsToGraph.GetEscapedMemoryNodes().Contains(memoryNode))
+        return "style=filled, fillcolor=\"yellow\", ";
+    return "";
+  };
+
   auto nodeShape = [](const PointsToGraph::Node & node)
   {
     static std::unordered_map<std::type_index, std::string> shapes(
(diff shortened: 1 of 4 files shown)


jlm/rvsdg/view.hpp
```diff
@@ -16,18 +16,46 @@ namespace jlm::rvsdg
...
 void
 view(const jlm::rvsdg::region * region, FILE * out);
 
-static inline void
+/**
+ * Recursively traverses and prints the root region of the given rvsdg graph to the given file.
+ * @param graph the rvsdg graph to be printed.
+ * @param out the file to be written to.
+ */
+inline void
```
phate, line 53: I think the `static` needs to stay, otherwise the linker will give you an error when trying to link two object files where `view.hpp` was included. The `include` can go as far as I can see.

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -179,23 +190,39 @@ PointsToGraph::ToDot(const PointsToGraph & pointsToGraph)
...
-        node.DebugString(),
+        nodeLabel(node),
         "\" ",
         "shape = \"",
         nodeShape(node),
         "\"]; }\n");
   };
 
-  auto edgeString = [](const PointsToGraph::Node & node, const PointsToGraph::Node & target)
+  auto edgeString = [&](const PointsToGraph::Node & node, const PointsToGraph::Node & target)
```
phate, line 219: I believe the `&` is not needed here.

phate (changes requested):
I am not excited about this, but I definitely see its usefulness. I will happily merge it.

The entire graphviz creation should be overhauled. What should happen is the introduction of a proper data model, a pretty printer for it, and a builder that allows to customize the graphviz output depending on where it is used (PointsToGraph, Rvsdg, LocationSet, ConstraintGraph, ...). This builder could then also take such names into consideration. However, this is a little bit more work and given the limited engineering time, it was never high priority. Maybe we should assign it a higher priority.

jlm/rvsdg/view.hpp
haved: That's what inline means, two object files can have their own definition, and they can safely be merged at link time, resulting in only one being left in the final binary. Static means each object file gets its own implementation.

Having both (top-level) static and inline will never make sense, and static "wins". There are also a bunch of rules about when inline is implicit, e.g. when you define method bodies inside a class/struct definition, or when you define a function with template parameters.

phate: You are absolutely right. I must have been asleep when I wrote it.

phate:
@haved I would merge this one as it is. The only other comment I had is not a big problem, unless you want to fix it?

haved:
I removed the `&`, you can merge

PR #298, opened 2023-12-05
Add BijectiveMap<K,V> datastructure class

haved:
A helper class for storing bijective mappings, with fast lookups in both directions.


jlm/util/BiMap.hpp
```diff
@@ -0,0 +1,254 @@
...
+#define JLM_UTIL_BI_MAP_HPP
+
+#include <jlm/util/common.hpp>
+
+#include <unordered_map>
+
+namespace jlm::util
+{
+/**
+ * Represents a bijective mapping between keys and values, with fast loopups by both key and value.
```
phate, line 16: > with fast loopups
should be changed to "with fast lookups". Could you be more precise here and specify what fast means, i.e., amortized constant time?

tests/jlm/util/TestBiMap.cpp
```diff
@@ -0,0 +1,168 @@
...
+    try                                                \
+    {                                                  \
+      code;                                            \
+      assert(false && #code " was supposed to throw"); \
+    }                                                  \
+    catch (jlm::util::error e)                         \
+    {                                                  \
+      (void)e;                                         \
+    }                                                  \
+  } while (false)
```
phate, line 25: We should put this somewhere else. This is a very useful code snippet. There are other places in the unit tests where this piece would be useful.

jlm/util/BiMap.hpp
```diff
@@ -0,0 +1,254 @@
...
+  {
+    for (auto & [key, value] : ForwardMap_)
+      AddPairToReverse(key, value);
+  }
+
+  /**
+   * Removes all key/value mappings from the BiMap.
+   */
+  void
+  Clear() noexcept
```
phate, line 68: Could we get a unit test on this one? I did not find one.

jlm/util/BiMap.hpp
```diff
@@ -0,0 +1,254 @@
...
+    ReverseMap_.clear();
+  }
+
+  /**
+   * Gets the number of key/value mappings in the BiMap.
+   */
+  [[nodiscard]] std::size_t
+  Size() const noexcept
+  {
+    return ForwardMap_.size();
```
phate, line 80: Could we insert a `JLM_ASSERT(ForwardMap_.size() == ReverseMap_.size())` for good measure before the return statement?

jlm/util/BiMap.hpp
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  /**
+   * Inserts a new key/value pair into the BiMap,
+   * @param key the key to be inserted, with the associated value
+   * @param value the value to be inserted, associated with the key
+   * @throws jlm::util::error if the key or value already exists in the BiMap
+   */
+  void
```
phate, line 89: Alternative suggestion for signature: `bool Insert(const K & key, const V & value)`, where the bool indicates whether the value was inserted or not. The user can then decide themselves whether an exception would be appropriate. Maybe you could also provide both Insert methods?

jlm/util/BiMap.hpp
```diff
@@ -0,0 +1,254 @@
...
+   * @param key the key to lookup in the BiMap
+   * @returns the value associated with the given \p key
+   * @throws jlm::util::error if the \p key does not exist in the BiMap
+   */
+  [[nodiscard]] const V &
+  LookupKey(const K & key) const
+  {
+    auto it = ForwardMap_.find(key);
+    if (it == ForwardMap_.end())
+      throw jlm::util::error("Key not found in BiMap");
```
phate, line 131: The same error message is used multiple times in the code. Could we refactor it out to a private static string in the class?

jlm/util/BiMap.hpp
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  /**
+   * @returns the underlying mapping from values to keys
+   */
+  [[nodiscard]] const std::unordered_map<V, K> &
+  GetReverseMap() const noexcept
+  {
+    return ReverseMap_;
+  }
```
phate, line 203: I really do not like these two functions as they are basically exposing the internal implementation of the bijective map to the caller. Why do you need them?

jlm/util/BiMap.hpp
```diff
@@ -0,0 +1,254 @@
...
+
+  /**
+   * Shorthand for iterating over all key/value pairs
+   * @returns a const iterator pointing to the end of the ForwardMap
+   */
+  [[nodiscard]] typename std::unordered_map<K, V>::const_iterator
+  end() const noexcept
+  {
+    return ForwardMap_.cend();
+  }
```
phate, line 223: The same here. The returned `const_iterator`s expose the implementation. Could we wrap them in your own BijectiveMap::ConstIterator?

jlm/util/BiMap.hpp
```diff
@@ -0,0 +1,254 @@
...
+    auto it = ReverseMap_.find(value);
+    if (it == ReverseMap_.end())
+      return false;
+
+    auto removed = ForwardMap_.erase(it->second);
+    JLM_ASSERT(removed == 1);
+
+    ReverseMap_.erase(it);
+    return true;
+  }
```
phate, line 185: Would it be useful to provide a RemoveKeys() and RemoveValues() method? In other words:

```
template <typename F>
size_t
RemoveKeys(const F & match)
```

jlm/util/BiMap.hpp
```diff
@@ -0,0 +1,254 @@
...
+
+  /**
+   * Inserts a new key/value pair into the BiMap,
+   * @param key the key to be inserted, with the associated value
+   * @param value the value to be inserted, associated with the key
+   * @throws jlm::util::error if the key or value already exists in the BiMap
+   */
+  void
+  Insert(const K & key, const V & value)
```
phate, line 90: Would it also be useful to provide a method for inserting multiple values?

phate (changes requested):
I like the code. Thanks for doing this. There would immediately be another place where this map would be applicable: https://github.com/phate/jlm/blob/7637fcc2095308db30377cf2a5999b15993aaff7/jlm/llvm/frontend/LlvmConversionContext.hpp#L35C1-L35C1

PR #305, opened 2023-12-05
make: Added clang format targets

sjalander:
One target for applying clang formatting and one for dry run that only reports formatting errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,8 @@ help:
 	@$(HELP_TEXT)
 	@$(HELP_TEXT_JLM)
 	@echo ""
+	@echo "JLM Aliases"
+	@echo "--------------------------------------------------------------------------------"
 	@echo "all                    Compile jlm in release mode, and run unit and C tests"
 	@echo "release                Alias for jlm-release"
 	@echo "debug                  Alias for jlm-debug and check"
--- a/Makefile.sub
+++ b/Makefile.sub
@@ -11,6 +11,11 @@ echo "jlc-release            Compile jlc compiler in release mode"
 echo ""
 echo "jlm-opt-debug          Compile jlm optimizer in debug mode"
 echo "jlm-opt-release        Compile jlm optimizer in release mode"
+echo ""
+echo "Clang format Targets"
+echo "--------------------------------------------------------------------------------"
+echo "format                 Format all cpp and hpp files"
+echo "format-dry-run         Report incorrect formatting of cpp and hpp files"
 endef
 
 # Try to detect llvm-config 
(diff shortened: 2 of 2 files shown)


haved:
Very nice initiative, will definitely be useful.

On my machine, the following works, to give all files that are either cpp or hpp:
```
find . -name "*.[ch]pp" -exec ...
```

A slight annoyance on my part is that `clang-format-16` isn't a thing, despite `/urs/bin/clang-format` being version 16. Using the `llvm-config --bindir` doesn't help much either, since clang-format is a project outside of the rest of llvm, as far as I can tell. Either we do something akin to
```
ifdef CLANGFORMAT
  $(info CLANGFORMAT specified by user)
else ifdef $(shell command -v clang-format-16 2> /dev/null)
  $(info Found clang-format-16)
  CLANGFORMAT ?= clang-format-16
else ifdef $(shell command -v clang-format 2> /dev/null)
  $(info Found clang-format (It should be version 16))
  CLANGFORMAT ?= clang-format
endif
```
or I can of course create an alias and not worry about it.

We should probably also change the `.github/workflows/tests.yml`'s `ClangFormat` job to run `make format`.

phate:
@caleridas : FYI, since you are overhauling make.

@haved: Make is the wrong tool for detecting installed dependencies on the system. There should come a configure step before or something like this. Also, I would not like to render the workflow dependent on this make target, as we are then dependent on detecting the right clang format. In contrast, currently I can specify the correct clang path in the workflow as we know the system it is running on (Ubuntu 22.04).

I am tempted to merge this as it is. Anyone has very strong objections?

haved:
I would be happy to have this merged, with or without the single `find` command modification.

I agree `make` is not the right tool for finding binary dependencies, and it would have to be a part of a configuration step, and ideally the LLVMCONFIG stuff should be as well.

I'm guessing autotools is a bit dated, and I got the impression that CMake is unpopular (understandably). I've never tried meson, bazel or SCons, so I can't judge them.

My only grievance with the current Makefile is that .cpp files are not always recompiled when headers change. This can be solved using the `-M` or `-MT` flag and storing dependencies, if it is not too much of a hassle to do in a manually written Makefile.

caleridas:
> @caleridas : FYI, since you are overhauling make.
> 
> @haved: Make is the wrong tool for detecting installed dependencies on the system. There should come a configure step before or something like this. Also, I would not like to render the workflow dependent on this make target, as we are then dependent on detecting the right clang format. In contrast, currently I can specify the correct clang path in the workflow as we know the system it is running on (Ubuntu 22.04).
> 
> I am tempted to merge this as it is. Anyone has very strong objections?

I will need to revert parts of it for the makefile changes. As far as config step goes, I was going to do a "dumb" shell script for now, could be substituted by autoconf later if desired (but could also do without).

phate:
@sjalander : Could you merge master into the branch such that I can merge the PR, please?

PR #306, opened 2023-12-05
Disabled buggy optimizations in reduction (-Jred)

sjalander:
This version of -Jred passes the llvm-test-suite

### DIFF
--- a/jlm/llvm/opt/reduction.cpp
+++ b/jlm/llvm/opt/reduction.cpp
@@ -98,7 +98,7 @@ enable_load_reductions(jlm::rvsdg::graph & graph)
   auto nf = LoadOperation::GetNormalForm(&graph);
   nf->set_mutable(true);
   nf->set_load_mux_reducible(true);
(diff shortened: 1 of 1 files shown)


jlm/llvm/opt/reduction.cpp
```diff
@@ -98,7 +98,7 @@ enable_load_reductions(jlm::rvsdg::graph & graph)
   auto nf = LoadOperation::GetNormalForm(&graph);
   nf->set_mutable(true);
   nf->set_load_mux_reducible(true);
-  nf->set_load_store_reducible(true);
+  //  nf->set_load_store_reducible(true);
```
phate, line 101: Could we explicitly set them to false instead of commenting them out? Also, could we add a comment above them why they are disabled? Maybe even with a link to the issues you created?

phate (changes requested):
Thanks a lot for wading through them to find the faulty ones!

PR #311, opened 2023-12-07
Improve call node accessors

### LINKED ISSUE #295: CallNode's Argument method has surprising indexing
The code in `CallNode` looks like the following:
```cpp
  [[nodiscard]] size_t
  NumArguments() const noexcept
  {
    return ninputs() - 1;
  }

  [[nodiscard]] jlm::rvsdg::input *
  Argument(size_t n) const
  {
    return input(n);
  }
```

Judging by this code, you would think that the first `ninputs() - 1` inputs are arguments for the function call, and the last input is the target function itself. This is not the case, and I think it would be clearer to do

```cpp
  [[nodiscard]] jlm::rvsdg::input *
  Argument(size_t n) const
  {
    return input(n + 1);
  }
```

Note that there is also `CallNode::GetFunctionInput()`, which returns `input(0)`.

It is also an open question if the `GetIoStateInput()`, `GetMemoryStateInput()` and `GetLoopStateInput()` inputs should be included in `NumArguments`/`Arguments` or not. On the one hand they don't correspond to arguments in the source program, but they line up nicely with the arguments in the lambda node's subregion.

phate:
This PR closes #295. It makes the following changes:
1. Fixes a bug in the CallNode::Argument() method
2. Adds documentation to all call node accessors.
3. Adds a unit test for the accessors

### DIFF
--- a/jlm/llvm/ir/operators/call.hpp
+++ b/jlm/llvm/ir/operators/call.hpp
@@ -287,7 +287,8 @@ public:
   [[nodiscard]] jlm::rvsdg::input *
   Argument(size_t n) const
   {
-    return input(n);
+    JLM_ASSERT(n < NumArguments());
+    return input(n + 1);
   }
 
   [[nodiscard]] size_t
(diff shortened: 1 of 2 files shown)


haved (approved):
Great work, I'm surprised this change didn't need any other changes in the codebase, but I can't find any usage of `Argument()` (other than the ones added by me)

PR #315, opened 2023-12-10
Ensure creation of unique intermediate files throughout compilation

### LINKED ISSUE #312: Ensure the creation of unique file names throughout compilation
Currently, we do not create unique names for the intermediate files when compiling with [jlc](https://github.com/phate/jlm/blob/4103aae7d93ddd86475ef832b54a8b74d03a3ad6/jlm/tooling/CommandGraphGenerator.cpp#L19C1-L30C2) or [jhls](https://github.com/phate/jlm/blob/4103aae7d93ddd86475ef832b54a8b74d03a3ad6/jlm/tooling/CommandGraphGenerator.cpp#L189C1-L199C2). This is how we currently create the intermediate files for jlc:

```
util::filepath
JlcCommandGraphGenerator::CreateJlmOptCommandOutputFile(const util::filepath & inputFile)
{
  return util::strfmt("/tmp/tmp-", inputFile.base(), "-jlm-opt-out.ll");
}

util::filepath
JlcCommandGraphGenerator::CreateParserCommandOutputFile(const util::filepath & inputFile)
{
  return util::strfmt("/tmp/tmp-", inputFile.base(), "-clang-out.ll");
}
```

This could lead to problems when compiling multiple projects in parallel with jlc (and the projects contain files with the same file name).

We already have the [machinery in jlm](https://github.com/phate/jlm/blob/4103aae7d93ddd86475ef832b54a8b74d03a3ad6/jlm/util/file.hpp#L203C1-L216C4) for creating unique filenames, but it also needs to be used.

phate:
We currently create intermediate files while compiling, but might run into problems when we compile in parallel as the intermediate file names are derived from the input file names. The intermediate file names were created by adding a suffix to the input file name. If we compile two input files with the same name in parallel, we create the same intermediate file names even though those two intermediate files originate from different input file names. This PR resolves this issue by adding some randomness into the intermediate file name. It solves this problems for the jlc and jhls command pipeline. Closes #312.

### DIFF
--- a/jlm/tooling/CommandGraphGenerator.cpp
+++ b/jlm/tooling/CommandGraphGenerator.cpp
@@ -106,11 +106,12 @@ JlcCommandGraphGenerator::GenerateCommandGraph(const JlcCommandLineOptions & com
   {
     auto lastNode = &commandGraph->GetEntryNode();
 
+    util::filepath lastNodeOutputFile("");
     if (compilation.RequiresParsing())
     {
-      auto outputFile = CreateParserCommandOutputFile(compilation.InputFile());
+      lastNodeOutputFile = CreateParserCommandOutputFile(compilation.InputFile());
       auto & parserCommandNode =
-          CreateParserCommand(*commandGraph, outputFile, compilation, commandLineOptions);
+          CreateParserCommand(*commandGraph, lastNodeOutputFile, compilation, commandLineOptions);
 
       lastNode->AddEdge(parserCommandNode);
       lastNode = &parserCommandNode;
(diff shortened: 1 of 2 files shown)


jlm/tooling/CommandGraphGenerator.cpp
```diff
@@ -275,7 +291,7 @@ JhlsCommandGraphGenerator::GenerateCommandGraph(const JhlsCommandLineOptions & c
       auto & parserNode = ClangCommand::CreateParsingCommand(
           *commandGraph,
           compilation.InputFile(),
-          tmp_folder.to_str() + CreateParserCommandOutputFile(compilation.InputFile()).to_str(),
+          CreateParserCommandOutputFile(tmp_folder, compilation.InputFile()).to_str(),
```
sjalander, line 294: Line 265/281
Using "/tmp/" for the tmp_folder, should it not be std::filesystem::temp_directory_path().string()

sjalander (changes requested):
Still one use of hard-coded "/tmp/" in CommandGraphGenerator.cpp

jlm/tooling/CommandGraphGenerator.cpp
phate: The entire tmp_folder creation is messy and should not be part of this function in the first place.

phate: Fixed

sjalander (approved):

Issue #319, opened 2023-12-12
--JAASteensgaardAgnostic cause Assertion `lambda.nfctarguments() == callNode.ninputs() - 1' failed


sjalander:
**Recreate**
_make_

**Output**

jlc: jlm/llvm/opt/alias-analyses/Steensgaard.cpp:1132: auto jlm::llvm::aa::Steensgaard::AnalyzeCall(const CallNode &)::(anonymous class)::operator()(const CallNode &, const lambda::node &) const: Assertion `lambda.nfctarguments() == callNode.ninputs() - 1' failed.

[20051012-1.zip](https://github.com/phate/jlm/files/13645814/20051012-1.zip)

phate:
Handled by PR #326

Issue #274, opened 2023-11-05
-JAASteensgaardRegionAware Assertion `lambda.nfctarguments() == callNode.ninputs() - 1' failed


sjalander:
**Recreate**
_make_

**Code**

void
print_pgm()
{
}

void
search(void)
{
   print_pgm(3);
}

**Assert**

jlc: jlm/llvm/opt/alias-analyses/Steensgaard.cpp:1035: auto jlm::llvm::aa::Steensgaard::AnalyzeCall(const jlm::llvm::CallNode &)::(anonymous class)::operator()(const jlm::llvm::CallNode &, const lambda::node &) const: Assertion `lambda.nfctarguments() == callNode.ninputs() - 1' failed.

**Back trace**

#7  0x00005555557540a0 in jlm::llvm::aa::Steensgaard::AnalyzeCall(jlm::llvm::CallNode const&)::$_21::operator()(jlm::llvm::CallNode const&, jlm::llvm::lambda::node const&) const (this=0x7fffffffc8a8, callNode=..., lambda=...)
    at jlm/llvm/opt/alias-analyses/Steensgaard.cpp:1035
#8  0x0000555555753fa8 in jlm::llvm::aa::Steensgaard::AnalyzeCall (
    this=0x7fffffffdbc8, callNode=...)
    at jlm/llvm/opt/alias-analyses/Steensgaard.cpp:1127
#9  0x000055555575a301 in jlm::llvm::aa::Steensgaard::AnalyzeSimpleNode(jlm::rvsdg::simple_node const&)::$_4::operator()<jlm::llvm::aa::Steensgaard, jlm::rvsdg::simple_node const>(jlm::llvm::aa::Steensgaard&, jlm::rvsdg::simple_node const&) const (this=0x555555a19990, s=..., n=...)
    at jlm/llvm/opt/alias-analyses/Steensgaard.cpp:886
#10 0x000055555575a2c5 in std::__invoke_impl<void, jlm::llvm::aa::Steensgaard::AnalyzeSimpleNode(jlm::rvsdg::simple_node const&)::$_4&, jlm::llvm::aa::Steensgaard&, jlm::rvsdg::simple_node const&>(std::__invoke_other, jlm::llvm::aa::Steensgaard::AnalyzeSimpleNode(jlm::rvsdg::simple_node const&)::$_4&, jlm::llvm::aa::Steensgaard&, jlm::rvsdg::simple_node const&) (__f=..., __args=..., __args=...)
[...]

PR #326, opened 2023-12-19
Handle mismatch of call/lambda arguments in Steensgaard

phate:
LLVM permits code where the number of call arguments mismatch with the number of lambda arguments, even though it is a direct call. A simple example would be:

```
unsigned int                                                                                           
foo ()                                                                                                 
{                                                                                                      
  return 5;                                                                                            
}                                                                                                      
                                                                                                       
int                                                                                                    
main (void)                                                                                            
{                                                                                                      
  unsigned int x = 8;                                                                                  
  return foo(x);                                                                                       
} 
```

The Steensgaard alias analysis assumed that for direct calls the number of call and lambda arguments always coincide. This PR revises this assumption, and closes issue #319 and #274.

### DIFF
--- a/jlm/llvm/ir/operators/operators.hpp
+++ b/jlm/llvm/ir/operators/operators.hpp
@@ -1382,6 +1382,18 @@ public:
     return tac::create(op, arguments);
   }
 
+  static rvsdg::output *
+  Create(rvsdg::region & region, const std::vector<rvsdg::output *> & operands)
+  {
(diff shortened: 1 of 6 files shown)


sjalander (approved):

haved:
This applies to Andersen as well, I will create an issue for it specifically

phate:
This applies to all transformations that have to match call-side arguments/results with lambda arguments/results. Function inlining comes to mind as another one.

PR #329, opened 2023-12-21
jhls and jlm-hls exits if CIRCT is not enabled

### LINKED ISSUE #322: jhls no longer informs the user if it hasn't been compiled with CIRCT support
jhls should inform the user if it was compiled without CIRCT being enabled.
An assertion complaining about directory not existing is instead the result.

sjalander:
Both tools now informs the user if they are used and jlm has not been compiled with the CRICT backend enabled. Closes #322.

### DIFF
--- a/tools/jhls/jhls.cpp
+++ b/tools/jhls/jhls.cpp
@@ -9,6 +9,12 @@
 int
 main(int argc, char ** argv)
 {
+#ifndef CIRCT
+  ::llvm::outs() << "jhls has not been compiled with the CIRCT backend enabled.\n";
+  ::llvm::outs() << "Recompile jlm with -DCIRCT=1 if you want to use jhls.\n";
+  exit(0);
+#endif
+
   using namespace jlm::tooling;
 
   auto & commandLineOptions = JhlsCommandLineParser::Parse(argc, argv);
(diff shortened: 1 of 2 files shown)


tools/jhls/jhls.cpp
```diff
@@ -9,6 +9,12 @@
 int
 main(int argc, char ** argv)
 {
+#ifndef CIRCT
+  ::llvm::outs() << "jhls has not been compiled with the CIRCT backend enabled.\n";
+  ::llvm::outs() << "Recompile jlm with -DCIRCT=1 if you want to use jhls.\n";
+  exit(0);
```
phate, line 15: `exit(EXIT_SUCCESS)` rather than `exit(0)`

tools/jlm-hls/jlm-hls.cpp
```diff
@@ -44,6 +44,13 @@ llvmToFile(jlm::llvm::RvsdgModule & module, std::string fileName)
 int
 main(int argc, char ** argv)
 {
+
+#ifndef CIRCT
+  ::llvm::outs() << "jlm-hls has not been compiled with the CIRCT backend enabled.\n";
+  ::llvm::outs() << "Recompile jlm with -DCIRCT=1 if you want to use jlm-hls.\n";
+  exit(0);
```
phate, line 51: `exit(EXIT_SUCCESS)` rather than `exit(0)`

phate (changes requested):

phate (approved):

PR #336, opened 2023-12-28
Clean up Steensgaard

phate:
This PR only performs structural changes to the Steensgaard analysis. It should not change anything semantically. It is meant as a prerequisite before I start to tackle actual issues in the analysis.


haved (approved):
Looks very good, should make it easier to make future changes.

PR #337, opened 2023-12-29
Improve handling of imported entities in Steensgaard

phate:
This PR does the following:

1. Make MemoryLocation the base class of the ImportLocation class. This simplifies some code in the points-to graph construction pass.
2. Add documentation to the ImportLocation class.
3. Drop PointsToUnknown flag from imported memory locations. We know what they can point to, it is only to external memory and all memory locations that escaped from the module.
4. Move a FIXME that concerns the ImportLocation class also to the ImportLocation class
5. Add some documentation to the DummyLocation class
6. Clean up the AnalyzeImports method code a little bit.

This PR removes several FIXMEs from the code.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/Steensgaard.cpp
+++ b/jlm/llvm/opt/alias-analyses/Steensgaard.cpp
@@ -425,6 +425,10 @@ private:
  *
  * This class represents all global variable and function locations that are imported to the
  * translation unit.
+ *
+ * FIXME: We should be able to further distinguish imported locations between function and data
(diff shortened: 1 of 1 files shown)


sjalander (approved):

haved (approved):
LGTM.

I've been thinking about a more rigid way of handling memory objects that are "don't attempt to track pointers written to this memory object", which can be used for allocas, global variables and import nodes of non-pointer type, as well as functions. That is of course a concern for another day.

PR #338, opened 2024-01-03
Fix several bugs in points-to graph

phate:
This PR fixes four bugs introduced by this [PR](https://github.com/phate/jlm/pull/324):

1. It fixes the instantiation for the register-set node iterators. They were never instantiated and the code actually did not compile. A unit test was added that this cannot happen again.
2. The register-set nodes are now part of the DOT output. They were not printed before. :-/ This was only manually checked, and no unit test was added. We would need a graphviz data model to check this conveniently, i.e., without a string parsing mess.
3. It fixes the DebugString() method of the RegisterSetNode class. It segfaulted.
4. The iteration through the register set nodes of a points-to graph resulted in multiple visitations to the same register set nodes. The consequence was a messed up graphviz output. A unit test was added to ensure that this problem won't occur again.

Finally, the register-set node debug string was improved, making the dot graph more readable.

Moral of the story: Don't be too lazy to write unit test kids. All of these bugs could have been avoided if I would have invested in some unit tests for the points-to graph when I created the data structure. Oh well....

### DIFF
--- a/jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
+++ b/jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
@@ -357,6 +357,9 @@ PointsToGraph::ToDot(
   for (auto & registerNode : pointsToGraph.RegisterNodes())
     dot += printNodeAndEdges(registerNode);
 
+  for (auto & registerSetNode : pointsToGraph.RegisterSetNodes())
+    dot += printNodeAndEdges(registerSetNode);
+
   dot += nodeString(pointsToGraph.GetUnknownMemoryNode());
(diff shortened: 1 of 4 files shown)


haved (approved):
Looks good, thank you for discovering all this and writing the tests.
It would be nice to avoid the use of std::function in the iterators, as it adds indirect function calls to every visited item. If the function instead was part of the template for NodeIterator, it could be done at compile time.

Deduplicating NodeIterator and NodeConstIterator using template parameters /might/ also be possible, but I don't have many good experiences with const and templates.

phate:
Great suggestion! I added it as part of the template. Please have a look again whether it is acceptable now.

haved (approved):
Looks great, nice to get rid of the lambdas!

PR #339, opened 2024-01-05
Fix load-mux reduction

### LINKED ISSUE #323: -JAAAndersenAgnostic -JNodeReduction causes an assert in is_load_mux_reducible()
**Recreate**
_make_

**Output**

jlc: jlm/llvm/ir/operators/load.cpp:51: bool jlm::llvm::is_load_mux_reducible(const std::vector<rvsdg::output *> &): Assertion `operands.size() >= 2' failed.

[fbench.zip](https://github.com/phate/jlm/files/13671241/fbench.zip)

### LINKED ISSUE #331: Programs with loads from NULL have undefined behavior, but jlm shouldn't crash while compiling them
Related to #323.

Take the following C program:
```c
static int *refarr;

int
main()
{
    return *refarr;
}
```
This program is not well-formed.
Since `refarr` is static, it will always be `NULL`, and clang knows this.
The produced LLVM IR with -O0 is

<details>
<summary>Click to expand</summary>

```llvm
@refarr = internal global ptr null, align 8

define dso_local i32 @main() {
  %1 = alloca i32, align 4
  store i32 0, ptr %1, align 4
  %2 = load ptr, ptr @refarr, align 8
  %3 = load i32, ptr %2, align 4
  ret i32 %3
}
```
</details>

As can be seen, clang makes up an `alloca` instruction in `main()`, stores 0 to it, and never reads from it.
Compilers are of course allowed to do whatever they like with ill-formed programs, so we can't fault it.

By applying `jlm-opt --AAAndersenAgnostic`, the following RVSDG is produced:
image

The alias analysis correctly figures out that the last load, the load from `refarr`, can't point to anything, which means the load doesn't depend on any memory state. This leads to a load node with a single operand, which is illegal in the `jlm::llvm::is_load_mux_reducible` function.

My question is: should we give loads and stores affecting 0 memory states a dummy memory state edge input, or allow them to have zero memory state edge inputs? (A well-formed program should never have loads or stores like that, but we should at least gracefully handle ill-formed programs, and not just crash).

phate:
The load-mux reduction was unable to handle loads without state edges. This can happen when the compiler can statically determine that the address of the load is NULL. Even though this is an undefined program, the compiler should not panic and throw a hard error. This PR fixes the reduction and adds a unit test for this case.

Closes #331
Closes #323

### DIFF
--- a/jlm/llvm/ir/operators/load.cpp
+++ b/jlm/llvm/ir/operators/load.cpp
@@ -59,7 +59,7 @@ is_load_mux_reducible(const std::vector<rvsdg::output *> & operands)
   auto memStateMergeNode = rvsdg::node_output::node(operands[1]);
   if (!is<MemStateMergeOperator>(memStateMergeNode))
     return false;
-  
+
   return true;
 }
 
--- a/jlm/llvm/ir/operators/load.cpp
+++ b/jlm/llvm/ir/operators/load.cpp
@@ -48,15 +48,18 @@ LoadOperation::copy() const
 static bool
 is_load_mux_reducible(const std::vector<rvsdg::output *> & operands)
 {
-  JLM_ASSERT(operands.size() >= 2);
-
-  auto memStateMergeNode = rvsdg::node_output::node(operands[1]);
-  if (!is<MemStateMergeOperator>(memStateMergeNode))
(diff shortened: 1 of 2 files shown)


haved (approved):
Looks great, very readable tests!

PR #340, opened 2024-01-05
Make Andersen emit PointsToGraphs with RegisterSetNodes

haved:
Instead of emitting one `PointsToGraph::RegisterNode` per `rvsdg::output*` represented in the PointsToGraph, it groups all `rvsdg::output*`s associated with the same `PointerObject`, and creates one `RegisterSetNode` for them. This should enable the future removal of `RegisterNode`s.

While translating the tests, I observed that the number of `RegisterSetNode`s was about 50-90% of the amount of `RegisterNode`s the test's PointsToGraph used to have, a nice improvement.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/Andersen.cpp
+++ b/jlm/llvm/opt/alias-analyses/Andersen.cpp
@@ -430,7 +430,7 @@ Andersen::AnalyzePhi(const phi::node & phi)
 
     // Map the output register to the recursion result's pointer object
     auto & outputRegister = *rv;
-    Set_->MapRegisterToExistingPointerObject(outputRegister, argumentRegisterPO);
+    Set_->MapRegisterToExistingPointerObject(outputRegister, resultRegisterPO);
   }
 }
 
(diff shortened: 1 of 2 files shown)


jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -430,7 +430,7 @@ Andersen::AnalyzePhi(const phi::node & phi)
 
     // Map the output register to the recursion result's pointer object
     auto & outputRegister = *rv;
-    Set_->MapRegisterToExistingPointerObject(outputRegister, argumentRegisterPO);
+    Set_->MapRegisterToExistingPointerObject(outputRegister, resultRegisterPO);
```
haved, line 433: This modification would not affect the PointsToGraph previously, but now it actually makes a difference if the phi node's output is mapped to the result or the argument. (As you can see, the comment was previously a lie, now the comment is true)

tests/jlm/llvm/opt/alias-analyses/TestAndersen.cpp
```diff
@@ -8,12 +8,12 @@
...
 #include <jlm/llvm/opt/alias-analyses/Andersen.hpp>
-#include <jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp>
 #include <jlm/llvm/opt/alias-analyses/PointsToGraph.hpp>
-#include <jlm/rvsdg/node.hpp>
-#include <jlm/rvsdg/view.hpp>
 #include <jlm/util/Statistics.hpp>
 
+// TODO: Remove
+#include <jlm/rvsdg/view.hpp>
+
```
phate, line 16: This should be removed again, I guess?

phate (approved):
LGTM

tests/jlm/llvm/opt/alias-analyses/TestAndersen.cpp
haved: Darn, of course

PR #343, opened 2024-01-11
Let Andersen track pointees in aggregate types

haved:
Makes the Andersen analysis attach `PointerObject`s and track points-to sets of all registers that are either of type `PointerType`, or aggregate types where one of the subtypes is `PointerType`.

This means we no longer need to expose pointers that enter structs, and re-synthesize them at getelementptr (by pointing to external), which gives better analysis.

Also tightens the asserts in Andersen a bit, requiring all nodes with pointer type inputs and outputs to be properly handled. This made it clear that handlers for `valist_op` and `ptr2bits_op` were missing. Their logic is simple enough, so they were added.


jlm/llvm/ir/types.hpp
```diff
@@ -626,6 +626,18 @@ IsOrContains(const jlm::rvsdg::type & type)
   return false;
 }
 
+/**
+ * Given a type, determines if it is one of LLVM's aggregate types.
+ * Vectors are not considered to be aggregate types, despite being based on a subtype.
+ * @param type the type to check
+ * @return true if the type is an aggreate type, false otherwise
+ */
+static inline bool
```
phate, line 635: We had this discussion before. I believe we concluded that `static inline` makes no sense. It should only be `inline`, right?

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -47,13 +61,21 @@ Andersen::AnalyzeSimpleNode(const rvsdg::simple_node & node)
...
   else
   {
-    // TODO: Make sure all simple nodes involving pointers are correctly handled by the analysis
+    // This node operation is unknown, make sure it doesn't consume any pointers
+    for (size_t i = 0; i < node.ninputs(); i++)
+    {
+      if (IsTypePointing(node.input(i)->type()))
+        std::cerr << node.operation().debug_string() << std::endl;
+      JLM_ASSERT(!IsTypePointing(node.input(i)->type()));
+    }
```
phate, line 78: You only check the inputs. Why not the outputs of the node?

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -12,6 +12,18 @@
...
 {
 
+/**
+ * If values of the given type may point to things, the set of pointees needs to be tracked.
+ * If the type is an aggregate type, pointees must be tracked if any subtype is pointing.
+ * @param type the rvsdg type to be checked
+ * @return true if pointees should be tracked for all values of the given type, otherwise false
+ */
+bool
+IsTypePointing(const rvsdg::type & type)
```
phate, line 22: Do you think a better name would be: `IsOrContainsPointerType()`? I had to read the documentation multiple times before I fully understood how the code in the body relates to the name.

jlm/llvm/opt/alias-analyses/Andersen.cpp
phate: I just saw that you also added the same checks for the arguments and results of a region. In combination with this check, it should then cover everything. Ignore the above comment, please.

phate (approved):
Nothing major, just some minor suggestions.

jlm/llvm/ir/types.hpp
haved: True, I switched back to CLion (From CLion Nova), and it doesn't highlight it, so I'll have to remember manually

jlm/llvm/opt/alias-analyses/Andersen.cpp
haved: That would be a better name, I'll change it

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -47,13 +61,21 @@ Andersen::AnalyzeSimpleNode(const rvsdg::simple_node & node)
...
   else
   {
-    // TODO: Make sure all simple nodes involving pointers are correctly handled by the analysis
+    // This node operation is unknown, make sure it doesn't consume any pointers
+    for (size_t i = 0; i < node.ninputs(); i++)
+    {
+      if (IsOrContainsPointerType(node.input(i)->type()))
+        std::cerr << node.operation().debug_string() << std::endl;
+      JLM_ASSERT(!IsOrContainsPointerType(node.input(i)->type()));
+    }
```
phate, line 78: This will actually result in some cost in release build. The JLM_ASSERT is only enabled in a debug build (-DJLM_ENABLE_ASSERT), but not in a release build. If you also print the debug string, then the loop does not only contain noops, and the compiler won't be able to optimize it away. Thus, you will actually iterate through all inputs for every single node that is not handled by the analysis (integer operations, float operations, etc.) in optimized code. This is a high price to pay for the convenience of the operation name that you could also get by firing up the debugger (when it is needed).

phate (changes requested):
I am sorry to send it back, but I think we should sort this out.

jlm/llvm/opt/alias-analyses/Andersen.cpp
phate: I am sorry, but I did not see this the first time looking at the PR.

haved: I completely understand, good thing you caught it. It was only there to quickly get the missing operations, and should have had `TODO Remove` plastered all over it.

phate (approved):

PR #344, opened 2024-01-13
-Wno-error=comment is needed when compiling HLS with g++

sjalander:
The MLIR headers that the HLS backend depends on have comments with three slashes '///', which cause g++ to throw an error unless -Wno-error=comment is used.

This case is triggered if jlm is configured to use g++, i.e., ./configure.sh --circt-path ../circt/local


configure.sh
```diff
@@ -85,7 +85,7 @@ else
 fi
 
 if [ "${CIRCT_ENABLED}" == "yes" ] ; then
-	CPPFLAGS_CIRCT="-DCIRCT=1 -I${CIRCT_PATH}/include"
```
phate, line 88: If I am not mistaken, then the `-Wno-error=comment` should go into `CXXFLAGS_CIRCT` and not `CPPFLAGS_CIRCT`.

phate (changes requested):

configure.sh
phate: Also, why was this not caught by the CI? Is it because we only compile HLS backend with clang?

sjalander: That's the case.

sjalander: Changed to CXXFLAGS_CIRCT

phate (approved):

PR #345, opened 2024-01-13
Avoid building HLS backend without CIRCT

sjalander:
The HLS backend depends on CIRCT and this change excludes building the backend unless the build has been configured by specifying the --circt-path


sjalander:
Also moved jhls and jlm-hls into separate Makefile.sub, such that they can also be conditionally built based on the use of --circt-path

phate:
@caleridas I will leave the review to you.

caleridas:
It might be sensible to structure this as an "--enable-hls" or so switch to configure.sh (rather than an implicit --with-circt-path). But more a matter of cosmetics, not really of significance.

caleridas (approved):

PR #346, opened 2024-01-14
Add mandatory doxygen workflow

phate:
This PR does the following:
1. Fixes all warnings and errors that doxygen produces when creating the documentation
2. Configures doxygen to fail when it encounters an error or warning
3. Moves documentation creation into its own workflow

### DIFF
--- a/.github/workflows/Doxygen.yml
+++ b/.github/workflows/Doxygen.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Install dependencies
-        run: sudo apt-get install graphviz doxygen
+        run: sudo apt-get install graphviz doxygen llvm-16-dev
       - name: Configure jlm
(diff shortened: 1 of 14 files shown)


haved (approved):
Looks very nice, I assume all changes were done to appease the doxygen warnings, even if they do look a bit larger in certain files.

Not that it is relevant to this PR, but would it make sense to host the generated documentation for the master branch somewhere? With a web based viewer as well it's about time we need a website :)

phate:
Yes, all changes were done to appease doxygen. The 19.1 version of doxygen was not able to handle the template stuff in InterProceduralGraphConversion.cpp and Annotation.cpp. My local version (19.4) was able to handle it, but it was easier to rewrite/simplify the code than to compile doxygen ourselves.

Yes, hosting the documentation somewhere would be the next logical step. I already had it in mind, but it is not high on my priority list right now.

PR #350, opened 2024-01-17
Propagate points-to flags between sets in Steensgaard

### LINKED ISSUE #335: Compiling 525.x264 with --AASteensgaardAgnostic results in error
When compiling `ldecod_src/img_process.c`, the compilation throws the following error:

```
jlc -c -o ldecod_src/img_process.o -DSPEC -DNDEBUG -Ildecod_src/inc -Ix264_src -Ix264_src/extras -Ix264_src/common -DSPEC_AUTO_SUPPRESS_OPENMP -DSPEC_AUTO_BYTEORDER=0x12345678   -JAASteensgaardAgnostic --JlmOptPassStatistics=print-steensgaard-analysis --JlmOptPassStatistics=print-memory-node-provisioning --JlmOptPassStatistics=print-basicencoder-encoding --JlmOptPassStatistics=print-dne-stat              -DSPEC_LP64 -fcommon  ldecod_src/img_process.c
terminate called after throwing an instance of 'jlm::util::error'                                      
  what():  Number of memory states cannot be zero.                                                     
specmake: *** [/home/reissman/Documents/spec-cpu2017-jlm/benchspec/Makefile.defaults:347: ldecod_src/img_process.o] Aborted (core dumped) 
specmake: *** Waiting for unfinished jobs....
```

The issue has potentially the same cause as #334.

phate:
This PR introduces a points-to flag propagation pass in the Steensgaard alias analysis such that the points-to flags are properly propagated in the disjoint location set graph. So far, the points-to flags were only annotated for each disjoint location set, but were not propagated to the points-to set. However, all of the points-to flags are contagious, i.e., if a set is marked with a flag, then the points-to set must also be marked with it.

This should put the points-to flags on solid footing as:
1. The points-to flags are added for each location on creation time.
2. The flags are propagated to the root element of a set when a location is joined to a set.
3. The flags are propagated between sets with the just introduced propagation pass.

This PR introduces the third point from above. The first and second point are already present in the code.

Closes #335

### DIFF
--- a/jlm/llvm/opt/alias-analyses/Steensgaard.cpp
+++ b/jlm/llvm/opt/alias-analyses/Steensgaard.cpp
@@ -813,10 +813,22 @@ public:
   }
 
   void
+  StartPointsToFlagsPropagationStatistics(const LocationSet & disjointLocationSet) noexcept
+  {
+    NumDisjointSets_ = disjointLocationSet.NumDisjointSets();
+    NumLocations_ = disjointLocationSet.NumLocations();
+    PointsToFlagsPropagationTimer_.start();
+  }
+
+  void
+  StopPointsToFlagsPropagationStatistics() noexcept
+  {
+    PointsToFlagsPropagationTimer_.stop();
+  }
+
+  void
   StartPointsToGraphConstructionStatistics(const LocationSet & locationSet)
   {
-    NumDisjointSets_ = locationSet.NumDisjointSets();
-    NumLocations_ = locationSet.NumLocations();
     PointsToGraphConstructionTimer_.start();
   }
 
@@ -893,6 +905,9 @@ public:
         "#UnknownMemorySources:",
         NumUnknownMemorySources_,
         " ",
+        "PointsToFlagsPropagationTime[ns]:",
+        PointsToFlagsPropagationTimer_.ns(),
+        " ",
         "PointsToGraphConstructionTime[ns]:",
         PointsToGraphConstructionTimer_.ns(),
         " ",
@@ -924,6 +939,7 @@ private:
   size_t NumUnknownMemorySources_;
 
   util::timer AnalysisTimer_;
+  util::timer PointsToFlagsPropagationTimer_;
   util::timer PointsToGraphConstructionTimer_;
   util::timer UnknownMemoryNodeSourcesRedirectionTimer_;
 };
@@ -1714,6 +1730,11 @@ Steensgaard::Analyze(
   // std::cout << LocationSet_->ToDot() << std::flush;
   statistics->StopSteensgaardStatistics();
 
+  // Propagate points-to flags in disjoint location set
+  statistics->StartPointsToFlagsPropagationStatistics(*LocationSet_);
+  PropagatePointsToFlags();
+  statistics->StopPointsToFlagsPropagationStatistics();
+
   // Construct PointsTo graph
   statistics->StartPointsToGraphConstructionStatistics(*LocationSet_);
   auto pointsToGraph = ConstructPointsToGraph();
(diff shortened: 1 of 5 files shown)


jlm/llvm/opt/alias-analyses/Steensgaard.cpp
```diff
@@ -1733,6 +1754,37 @@ Steensgaard::Analyze(
...
+      auto locationFlags = location->GetPointsToFlags();
+      auto pointsToLocationFlags = pointsToLocation.GetPointsToFlags();
+      auto combinedFlags = locationFlags | pointsToLocationFlags;
+      if (pointsToLocationFlags != combinedFlags)
+      {
+        pointsToLocation.SetPointsToFlags(combinedFlags);
+        pointsToFlagsChanged = true;
+      }
+    }
+  } while (pointsToFlagsChanged);
```
phate, line 1785: This algorithm is a simple fix-point computation. It behaves quadratic in the number of disjoint location sets. I can foresee that its runtime might become a problem in the future. However, I currently care more about correctness than performance. Once the data shows that it is a problem, I will tackle it. Thus, please ignore its performance aspect right now.

haved:
I might be missing something, but I'm not seeing a change in this PR that copies flags to the root of disjoint sets during unions. Also I'm a bit curious how the "PointsToNone" flag works. Does it indicate that there is no pointee location, because that seems a bit incompatible with all the other flags.

Regarding the propagation, it seems like locations escaping is being encoded by some other location pointing to it having the "pointsToEscaped" flag. This, combined with this PR, effectively gives us escape propagation, but I'm unsure if any of the other flags need propagation this way. Again PointsToNone comes to mind, but I might be misunderstanding it entirely.

phate:
Thanks for your feedback. Let me address your concerns individually:

> I might be missing something, but I'm not seeing a change in this PR that copies flags to the root of disjoint sets during unions.

The copying of flags to the root of the disjoint sets was already there and works nicely. It is part of the [Join()](https://github.com/phate/jlm/blob/7ad640454dbb91c82ed60471d76adc64c6d1e00f/jlm/llvm/opt/alias-analyses/Steensgaard.cpp#L714C1-L742C4) method. The two root elements are set to the combination of both flags and the consecutive Merge() will give back one of them as new root of the combined set.

> Also I'm a bit curious how the "PointsToNone" flag works. Does it indicate that there is no pointee location, because that seems a bit incompatible with all the other flags.

Yes, it should indicate that there is no pointee location, and yes, it is incompatible with the other flags. I also realized while working on this PR that this is "broken". A set can have a combination of flags such as `PointsToNone | PointsToEscapedMemory | ....`, i.e., pointing nowhere while at the same time pointing to escaped memory, which does not make sense. I need to fix this, but that is for another PR. It currently works, because the PointsToNone is nowhere checked after the analysis, i.e., in propagation or points-to graph construction. We only care about the other flags. Thus, the easiest fix for this (I believe) would be to simply remove the `PointsToNone` flag, which makes me happy as it reduces complexity.

> Regarding the propagation, it seems like locations escaping is being encoded by some other location pointing to it having the "pointsToEscaped" flag. This, combined with this PR, effectively gives us escape propagation, but I'm unsure if any of the other flags need propagation this way. Again PointsToNone comes to mind, but I might be misunderstanding it entirely.

Yes, effectively it is an escape analysis. The other flags (PointsToExternal and PointsToUnknown) will eventually disappear as we discussed previously. There are two parts to this analysis. One is the part above where we track which locations point to escaped memory. The second part is the annotation of registers that are [escaping the module](https://github.com/phate/jlm/blob/7ad640454dbb91c82ed60471d76adc64c6d1e00f/jlm/llvm/opt/alias-analyses/Steensgaard.cpp#L148C1-L163C4). Those are then used in the [CollectEscapedMemoryNodes()](https://github.com/phate/jlm/blob/7ad640454dbb91c82ed60471d76adc64c6d1e00f/jlm/llvm/opt/alias-analyses/Steensgaard.cpp#L1757C1-L1806C2) method as starting point to determine which memory locations escaped. Finally, in the points-to graph construction phase, edges are inserted in the points-to graph from the registers that have the PointsToEscapedMemory flag to all memory locations that are found in the CollectEscapedMemoryNodes() method.

haved (approved):
Looks good, your comment answered all my questions, and it's always nice to improve correctness

PR #354, opened 2024-01-18
Add Statistics tracking to Andersen

### LINKED ISSUE #352: Add Andersen alias analysis pass statistics
The Andersen alias analysis pass [declares](https://github.com/phate/jlm/blob/f0ba9b5fec9f0bddb0a0f086230676c9766c8401/jlm/llvm/opt/alias-analyses/Andersen.hpp#L25C1-L25C20) a Statistics class, but never provides an implementation for it. I wanted to incorporate the Andersen alias analysis into the memory state encoding investigation, but in order to do so, it needs to produce some statistics.

haved:
Fixes #352

The tracked statistics have the same names as Steensgaard where applicable, but there are of course some differences.

`NumUnknownMemorySources_` became `NumExternalMemorySources_`, since Andersen only uses external.

Note that time spent connecting all PointingToExternal nodes to all Escaped nodes, is called `PointsToGraphConstructionExternalToEscapedTime[ns]`, since it is a subslice of  `PointsToGraphConstructionTime[ns]`.

The papers I've been reading about optimizing Andersen can have quite different internal representations for sets and set constraints, which means some of these statistics might change, but I will of course try to keep things stable when possible, to make comparisons possible.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/Andersen.cpp
+++ b/jlm/llvm/opt/alias-analyses/Andersen.cpp
@@ -43,7 +43,7 @@ public:
         NumSupersetOfAllPointeesConstraints_(0),
         NumHandleEscapingFunctionConstraints_(0),
         NumFunctionCallConstraints_(0),
-        ConstraintSolvingIterations_(0),
+        NumConstraintSolvingIterations_(0),
         NumPointsToGraphNodes_(0),
         NumRegisterNodes_(0),
         NumAllocaNodes_(0),
@@ -97,13 +97,15 @@ public:
   }
 
   void
-  StartPointsToGraphConstructionStatistics(
-      const PointerObjectSet & set,
-      size_t constraintSolvingIterations)
+  StopConstraintSolvingStatistics(size_t numConstraintSolvingIterations) noexcept
   {
     ConstraintSolvingTimer_.stop();
-    ConstraintSolvingIterations_ = constraintSolvingIterations;
+    NumConstraintSolvingIterations_ = numConstraintSolvingIterations;
+  }
 
+  void
+  StartPointsToGraphConstructionStatistics()
+  {
     PointsToGraphConstructionTimer_.start();
   }
 
@@ -176,7 +178,7 @@ public:
         NumFunctionCallConstraints_,
         " ",
         "#ConstraintSolvingIterations:",
-        ConstraintSolvingIterations_,
+        NumConstraintSolvingIterations_,
         " ",
         "#PointsToGraphNodes:",
         NumPointsToGraphNodes_,
@@ -243,7 +245,7 @@ private:
   size_t NumFunctionCallConstraints_;
 
   // How many iterations constraint solving used before a fixed point was established
-  size_t ConstraintSolvingIterations_;
+  size_t NumConstraintSolvingIterations_;
 
   // Counts of nodes in the final PointsToGraph
   size_t NumPointsToGraphNodes_;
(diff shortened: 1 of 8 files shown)


jlm/tooling/CommandLine.hpp
```diff
@@ -177,6 +177,7 @@ class JlmOptCommandLineOptions final : public CommandLineOptions
   struct StatisticsCommandLineArgument
   {
     inline static const char * Aggregation_ = "print-aggregation-time";
+    inline static const char * AndersenAnalysis_ = "print-andersen-analysis";
```
phate, line 180: Let's just all it "AndersenAliasAnalysis" or "AndersenAnalysis". All these command line arguments need overhauling to something more sane.

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -443,14 +443,23 @@ PointerObjectConstraintSet::AddConstraint(ConstraintVariant c)
...
+const std::vector<PointerObjectConstraintSet::ConstraintVariant> &
+PointerObjectConstraintSet::GetConstraints() const noexcept
+{
+  return Constraints_;
+}
+
+size_t
 PointerObjectConstraintSet::Solve()
 {
+  size_t iterations = 0;
```
phate, line 455: `numIterations`

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -24,6 +24,254 @@ IsOrContainsPointerType(const rvsdg::type & type)
   return IsOrContains<PointerType>(type);
 }
 
+/**
+ * Class collecting statistics from a pass of Andersen's alias analysis
+ */
+class Andersen::Statistics final : public jlm::util::Statistics
```
phate, line 30: I got a little bit confused that you did not have explicit Stop*() methods for every Start*() method. It took me a while to understand what was going on.

phate (approved):
LGTM. Only found some minor suggestions.

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -24,6 +24,254 @@ IsOrContainsPointerType(const rvsdg::type & type)
   return IsOrContains<PointerType>(type);
 }
 
+/**
+ * Class collecting statistics from a pass of Andersen's alias analysis
```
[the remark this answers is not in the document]
haved: Depends on how you look at it. I think of a "pass" as arriving at a full set constraint solution and creating the PointsToGraph.

jlm/llvm/opt/alias-analyses/Andersen.cpp
haved: It looks better to have corresponding start and stop functions, so I changed it

jlm/tooling/CommandLine.hpp
haved: I'm sorry, I don't quite understand what you are requesting here. Do you want me to remove the trailing underscore? I can add the word "Alias" of course, but your comment makes it seem like that is not what you are after?

phate: The command line flag is currently called "print-andersen-analysis". I would like to move away from this unhandy schema, and rather just call it by its name: "AndersenAliasAnalysis" or "AndersenAnalysis":

```
inline static const char * AndersenAnalysis_ = "AndersenAliasAnalysis";
```

haved: But then the command line flag for enabling statistics collection would be `--AndersenAliasAnalysis`, which looks a bit confusing, and is different from all other statistics enabling flags?

Is it not better left for another PR that:
 - changes the statistics enabling flags in a consistent manner, possibly moving to a
`--statistics=Comma,Separated,List` format?
 - unifies the list between `JlmOptCommand` and `JlcCommand`, avoiding all the duplication between them?

phate: You are absolutely right. I created an issue for it: #357 . The list between jlc and jlm-opt should already be unified and in sync (TM).

PR #355, opened 2024-01-18
Add --enable-asserts to configure.sh

haved:
The set of changes is now:

Remove `CLANG_BIN` and `-DJLM_DEBUG` as they were unused.

Remove `-fPIC` from the common flags, reserving it for only `*.lo` build outputs.

Add `--enable-asserts` to control asserts, by default asserts are disabled.

Add the `--enable-asserts` flag to the GitHub workflows for clang and gcc builds.


caleridas:
I am not even certain whether the concept of "non-optimized builds" is terrible sensible, might as well add optimization flags to the debug builds. For the extremely rare cases of debugger getting "too confused" can always still hack the makefile.
Also, could possibly unconditionally enable generation of debug symbols. It slightly slows compilation and increases size of built binaries -- the latter is then usually handled by stripping the binaries (and/or splitting debug symbols into separate fails) when packaging the bulid artifacts for distribution.

phate:
Thanks for starting the discussion. You raise a lot of good points. Let me try to address them individually:

> When we run tests in GitHub workflows, we should enable jlm asserts, but we still want to have an optimized build to avoid llvm-eval-suite taking even longer than it already does (18 minutes).

I fully agree. I was waiting for the rework of the makefiles PR to get merged such that I could tackle "build" workflows. We are currently building jlm several times:

1. Once with clang in release
2. Once with gcc in release
3. Once with HLS and clang
4. Soon we will have another build with MLIR and clang(?)

However, we also have several desired builds missing, such as a build with asserts or a debug build. I wanted to rework this, cut down on the number of builds, and restructure all of this a bit.

> The new release-safe target is pretty much JLM_ENABLE_ASSERTS -O3, and with no debug info.

I think that this is the wrong way to go about it. I would rather introduce an `--enable-asserts` flag that would add `JLM_ENABLE_ASSERTS` to the build flags. This way we can happily combine it with "debug" or "release" target. Maybe it would also make sense to actually remove the "debug" and "release" targets, and rather introduce an "--enable-optimizations" flag, but I think that is up for a later discussion.

> I was unsure what to do about JLM_DEBUG, but as far as I can tell, it is never referenced from anywhere, so it shouldn't matter.

You are absolutely right. I was not aware of this. We originally had `JLM_DEBUG_ASSERT` and this macro was controlled by the `JLM_DEBUG`, which was only added in a debug build. I decoupled the "assert functionality" at some point from the debug build and introduced `JLM_ENABLE_ASSERTS`. I think we can remove the `JLM_DEBUG` macro, as it is not used anywhere any longer. It is always possible to (re-)introduce it when it is needed.

 > I also moved -gdwarf-4 -g down to debug only, unless we want debug info in release builds for some reason.
 
 The `-gdwarf-4` was at some point necessary in order to get valgrind working properly. It was introduced in this [commit](https://github.com/phate/jlm/commit/ae2d9963516e176d3820ce7331e26809fd46b542). I just had a look at the valgrind job output from this PR, and even though jlm was compiled without `-gdwarf-4`, valgrind seem to do its job. Thus, it seems that it is not longer necessary. Regarding `-g`, I am with @caleridas  on this one. It does not hurt to have it and can be stripped at the end if necessary.
 
 > I also noticed that CLANG_BIN isn't referenced from anywhere, do we need it?
 
 I believe that it was originally used [here](https://github.com/phate/jlm/blob/7109418c2cfc4c52fdc247b4353fd41c193189b4/jlm/tooling/Makefile.sub#L18C1-L18C94). It does not seem to be used any longer, so let's just remove it.

haved:
I tried to change things closer to what you suggested. When I removed `-gdwarf-4`, valgrind started complaining, so I added it back. (We can either use `-gdwarf-4 -g`, or no debug info. The issues occur when we use only `-g`)

configure.sh
```diff
@@ -66,31 +72,36 @@ while [[ "$#" -ge 1 ]] ; do
 done
 
 
-CXXFLAGS_COMMON="--std=c++17 -Wall -Wpedantic -Wextra -Wno-unused-parameter -Werror -Wfatal-errors -gdwarf-4 -g -fPIC"
-
+CXXFLAGS_COMMON="--std=c++17 -Wall -Wpedantic -Wextra -Wno-unused-parameter -Werror -Wfatal-errors -fPIC -gdwarf-4 -g"
```
phate, line 75: This is just noise. Nothing really changed aside the order of the flags.

configure.sh
```diff
@@ -66,31 +72,36 @@ while [[ "$#" -ge 1 ]] ; do
...
 
-CPPFLAGS_CIRCT=""
-
 if [ "${TARGET}" == "release" ] ; then
 	CXXFLAGS_TARGET="-O3"
+	CPPFLAGS_TARGET=""
 elif [ "${TARGET}" == "debug" ] ; then
 	CXXFLAGS_TARGET="-O0"
-	CPPFLAGS_TARGET="-DJLM_DEBUG -DJLM_ENABLE_ASSERTS"
+	CPPFLAGS_TARGET=""
```
phate, line 85: CPPFLAGS_TARGET can be removed. It is always empty, no?

configure.sh
```diff
@@ -66,31 +72,36 @@ while [[ "$#" -ge 1 ]] ; do
...
+	CPPFLAGS_EXTRA="${CPPFLAGS_EXTRA} -DJLM_ENABLE_ASSERTS"
+fi
+
+CXXFLAGS_CIRCT=""
+CPPFLAGS_CIRCT=""
+
 if [ "${CIRCT_ENABLED}" == "yes" ] ; then
-	CPPFLAGS_CIRCT="-I${CIRCT_PATH}/include"
 	CXXFLAGS_CIRCT="-Wno-error=comment"
+	CPPFLAGS_CIRCT="-I${CIRCT_PATH}/include"
```
phate, line 102: This is also just noise. It only changed the order.

configure.sh
```diff
@@ -66,31 +72,36 @@ while [[ "$#" -ge 1 ]] ; do
...
 else
 	echo "No build type set. Please select either 'debug' or 'release'." >&2
 	exit 1
 fi
 
+CPPFLAGS_EXTRA=""
+
+if [ "${ENABLE_ASSERTS}" == "yes" ] ; then
+	CPPFLAGS_EXTRA="${CPPFLAGS_EXTRA} -DJLM_ENABLE_ASSERTS"
+fi
```
phate, line 95: What do you think about `CPPFLAGS_ASSERTS` as name?

configure.sh
```diff
@@ -33,6 +35,10 @@ while [[ "$#" -ge 1 ]] ; do
 			TARGET="$1"
 			shift
 			;;
+		--enable-asserts)
+			ENABLE_ASSERTS="yes"
+			shift
+		;;
```
phate, line 41: Indentation. All the other ones have the ;; on the same level as the shift.

phate (changes requested):

caleridas (approved):

haved:
I will change things to remove PR noise, as you say, @phate.

I was a bit unsure about the"policy" regarding using undefined variables, since there currently is a bit of a mix between assigning `""` to variables as a default, and just using them without defining them and relying on bash expanding undefined variables to the empty string. If we agree that the latter is fine, then that definitely saves us some `CXXFLAGS_xxx=""`

Finally, I was wondering if the `-fPIC` flag is supposed to be one of the common flags, since `Makefile.rules` seems to indicate that the only difference between `*.la` and `*.lo` is the addition of that flag, which is already enabled for both as far as I can tell? I never remember quite which flags are needed for shared libraries, do you remember what the difference between `*.la` and `*.lo` is supposed to be, @caleridas?

caleridas:
> I will change things to remove PR noise, as you say, @phate.
> 
> I was a bit unsure about the"policy" regarding using undefined variables, since there currently is a bit of a mix between assigning `""` to variables as a default, and just using them without defining them and relying on bash expanding undefined variables to the empty string. If we agree that the latter is fine, then that definitely saves us some `CXXFLAGS_xxx=""`

usually I initialize shell scripts with "set -eu" such that undefined variables cause an error -- that catches typos. I forgot it in this script, but stuck with the habit of being pedantic with variables

> 
> Finally, I was wondering if the `-fPIC` flag is supposed to be one of the common flags, since `Makefile.rules` seems to indicate that the only difference between `*.la` and `*.lo` is the addition of that flag, which is already enabled for both as far as I can tell? I never remember quite which flags are needed for shared libraries, do you remember what the difference between `*.la` and `*.lo` is supposed to be, @caleridas?

-fPIC should only be for .lo, when building shared libraries. I think presently nobody cares about shared library builds, but I am going to use it. Globally, -fPIC can be removed.

haved:
I think the PR is now at a sensible set of changes, with less noise. It does however add `set -eu` to minimize the risk of typos causing sneaky errors.

We can of course discuss if asserts should be enabled or disabled by default.

configure.sh
```diff
@@ -66,31 +73,34 @@ while [[ "$#" -ge 1 ]] ; do
...
-CXXFLAGS_COMMON="--std=c++17 -Wall -Wpedantic -Wextra -Wno-unused-parameter -Werror -Wfatal-errors -gdwarf-4 -g -fPIC"
-
+CXXFLAGS_COMMON="--std=c++17 -Wall -Wpedantic -Wextra -Wno-unused-parameter -Werror -Wfatal-errors -gdwarf-4 -g"
 CPPFLAGS_COMMON="-I. -Itests"
 
 CPPFLAGS_LLVM=$(${LLVM_CONFIG_BIN} --cflags)
 
-CPPFLAGS_CIRCT=""
-
+CPPFLAGS_TARGET=""
```
phate, line 81: `CPPFLAGS_TARGET` is always empty. It is never set to something. I think it can be removed completely.

phate (approved):
I only found a minor issue. Great work, I am looking forward to get these changes in. However, @caleridas I would like you to approve it as well.

haved:
Fixed your suggestion, @phate. It is up to you if you want this in before or after #358.

PR #356, opened 2024-01-20
Clean up common_executable build macro

### LINKED ISSUE #353: CPPFLAGS and LDFLAGS naming mismatch
I believe there is a discrepancy about the naming of the CPPFLAGS attribute for the individual tools in the project.

The CPPFLAGS attribute for jlc is called [jlc_EXTRA_CPPFLAGS](https://github.com/phate/jlm/blob/f0ba9b5fec9f0bddb0a0f086230676c9766c8401/tools/Makefile.sub#L14), for jhls it is called [jhls_EXTRA_CPPFLAGS](https://github.com/phate/jlm/blob/f0ba9b5fec9f0bddb0a0f086230676c9766c8401/tools/jhls/Makefile.sub#L15C1-L15C60), for jlm-hls it is called [jlm-hls_EXTRA_CPPFLAGS](https://github.com/phate/jlm/blob/f0ba9b5fec9f0bddb0a0f086230676c9766c8401/tools/jlm-hls/Makefile.sub#L19C1-L20C34), but for jlm-opt it is called [jlm-opt_CPPFLAGS](https://github.com/phate/jlm/blob/f0ba9b5fec9f0bddb0a0f086230676c9766c8401/tools/Makefile.sub#L31C1-L31C17). The [help for the common_executable macro](https://github.com/phate/jlm/blob/f0ba9b5fec9f0bddb0a0f086230676c9766c8401/Makefile.macros#L94C1-L107C73) does not mention either one of them, even though it mentions the EXTRA_LDFLAGS.

A similar issue seems to be going on for the LDFLAGS for common_executable. The help mentions EXTRA_LDFLAGS explicitly, but none of the tools uses this attribute, but rather LDFLAGS.

I am not fully sure what was exactly meant as the correct way of setting up these macros. My suspicion is that everything should be EXTRA_* and that EXTRA_CPPFLAGS needs to be added to the help, but I am not certain.

phate:
This PR does the following:

1. Removes all extra CPPFLAGS from the executables in the project. These extra CPPFLAGS were never handled in the common_executable macro and are not necessary for a successful compilation.
2. Ensure consistent naming of EXTRA_LDFLAGS of the common_executable macro.

Closes #353

### DIFF
--- a/Makefile.macros
+++ b/Makefile.macros
@@ -102,7 +102,7 @@ endef
 #   - foo_LIBS:
 #     extra libraries built in this project linked into the binary
 #     (just specify the name without extension)
-#   - foo_EXTRA_LDFLAGS:
+#   - foo_LDFLAGS:
 #     additional LDFLAGS to be passed to link the executable
 #     use this to pull in external libraries (not built in this project)
 define common_executable
--- a/Makefile.macros
+++ b/Makefile.macros
@@ -102,7 +102,7 @@ endef
 #   - foo_LIBS:
 #     extra libraries built in this project linked into the binary
 #     (just specify the name without extension)
-#   - foo_LDFLAGS:
+#   - foo_EXTRA_LDFLAGS:
 #     additional LDFLAGS to be passed to link the executable
 #     use this to pull in external libraries (not built in this project)
 define common_executable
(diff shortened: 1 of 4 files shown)


Makefile.macros
```diff
@@ -102,7 +102,7 @@ endef
 #   - foo_LIBS:
 #     extra libraries built in this project linked into the binary
 #     (just specify the name without extension)
-#   - foo_EXTRA_LDFLAGS:
+#   - foo_LDFLAGS:
```
caleridas, line 105: apologies that I made this inconsistent, but the original reasoning was something like this:
- there may be a global LDFLAGS, and it might at times be necassary to inject something into it from outside the build system for special builds; these might then need to be honored for all targets
- at the same time, binaries and tests might link to external libraries -- they might not wish to "override" the global LDFLAGS, but amend to them; hence the intent as "extra LDFLAGS" ("add to" global LDFLAGS, not "override" global LDFLAGS)
However I admit that this got slightly confused when putting together the macros. Given that there is still TEST_EXTRA_LDFLAGS, maybe they should then at least be made consistent in their meaning? I don't mind dropping the "EXTRA" portion at all, maybe I should not have put that in.

caleridas (approved):

Makefile.macros
phate: Renamed to `*EXTRA_LDFLAGS`. Could you have a second look, please?

caleridas (approved):

PR #358, opened 2024-01-21
Shell script for building CIRCT + CI update

sjalander:
The `build-circt.sh` script clones and builds a compatible version of CIRCT.

The CI is updated to use this build script instead of depending on the jlm-eval-suite repository.

Renamed `--circt-path` to `--enable-hls` in the `configure.sh` script to be more descriptive.

At this point, the path provided to `--enable-hls` has to be absolute, but changes to the HLS tools are planned for another PR that will eliminate this requirement.


.github/workflows/hls.yml
```diff
@@ -8,53 +8,35 @@ on:
...
+    # Compile jlm with the HLS backend enabled
     # --------
 
-    - name: Configure jlm
-      run: ./configure.sh --circt-path $(pwd)/jlm-eval-suite/circt/local CXX=clang++-16
-    - name: Compile jlm
+    - name: "Build CIRCT"
+      uses: ./.github/actions/BuildCirct
+    - name: "Configure jlm"
+      run: ./configure.sh --enable-hls ${{ github.workspace }}/build-circt/circt CXX=clang++-15
```
phate, line 25: `CXX=clang++-16` and not clang++-15

circt.sh
```diff
@@ -0,0 +1,39 @@
...
+}
+
+while [[ "$#" -ge 1 ]] ; do
+        case "$1" in
+                --get-commit-hash)
+                        commit >&2
+                        exit 1
+                        ;;
+        esac
+done
```
phate, line 20: Could we get a `--help` flag similar to the configure.sh script?

circt.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+GIT_COMMIT=093cdfe482530623fea01e1d3242af93e533ba54
+
+CIRCT_BUILD=./build-circt
```
phate, line 5: The script should take the `CIRCT_PATH` as argument and set it. You are using this build path in the CI workflows, but it is implicitly hidden here.

configure.sh
```diff
@@ -14,8 +13,9 @@ function usage()
 	echo "The following options can be set, with defaults specified in brackets:"
 	echo "  --target MODE         Sets the build mode. Supported build modes are"
 	echo "                        'debug' and 'release'. [${TARGET}]"
-	echo "  --circt-path PATH     Sets the path for the CIRCT tools and enables"
-	echo "                        building with CIRCT support. [${CIRCT_PATH}]"
+	echo "  --enable-hls PATH     Sets the path for the CIRCT tools and enables"
+	echo "                        building the HLS backend. [${CIRCT_PATH}]"
+	echo "                        The PATH should be absolut, i.e., not relative."
```
phate, line 18: Why does it need an absolute path? Is there no way to make a path absolute given a relative path?

configure.sh
```diff
@@ -5,7 +5,6 @@ CIRCT_PATH=
 TARGET="release"
 LLVM_CONFIG_BIN="llvm-config-16"
 ENABLE_COVERAGE="no"
-CIRCT_ENABLED="no"
```
phate, line 8: Why did you remove this one? It should rather stay and be renamed `ENABLE_HLS` as that is what it does, no?

configure.sh
```diff
@@ -105,6 +105,7 @@ rm -rf build ; ln -sf build-"${TARGET}" build
 CXXFLAGS=${CXXFLAGS} ${CXXFLAGS_COMMON} ${CXXFLAGS_TARGET} ${CXXFLAGS_CIRCT}
 CPPFLAGS=${CPPFLAGS} ${CPPFLAGS_COMMON} ${CPPFLAGS_TARGET} ${CPPFLAGS_LLVM} ${CPPFLAGS_CIRCT}
 CIRCT_PATH=${CIRCT_PATH}
+CIRCT_ENABLED=${CIRCT_ENABLED}
```
phate, line 108: `HLS_ENABLED`, no? That the HLS backend uses CIRCT is more of a nuance.

circt.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
phate, line 1: We have script folder in the jlm repository. I have the feeling this script would be better suited there.

phate, line 1: The name `circt.sh` is not very descriptive. Maybe `build-circt.sh` as name?

.github/actions/BuildCirct/action.yml
```diff
@@ -4,39 +4,25 @@ description: "Builds CIRCT, which is used for the HLS backend"
...
       with:
         path: |
-          jlm-eval-suite/circt/local
+          ./build-circt/circt
         key: ${{ runner.os }}-circt-${{ steps.get-circt-hash.outputs.hash }}
 
-    # Install dependencies if we didn't hit in the cache
     - name: "Install LLVM, Clang, MLIR, and Ninja"
-      if: steps.cache-circt.outputs.cache-hit != 'true'
       uses: ./.github/actions/InstallLlvmDependencies
```
phate, line 22: Do we need to install LLVM dependencies if we do not need to build anything?

phate (changes requested):

.github/actions/BuildCirct/action.yml
sjalander: Your comment for the MLIR frontend was that I had LLVM followed by MLIR in the actual tests, and that the MLIR build action depended implicitly on LLVM.
This is the same case. There is an implicit dependency, but only if CIRCT Is being built. But the test building CIRCT needs LLVM.

configure.sh
sjalander: Because in the root Makefile there's an "ifdef CIRCT_ENABLED" to decide if to include it in the compilation or not.
With setting it to "no" the variable gets defined and always included in the compilation.

configure.sh
sjalander: The starting point of a relative path becomes from where a command is executed.
So without an absolute path, I had to place the CIRCT build at different locations for the HLS tests to work.

circt.sh
sjalander: Added

circt.sh
sjalander: Added `--build-path` and `--install-path` to the script such that they can be set manually if one want to place it somewhere else than the default location.

configure.sh
sjalander: I decided to separate the enabling of HLS and the path to CIRCT into two flags.
`--enable-hls` is now used for enabling builds with the HLS backend.
`--circt-path` can be used to change the default location of where CIRCT is expected to be (i.e., the default location for scripts/build-circt.sh). An absolute path is still needed, but in most cases one wouldn't have to touch this and the full default path is shown when using `./configure.sh --help`.

configure.sh
sjalander: I changed this in combination with separating `--enable-hls` and `--circt-path`

circt.sh
sjalander: Moved

sjalander: Renamed

sjalander: I found out that we have a scipts folder, so renamed it to scripts.

circt.sh
sjalander: I added explicit paths in the action and workflow. Not sure if that's what you wanted.
The drawback is that any changes to the PATHS in build-circt.sh and configure.sh now requires the CI to also be updated.

configure.sh
sjalander: It's now HLS_ENABLED

phate:
I added @haved on this PR for review. He provided some great input on the other PR that touched the shell scripts.

README.md
```diff
@@ -1,5 +1,7 @@
 # JLM: A research compiler based on the RVSDG IR
 [![Tests](https://github.com/phate/jlm/actions/workflows/tests.yml/badge.svg)](https://github.com/phate/jlm/actions/workflows/tests.yml)
+[![HLS](https://github.com/phate/jlm/actions/workflows/hls.yml/badge.svg)](https://github.com/phate/jlm/actions/workflows/hls.yml)
+[![Doxygen](https://github.com/phate/jlm/actions/workflows/Doxygen.yml/badge.svg)](https://github.com/phate/jlm/actions/workflows/Doxygen.yml)
```
phate, line 4: This is new compared to the last review. Could we not try to do everything all at once, please?

configure.sh
```diff
@@ -14,8 +13,9 @@ function usage()
 	echo "The following options can be set, with defaults specified in brackets:"
 	echo "  --target MODE         Sets the build mode. Supported build modes are"
 	echo "                        'debug' and 'release'. [${TARGET}]"
-	echo "  --circt-path PATH     Sets the path for the CIRCT tools and enables"
-	echo "                        building with CIRCT support. [${CIRCT_PATH}]"
+	echo "  --enable-hls          Enable the HLS backend, which depends on CIRCT."
+	echo "  --circt-path PATH     Sets the path for the CIRCT tools."
+	echo "                        [${CIRCT_PATH}]"
```
phate, line 18: Could we merge these two flags into one? In other words, `--enable-hls PATH`. One of the flags does not make sense without the other.

phate: This would also render the need for the circt path mandatory, which means the CIRCT_PATH can be initialized with `""`.

scripts/build-circt.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
```
phate, line 2: Please add an `set -eu`.

scripts/build-circt.sh
```diff
@@ -0,0 +1,65 @@
...
+	echo ${GIT_COMMIT}
+}
+
+function usage()
+{
+	echo "Usage: ./build-circt.sh [OPTION] [VAR=VALUE]"
+	echo ""
+	echo "  --build-path PATH     The path where to build CIRCT."
+	echo "                        [${CIRCT_BUILD}]"
+	echo "  --install-path PATH   The path where to install CIRCT."
```
phate, line 19: Could we make these two arguments mandatory for the script and not provide default values for it? Default values makes everything always implicit, potentially leading to nasty surprises.

phate (changes requested):

.github/workflows/hls.yml
phate: This only provides the compiler for building jlm, not the library dependencies. In other words, you build jlm that depends on LLVM 16 with the clang15, which is perfectly fine to do, but not what we really wanted.

configure.sh
phate: Okay, I just saw your comment now after I already did the review. I absolutely do not agree with splitting the two flags up. See other comment.

README.md
sjalander: Removed

configure.sh
sjalander: This is intentional. The whole point is that only advanced users should need to think about where the build happens. Similar to not having to think about that there is a build-release and build-debug directory.
So, instead of merging them, I would argue for removing --circt-path completely and always force it to be in one specific location.

scripts/build-circt.sh
sjalander: Changed, but this has the minor drawback that the script now stops if it has been executed once, since the cloning of the CIRCT directory has already happened.

scripts/build-circt.sh
sjalander: Same as above. There is a default location for where the build happens, and the normal user never has to think about it.

sjalander: As the README.md says, the normal user would simply do
```
./scripts/build-circt.sh
./configure --enable-hls
```
instead of having to specify paths where to place things.

sjalander: I would even go so far as if `--enable-hls` is given and the `build-circt` directory does not exist, then the `./scripts/build-circt.sh` is called under the hood.

sjalander: One option would be to add a make target for building CIRCT that provides the configured path.
That would eliminate the default paths in the build-circt.sh script.```

In the root Makefile the lines starting with .PHONY could be added

```
ifdef HLS_ENABLED
include jlm/hls/Makefile.sub
include tools/jhls/Makefile.sub
include tools/jlm-hls/Makefile.sub

.PHONY: build-circt
build-circt: $(CIRCT_PATH)

$(CIRCT_PATH):
        ./scripts/build-circt.sh --build-path ./build-circt --install-path $(CIRCT_PATH)
endif
```

configure.sh
```diff
@@ -1,11 +1,10 @@
 #!/bin/bash
 
 # Default values for all tunables.
-CIRCT_PATH=
+CIRCT_PATH=${PWD}/build-circt/circt
```
haved, line 4: If I remember correctly this should be in quotes in case the person running the script has spaces in one of the parent folders

scripts/build-circt.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+set -eu
+
+GIT_COMMIT=093cdfe482530623fea01e1d3242af93e533ba54
+
+CIRCT_BUILD=${PWD}/build-circt
```
haved, line 6: I'm realizing there might already be quite a few places where we assume that file paths don't contain spaces. It might be better to just get this PR through first, then do a pure cleanup PR by testing everything in a folder with spaces, (if we care?)

phate:
@sjalander Could you merge master into this branch such that we could merge it?

phate (approved):

.github/workflows/hls.yml
```diff
@@ -8,53 +8,35 @@ on:
...
+    # Compile jlm with the HLS backend enabled
     # --------
 
-    - name: Configure jlm
-      run: ./configure.sh --circt-path $(pwd)/jlm-eval-suite/circt/local CXX=clang++-16
-    - name: Compile jlm
+    - name: "Build CIRCT"
+      uses: ./.github/actions/BuildCirct
+    - name: "Configure jlm"
+      run: ./configure.sh --enable-hls --circt-path ${{ github.workspace }}/build-circt/circt CXX=clang++-16
```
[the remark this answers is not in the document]
sjalander: Added

phate (approved):

PR #359, opened 2024-01-22
Add Rvsdg module class

phate:
This PR does the following:

1. Adds the RvsdgModule class. It should serve as top-level class for translation of modules in the future.
2. Utilize it as base class for the LLVM RVSDGModule class.

### DIFF
--- a/jlm/llvm/ir/RvsdgModule.hpp
+++ b/jlm/llvm/ir/RvsdgModule.hpp
@@ -92,6 +92,8 @@ is_export(const jlm::rvsdg::input * input)
 class RvsdgModule final : public rvsdg::RvsdgModule
 {
 public:
+  ~RvsdgModule() noexcept override = default;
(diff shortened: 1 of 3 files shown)


haved (approved):
LGTM

sjalander (commented):
What's the difference in definition between an RvsdgModule and an omega node, and why is a module needed while omega is implicit?

phate:
An omega node is a construct that we invented for the paper to fit in the pattern described by the other nodes in the paper, but never existed in the implementation. We can add it in the implementation if we want, but it does not add much value aside of adding another indirection to get to the root region. The RvsdgModule contains an RVSDG (jlm::rvsdg::graph), but might also contain additional information that are necessary, but are better not captured in the Rvsdg. Examples of such information are the target triple or data layout from the LLVM RVSDG module.

PR #360, opened 2024-01-22
Remove rcdtype from rvsdg namespace

phate:
This PR does the following:

1. Implements a llvm::StructType::Declaration class in the llvm namespace to replace the used rvsdg::rcddeclaration.
2. Utilizes llvm::StructType::Declaration instead of rvsdg::rcddeclaration.
3. Removes (now unused) record.hpp and record.cpp, which includes rvsdg::rcdtype and rvsdg::rcddeclaration.

This separates the concerns between rvsdg and llvm namespace more clearly and opens the path for properly keeping track of declarations instead of just letting them live for the entire program run.


jlm/llvm/ir/domtree.cpp
```diff
@@ -7,6 +7,7 @@
 #include <jlm/llvm/ir/cfg.hpp>
 #include <jlm/llvm/ir/domtree.hpp>
 
+#include <functional>
```
sjalander, line 10: This is the only change in this file, so is it needed here?
I can't see that the include have been removed from any other header, which might have included it earlier.

jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -9,6 +9,7 @@
 #include <jlm/llvm/ir/attribute.hpp>
 #include <jlm/llvm/ir/linkage.hpp>
 #include <jlm/llvm/ir/types.hpp>
+#include <jlm/rvsdg/graph.hpp>
```
sjalander, line 12: Same as before.

haved (approved):
The files with only includes modified stand out as maybe unintentional.
Otherwise very nice separation, LGTM.

jlm/llvm/ir/types.cpp
```diff
@@ -4,6 +4,7 @@
  */
 
 #include <jlm/llvm/ir/types.hpp>
+#include <jlm/util/strfmt.hpp>
```
sjalander, line 7: And here.

sjalander (commented):
It's not clear to me why the only change in some headers is an added include.

jlm/llvm/backend/rvsdg2jlm/context.hpp
```diff
@@ -9,6 +9,8 @@
 #include <jlm/llvm/ir/basic-block.hpp>
 #include <jlm/rvsdg/node.hpp>
 
+#include <unordered_map>
```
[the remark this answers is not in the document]
haved: The file does use `unordered_map`, so it makes sense to make it explicit, and not rely on one of the other headers including it for us

sjalander (approved):

phate:
@sjalander : The removed `record.hpp` had a bunch of included headers. Now, this `record.hpp` was included in `foo.hpp` and `foo.hpp` was included in `bla.hpp`. Removing `record.hpp` from `foo.hpp` also removes the included headers from `foo.hpp`, but also transitively from `bla.hpp`. If something in `bla.hpp` was dependent on one of the included headers from `record.hpp`, then this header is all in a sudden missing and the compiler gives me an error. This means I have to include the missing header in `bla.hpp` such that it compiles again. All of the header inclusions above were because the compiler told me so.

It is not good to have [transitive inclusion](https://softwareengineering.stackexchange.com/questions/262019/is-it-good-practice-to-rely-on-headers-being-included-transitively). However, it seems that there are tools for solving (some of the) issues:

https://github.com/flutter/flutter/issues/133417
https://include-what-you-use.org/
https://discourse.llvm.org/t/include-what-you-use-include-cleanup/5831

PR #367, opened 2024-01-25
Fix lifetime of StructType::Declarations in test RVSDGs

phate:
The declarations of struct types in the Test RVSDGs only lived for as long as we were in the SetupRvsdg() method. This is a problem as these declarations also needed to be alive after this method. It lead to dangling pointers. The only reason why valigrind-check did not catch this was because we never accessed these pointers again by coincidence.

This PR does the following:
1. Adds the possibility of adding these declarations to the RVSDG and InterProceduralGraph modules to extend their lifetime for as long as they are needed in these modules. This also enables us to remove a FIXME in the RVSDG construction, where these declarations were kept alive for as long as jlm was running.
2. Fixes the lifetime problem in the test RVSDG.


tests/TestRvsdgs.cpp
```diff
@@ -3036,8 +3037,9 @@ MemcpyTest2::SetupRvsdg()
 
   PointerType pointerType;
   arraytype arrayType(pointerType, 32);
-  auto structBDeclaration = StructType::Declaration::Create({ &arrayType });
```
phate, line 3039: This declaration was only alive as long as we were in SetupRvsdg().

tests/TestRvsdgs.cpp
```diff
@@ -3136,10 +3138,10 @@ LinkedListTest::SetupRvsdg()
   auto nf = rvsdg.node_normal_form(typeid(jlm::rvsdg::operation));
   nf->set_mutable(false);
 
-  auto declaration = StructType::Declaration::Create();
```
phate, line 3139: This declaration was only alive as long as we were in SetupRvsdg().

tests/TestRvsdgs.cpp
```diff
@@ -258,19 +258,20 @@ GetElementPtrTest::SetupRvsdg()
 {
   using namespace jlm::llvm;
 
-  auto declaration = StructType::Declaration::Create({ &jlm::rvsdg::bit32, &jlm::rvsdg::bit32 });
```
phate, line 261: This declaration was only alive as long as we were in SetupRvsdg().

sjalander (dismissed):
Maybe add documentation to the new functions in the headers.

phate:
I added some documentation to the added methods. @sjalander : I need a new review, please.

sjalander (approved):

PR #369, opened 2024-01-26
Enable load-store reduction again

### LINKED ISSUE #302: enable_load_reductions(graph) in -Jred throws a type_error
**Recreate**
Modify jlm/llvm/opt/reduction.cpp (line 134)

static void
{

  enable_load_reductions(graph);
  // enable_gamma_reductions(graph);
  // enable_unary_reductions(graph);
  // enable_binary_reductions(graph);

Then compile jlm with debug: 
make jlm-debug -j `nproc`

**Output**
  what():  Type error - expected : float, received : bit32
Aborted (core dumped)

[classify-f32.zip](https://github.com/phate/jlm/files/13560411/classify-f32.zip)

phate:
1. Fixes the load-store reduction. The problem was a missing check for the stored and loaded value type being the same. If they are not the same, then we might have to insert a truncation or conversion operation instead of just bypassing the load and store. I marked the place in the code with a FIXME such that we can improve upon it in the future.
2. Adds a unit test that exposes the problem.

Closes #302

### DIFF
--- a/jlm/llvm/ir/operators/load.cpp
+++ b/jlm/llvm/ir/operators/load.cpp
@@ -221,7 +221,7 @@ is_load_store_reducible(
     return false;
   }
 
-  // Check that all states to the load originate from the same store
+  // Check that all state edges to the load originate from the same store
   if (storeNode->NumStates() != loadOperation.NumStates())
   {
     return false;
--- a/jlm/llvm/ir/operators/load.cpp
+++ b/jlm/llvm/ir/operators/load.cpp
@@ -213,7 +213,7 @@ is_load_store_reducible(
     return false;
   }
 
-  // Check that the first state originates from a store
+  // Check that the first state edge originates from a store
   auto firstState = operands[1];
   auto storeNode = dynamic_cast<const StoreNode *>(rvsdg::node_output::node(firstState));
   if (!storeNode)
--- a/jlm/llvm/ir/operators/load.cpp
+++ b/jlm/llvm/ir/operators/load.cpp
@@ -201,6 +201,7 @@ is_multiple_origin_reducible(const std::vector<rvsdg::output *> & operands)
 // v2 s3 = load_op a s2
 // ... = any_op v2
 // =>
+// s2 = store_op a v1 s1
 // ... = any_op v1
 static bool
 is_load_store_reducible(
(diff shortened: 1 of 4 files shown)


jlm/llvm/ir/operators/load.cpp
```diff
@@ -197,30 +197,67 @@ is_multiple_origin_reducible(const std::vector<rvsdg::output *> & operands)
...
+  if (loadAddress != storeAddress)
+  {
+    return false;
+  }
+
+  // Check that the loaded and stored value type are the same
+  //
+  // FIXME: This is too restrictive and can be improved upon by inserting truncation or narrowing
+  // operations instead. For example, a store of a 32 bit integer followed by a load of a 8 bit
+  // integer can be converted to a trunc operation. Similarly, a store of a floating point
```
sjalander, line 251: Storing an FP value and then reading it back as an INT shouldn't perform any conversion, and shouldn't mean the same thing as using a machine instruction that convert FP to INT.

sjalander (dismissed):

jlm/llvm/ir/operators/load.cpp
phate: You are absolutely right. I removed the last sentence from the FIXME.

phate:
@sjalander or @haved : I need a new approval, please.

jlm/llvm/ir/operators/load.cpp
```diff
@@ -197,30 +197,66 @@ is_multiple_origin_reducible(const std::vector<rvsdg::output *> & operands)
   return states.size() != operands.size() - 1;
 }
 
+// s2 = store_op a v1 s1
+// v2 s3 = load_op a s2
+// ... = any_op v2
+// =>
```
haved, line 203: This comment kind of makes it appear as if the store disappears, I assume it doesn't?

haved (approved):
LGTM, just one small comment

jlm/llvm/ir/operators/load.cpp
phate: No, the store does not disappear. The only node that is removed is the load. I clarified in the function documentation by adding the store.

phate:
@sjalander and @haved : Clarified the documentation again after Hårvard's comment. I would need another approval.

haved (approved):
LGTM

PR #381, opened 2024-02-04
Install build rules

caleridas:
This provides build rules and glue to build packaged binary and development libraries. It also adds pkg-config files to simplify integration with other build systems (be it cmake, bazel, autotools,... which basically all understand pkgconfig well). Avoid name clashes of "too non-descriptive names" like libutil by adding jlm prefix to a few library names.


configure.sh
```diff
@@ -70,6 +80,18 @@ while [[ "$#" -ge 1 ]] ; do
 			usage >&2
 			exit 1
 			;;
+		--prefix=*)
```
haved, line 83: These flags are not mentioned in the help text, is that on purpose?
Also, the use of `--flag=<value>` instead of `--flag <value>` is different from the others, is this due to Debian standards for configure scripts or something?

Anyway, it might be a good argument for rewriting the configure-script in python with `argparse` to get more general parsing.

phate (changes requested):
@haved mentioned it already in his comment. The only issue I have with this PR is the inconsistency of the flags.

configure.sh
caleridas: "--flag=X" vs. "--flag X": This is really annoying, I figured that meta-buildsystems kind of expect both to work with --flag=X being used more often (most standard packaging scripts). Ideally, both should work but TBH I don't want the feature creep of fully reimpleminting autoconf & friends.
Rewriting in python: opens a different of issues depending on it as build-time requirement. Depending on where and how things are built, it is "#!/usr/bin/python", "#!/usr/bin/python3" "#!/bin/env python" "#!/bin/env python3" and if things turn ugly it is "#!/bin/env python2.7".... posix sh sucks badly from ergonomic point of view, but at least is realible.
Documentation is missing, will fix.

phate: As talked offline, we leave it as it is right now. I am not happy with the parameter syntax mix, but if that is what is needed right now, then let's go ahead. Just add the documentation and we can merge this one.

phate:
@caleridas Do you still plan to merge this one or abandon it?

caleridas:
> @caleridas Do you still plan to merge this one or abandon it?

Would like to get packaging scripts merged, but not in the next few days.

Issue #361, opened 2024-01-22
Improve statistics serialization


phate:
The statistics class has currently only a [ToString()](https://github.com/phate/jlm/blob/ae7fee46136e92eb1025b55b9577ffd5da045ba8/jlm/util/Statistics.hpp#L64C1-L65C24) method, returning the statistics of a transformation pass as a serialized string. The consequences of this method are manifold:

1.  It returns a string and no structured data. It is therefore very difficult to implement unit tests for the individual statistic classes of the transformations (without parsing the string again).
2. Each transformation needs to serialize the internal data structures themselves into a string.
3. The serialized string must have the same format as the string created from all the other transformation statistics, but the formatting is not centralized. This formatting should be centralized. We currently use a comma separated list of name:value pairs as exemplified [here](https://github.com/phate/jlm/blob/ae7fee46136e92eb1025b55b9577ffd5da045ba8/jlm/llvm/opt/alias-analyses/Steensgaard.cpp#L862C1-L916C4). For each value, we must be careful to comma separate them, insert the colon correctly, add the whitespaces, etc. This should be centralized.

The idea in my head:
1. Implement a pure abstract `GetMeasurements()` method in the Statistics class. This class should return a `std::vector<>()` of measurements, which are nothing else than name-value pairs.
2. Each statistics class of a transformation returns then its measurements, and at a central place, these measurements are serialized.

PR #382, opened 2024-02-05
Add measurements and timers to the Statistics base class

haved:
While modifying Andersen, I tried to evolve the `Andersen::Statistics` class with the changes, and was annoyed by two things:

 - Modifying the set of tracked measurements required 4 changes to the code (ctor init, field, assigning the value and serializing the value)
 - There was no nice way of making certain fields optional
 - Also, tests don't have access to statistics

In this PR is a suggestion for a modified base class for `Statistics`. I also modified `Andersen::Statistics` to make use of it. This PR should of course also have tests added, but I'm putting it up now to get feedback.

This PR should correspond to #361

### DIFF
--- a/jlm/llvm/opt/alias-analyses/Andersen.cpp
+++ b/jlm/llvm/opt/alias-analyses/Andersen.cpp
@@ -34,34 +34,15 @@ public:
 
   explicit Statistics(jlm::util::filepath sourceFile)
       : jlm::util::Statistics(Statistics::Id::AndersenAnalysis),
-        SourceFile_(std::move(sourceFile)),
-        NumRvsdgNodes_(0),
-        NumPointerObjects_(0),
-        NumRegistersMappedToPointerObject_(0),
-        NumSupersetConstraints_(0),
-        NumAllPointeesPointToSupersetConstraints_(0),
-        NumSupersetOfAllPointeesConstraints_(0),
-        NumHandleEscapingFunctionConstraints_(0),
-        NumFunctionCallConstraints_(0),
-        NumConstraintSolvingIterations_(0),
-        NumPointsToGraphNodes_(0),
-        NumRegisterNodes_(0),
-        NumAllocaNodes_(0),
-        NumDeltaNodes_(0),
-        NumImportNodes_(0),
-        NumLambdaNodes_(0),
-        NumMallocNodes_(0),
-        NumMemoryNodes_(0),
-        NumEscapedNodes_(0),
-        NumExternalMemorySources_(0)
+        SourceFile_(std::move(sourceFile))
   {}
 
   void
   StartAndersenStatistics(const jlm::rvsdg::graph & graph) noexcept
   {
-    NumRvsdgNodes_ = jlm::rvsdg::nnodes(graph.root());
-    AnalysisTimer_.start();
-    SetAndConstraintBuildingTimer_.start();
+    AddMeasurement("#RvsdgNodes", jlm::rvsdg::nnodes(graph.root()));
+    AddTimer("AnalysisTimer").start();
+    AddTimer("SetAndConstraintBuildingTimer").start();
   }
 
   void
@@ -69,10 +50,15 @@ public:
       const PointerObjectSet & set,
       const PointerObjectConstraintSet & constraints) noexcept
   {
-    SetAndConstraintBuildingTimer_.stop();
+    GetTimer("SetAndConstraintBuildingTimer").stop();
+    AddMeasurement("#PointerObjects", set.NumPointerObjects());
+    AddMeasurement("#RegistersMappedToPointerObject", set.GetRegisterMap().size());
 
-    NumPointerObjects_ = set.NumPointerObjects();
-    NumRegistersMappedToPointerObject_ = set.GetRegisterMap().size();
+    size_t numSupersetConstraints = 0;
+    size_t numAllPointeesPointToSupersetConstraints = 0;
+    size_t numSupersetOfAllPointeesConstraints = 0;
+    size_t numHandleEscapingFunctionConstraints = 0;
+    size_t numFunctionCallConstraints = 0;
     for (const auto & constraint : constraints.GetConstraints())
     {
       std::visit(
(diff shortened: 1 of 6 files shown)


jlm/util/Statistics.hpp
```diff
@@ -64,8 +75,164 @@ class Statistics
...
+  // Mutable version of @see GetTimer
+  [[nodiscard]] util::timer &
+  GetTimer(const std::string & name);
+
+  /**
+   * Removes the timer with the given \p name, if it exists.
+   * @return true if the timer existed and was removed, false otherwise
+   */
+  bool
+  RemoveTimer(const std::string & name);
```
phate, line 228: This interface feels a little bit inconsequent to me:

1. I would remove `RemoveTimer()` from it. Do we have a use case for this method?
2. You are offering several methods reflecting the operations on a timer, but at the same time also return the underlying timer for each of those methods, which could be used to perform the same operations.

Alternative interface suggestion:

```
util::timer &
AddTimer(std::string name);

[[nodiscard]] util::timer &
GetTimer(const std::string & name);
```

And then rely on the timer interface for performing the necessary operations. In addition, I would delete the copy constructor and copy assignment operator, as it would be way too easy to do this mistake:

`auto timer  = GetTimer("myTimer")` instead of `auto & timer = GetTimer("myTimer")`.

jlm/util/time.hpp
```diff
@@ -15,29 +15,71 @@ class timer final
 {
 public:
   constexpr timer()
+      : CountedNanoseconds_(0),
```
phate, line 18: How about `ElapsedTimeInNanoseconds`?

jlm/util/time.hpp
```diff
@@ -15,29 +15,71 @@ class timer final
...
+        Running_(false)
```
phate, line 19: How about `IsRunning_`?

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -34,113 +34,104 @@ class Andersen::Statistics final : public jlm::util::Statistics
...
 
   void
   StartAndersenStatistics(const jlm::rvsdg::graph & graph) noexcept
   {
-    NumRvsdgNodes_ = jlm::rvsdg::nnodes(graph.root());
-    AnalysisTimer_.start();
-    SetAndConstraintBuildingTimer_.start();
+    AddUInteger("#RvsdgNodes", jlm::rvsdg::nnodes(graph.root()));
+    StartTimer("AnalysisTimer");
+    StartTimer("SetAndConstraintBuildingTimer");
```
phate, line 45: Could factor out these magic strings into variables? In particular, those strings that are used multiple times, such as "SetAndConstraintBuildingTimer"?

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -150,77 +141,7 @@ class Andersen::Statistics final : public jlm::util::Statistics
...
-        " ",
-        "ConstraintSolvingTime[ns]:",
-        ConstraintSolvingTimer_.ns(),
-        " ",
-        "PointsToGraphConstructionTime[ns]:",
-        PointsToGraphConstructionTimer_.ns(),
-        " ",
-        "PointsToGraphConstructionExternalToEscapedTime[ns]:",
-        PointsToGraphConstructionExternalToEscapedTimer_.ns());
+        SerializeMeasurements());
```
phate, line 144: Could you add another `SerializeMeasurements()` method that prefixes the measurements with the Statistics::Id followed by the source file. Ultimately, this is what we want (and after we changed everything to use that one, we remove the ToString() method). However, I understand that the current method is necessary in order to not have the changes all at once all over the place. I can take care in the end to refactor all the other statistics.

jlm/util/Statistics.cpp
```diff
@@ -1,15 +1,230 @@
...
 
+#include <jlm/util/strfmt.hpp>
+
 namespace jlm::util
 {
 
 Statistics::~Statistics() = default;
 
+std::string
+Statistics::SerializeMeasurements() const
```
phate, line 17: This method does more than the name suggests (it also serializes the timers). How about just naming it `Serialize()`?

phate (changes requested):
Thanks for the initiative. I really appreciate that you try to improve the infrastructure instead of just living with it!

Could we split this in 3 individual PRs:
1. A PR for the timer. I just had minor suggestions there. I would also add some unit tests (I know there are none for it currently).
2. A PR for the changes in statistics. Overall, I am very positive to the change. It addresses the concerns I mentioned [here](https://github.com/phate/jlm/issues/361). We need to hammer out the details though. I would also add some unit tests on this PR.
3. A PR where you use the new interface in Andersen. Since the changes allow now to easily write unit tests, I would also add one here.

jlm/llvm/opt/alias-analyses/Andersen.cpp
haved: I tried doing something like that at first (giving `ToString` a default implementation) but encountered two issues:

 - where can I find the ID as a string? I'm guessing it eventually should correspond to the ID used on the command line, but right now there is no source of truth location for translating between the enum and a string.
 - not all Statistics have a field for the file name. The `RegionAwareMemoryNodeProvider`'s statistics don't include it in it's output. Should all Statistics have the fileName included? If so, we can maybe move the field to the superclass and add it to the ctor?

jlm/llvm/opt/alias-analyses/Andersen.cpp
haved: I'm a bit hesitant to put all the strings into variables. Some of the purpose of this PR was to reduce the number of separate lines of code that need to be modified whenever a statistic is added, removed or changed.

If we were to add variables for these strings, they would either only be accessible within the `.cpp` file, where most of them only see one use, or they would have to be in the `.hpp` file, which makes it somewhat easier for tests to retrieve measurements, but now requires you to modify two files whenever a statistic is added/removed/modified.

The only place I can see where having access to the strings is useful is when retrieving statistics in tests. By using variables instead of magic strings, IDEs can correctly perform renaming across files, which is nice. However, a test failure makes itself apparent "almost" as quickly as a compilation error, so I don't think it is worth it to place all string literals in variables in header files for only tests' sake.

For the timers, I agree, it is a bit annoying having to reference them by name twice. It will assert false if there is a mismatch, but variables would make it clearer. The question is if the variable should be a reference to the timer object, or a string.

haved:
@phate I pushed my changes to resolve as many of your comments as I could, but this PR will still be turned into two separate PRs. I just want your feedback on my response to one of your comments, before I make the Statistics class PR.

jlm/llvm/opt/alias-analyses/Andersen.cpp
phate: I should have been a little bit more elaborate in my original comment. Sorry! Why I mentioned the magical strings was that you are requesting timers multiple times (at least twice, once for start and once for stop), and in both cases you are using a string. Thus, instead of writing something like this:

```
class Statistics
{
public:
  void
  StartMySpecialStatistics()
  {
    StartTimer("MySpecialStatisticsTimer");
    ...
  }

  void
  StopMySpecialStatistics()
  {
     StopTimer("MySpecialStatisticsTimr"); // typo here
  }
}
```

I have something like this in mind:
```
class Statistics
{
private:
  const char * MySpecialStatisticsTimerTag_ = "MySpecialStatisticsTimer";
public:
  void
  StartMySpecialStatistics()
  {
    StartTimer(MySpecialStatisticsTimerTag_);
    ...
  }

  void
  StopMySpecialStatistics()
  {
     StopTimer(MySpecialStatisticsTimerTag_);
  }
}
```

This was only meant for:
1. The statistics class internally and was not supposed to "leak" into tests/outside.
2. Only for instances were you have to use them multiple times (to avoid typos, and make your life easier in terms of renaming by being able to utilize the IDE)
3. The idea was to avoid typos and potentially reference a wrong timer, but looking at the code you would throw an exception if one would have a typo. So this argument does not really count any longer (my bad, should have taken a closer look at the code).

I do not really see that the "additional number of lines" argument counts here as most of the renaming with those tags would be done by the IDE in the end. Personally, I would still try to use such tags for entities that I have to reference at least multiple times, but this is just personal preference.

Due to the exception, the original argument is void. Leave it as it is if you like. No strong opinion about this.

jlm/llvm/opt/alias-analyses/Andersen.cpp
phate: 1. As far as I know, there is currently no method for providing a string representation of a statistics ID. C++ does not come equipped with it, and I have not implemented one so far. You would need to implement it. We have methods for translating the statistics IDs to command line flags though. See [here](https://github.com/phate/jlm/blob/94ba20fd93ace0da2a0b52df6891cc3b46051bb4/jlm/tooling/CommandLine.cpp#L180C1-L269C2). I am a little bit reluctant of tying the internal string representation of an statistics id to the command line flag though. It would mean that by changing the name of an id (and its string representation), one would all in a sudden also change the command line flag, which is a public interface. I think they should be separate concerns.
2. Yes, all statistics should have the filename included. It makes no sense to leave this (important) piece of information out. Moving it to the superclass and adding it to the ctor would be the right thing to do.

phate:
@haved We can close this PR, right? It was replaced by the three individual PRs.

haved:
@phate Yes, closing it now

PR #384, opened 2024-02-06
Ensure unique use of statistics identifiers

phate:
We would like to use statistics identifiers as tags in our statistics logs, but for that they need to be (reasonably) descriptive and uniquely associated with a single pass in jlm. This PR accomplishes this, fixing parts of #361.


haved (approved):
LGTM, it's interesting to note how many places you have to modify when changing an id. It will be very nice to create a single source of truth eventually.

Issue #357, opened 2024-01-20
Improve statistics command line flag


phate:
The statistics command line flags for jlc and jlm-opt are very verbose and partly non-descriptive. The following should be done:

1. Improve the descriptiveness of the statistics command line flags. They should describe/reflect the (optimization) pass they are gathering statistics for.
2. We should transform the statistics flag into a comma separated list: `--statistics=statistics1, statistics2,...`

haved:
The duplication I was referencing is the set of statistics options listed both at
 https://github.com/phate/jlm/blob/master/jlm%2Ftooling%2FCommandLine.cpp#L512

and 

https://github.com/phate/jlm/blob/master/jlm%2Ftooling%2FCommandLine.cpp#L771

Which can be nicely deduplicated by replacing the individual command line flags by a single comma separated string.

Also, the mapping between StatisticsId and statistic name can use a BiMap :)

PR #386, opened 2024-02-07
Simplify jlc and jlm-opt command line option setup

phate:
We had a lot of unnecessary code for setting up the statistics command line options of jlm-opt and jlc. This PR simplifies it. It is part of #357.

### DIFF
--- a/jlm/tooling/CommandLine.hpp
+++ b/jlm/tooling/CommandLine.hpp
@@ -175,33 +175,11 @@ private:
     inline static const char * NodeReduction_ = "NodeReduction";
   };
 
-  static inline const util::BijectiveMap<util::Statistics::Id, std::string>
-      StatisticsIdCommandLineArguments_ = {
-        { util::Statistics::Id::Aggregation, "print-aggregation-time" },
-        { util::Statistics::Id::AgnosticMemoryNodeProvisioning,
-          "print-agnostic-memory-node-provisioning" },
-        { util::Statistics::Id::AndersenAnalysis, "print-andersen-analysis" },
-        { util::Statistics::Id::Annotation, "print-annotation-time" },
-        { util::Statistics::Id::CommonNodeElimination, "print-cne-stat" },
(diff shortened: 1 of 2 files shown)


phate:
I know that we can do quite a bit more (simplifying optimization flags setup, output format setup, statistics command line descriptions, etc.), but I would like to keep it there for now.

sjalander (approved):

jlm/tooling/CommandLine.hpp
```diff
@@ -174,33 +175,33 @@ class JlmOptCommandLineOptions final : public CommandLineOptions
...
-    inline static const char * ReduceNodes_ = "print-reduction-stat";
-    inline static const char * RegionAwareMemoryNodeProvisioning_ =
-        "print-memory-node-provisioning";
-    inline static const char * RvsdgConstruction_ = "print-rvsdg-construction";
-    inline static const char * RvsdgDestruction_ = "print-rvsdg-destruction";
-    inline static const char * RvsdgOptimization_ = "print-rvsdg-optimization";
-    inline static const char * SteensgaardAnalysis_ = "print-steensgaard-analysis";
-    inline static const char * ThetaGammaInversion_ = "print-ivt-stat";
-  };
+  static inline const util::BijectiveMap<util::Statistics::Id, std::string>
```
haved, line 178: Declaring this in the header with `inline` might add the construction of this object to the static initialization of all translation units that include `CommandLine.hpp`. Making it just static, and moving the initialization to `CommandLine.cpp` could work. I will test if it at all makes a difference.

haved: Or at the very least we could use `std::string_view` instead of `std::string`, since these strings are constants with static lifetime, to avoid copying into heap allocated strings all over the place.

phate: Great catch! I changed it, please have a look again.

haved (approved):
LGTM, I removed an unused static variable from JlmOptCommandLineOptions before accepting

PR #387, opened 2024-02-07
Add measurement/timer system and sourceFile to Statistics base class

haved:
The Statistics part of #382. Modifies all existing Statistics to provide the sourceFile path.

The mapping from id to name is now in `Statistics.cpp`, but many of the existing statistics used names that were different from the C++ enums, so I tried to make the classes keep their old names.

It might be better to have `GetName()` be an abstract method, and let subclasses provide names that way, instead of a single BijectiveMap. Let me know!

### DIFF
--- a/jlm/llvm/opt/InvariantValueRedirection.cpp
+++ b/jlm/llvm/opt/InvariantValueRedirection.cpp
@@ -19,8 +19,8 @@ class InvariantValueRedirection::Statistics final : public util::Statistics
 public:
   ~Statistics() override = default;
 
-  Statistics()
-      : util::Statistics(Statistics::Id::InvariantValueRedirection)
+  Statistics(util::filepath sourceFile)
+      : util::Statistics(Statistics::Id::InvariantValueRedirection, std::move(sourceFile))
   {}
 
   void
@@ -36,15 +36,15 @@ public:
   }
 
   [[nodiscard]] std::string
-  ToString() const override
+  Serialize() const override
   {
-    return util::strfmt("InvariantValueRedirection ", "Time[ns]:", Timer_.ns());
+    return util::strfmt("Time[ns]:", Timer_.ns());
   }
 
   static std::unique_ptr<Statistics>
-  Create()
+  Create(util::filepath sourceFile)
   {
-    return std::make_unique<Statistics>();
+    return std::make_unique<Statistics>(std::move(sourceFile));
   }
 
 private:
@@ -59,7 +59,7 @@ InvariantValueRedirection::run(
     util::StatisticsCollector & statisticsCollector)
 {
   auto & rvsdg = rvsdgModule.Rvsdg();
-  auto statistics = Statistics::Create();
+  auto statistics = Statistics::Create(rvsdgModule.SourceFileName());
 
   statistics->Start();
   RedirectInvariantValues(*rvsdg.root());
--- a/jlm/llvm/opt/InvariantValueRedirection.cpp
+++ b/jlm/llvm/opt/InvariantValueRedirection.cpp
@@ -19,8 +19,8 @@ class InvariantValueRedirection::Statistics final : public util::Statistics
 public:
   ~Statistics() override = default;
 
-  Statistics(util::filepath sourceFile)
-      : util::Statistics(Statistics::Id::InvariantValueRedirection, std::move(sourceFile))
+  explicit Statistics(const util::filepath & sourceFile)
+      : util::Statistics(Statistics::Id::InvariantValueRedirection, sourceFile)
   {}
 
   void
@@ -42,9 +42,9 @@ public:
   }
 
   static std::unique_ptr<Statistics>
-  Create(util::filepath sourceFile)
+  Create(const util::filepath & sourceFile)
   {
-    return std::make_unique<Statistics>(std::move(sourceFile));
+    return std::make_unique<Statistics>(sourceFile);
   }
 
 private:
--- a/jlm/llvm/opt/cne.cpp
+++ b/jlm/llvm/opt/cne.cpp
@@ -18,8 +18,8 @@ class cnestat final : public util::Statistics
 public:
   ~cnestat() override = default;
 
-  cnestat(util::filepath sourceFile)
-      : Statistics(Statistics::Id::CommonNodeElimination, std::move(sourceFile)),
+  explicit cnestat(const util::filepath & sourceFile)
+      : Statistics(Statistics::Id::CommonNodeElimination, sourceFile),
         nnodes_before_(0),
         nnodes_after_(0),
         ninputs_before_(0),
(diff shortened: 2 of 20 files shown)


jlm/llvm/opt/InvariantValueRedirection.cpp
```diff
@@ -19,8 +19,8 @@ class InvariantValueRedirection::Statistics final : public util::Statistics
 public:
   ~Statistics() override = default;
 
-  Statistics()
-      : util::Statistics(Statistics::Id::InvariantValueRedirection)
+  Statistics(util::filepath sourceFile)
```
phate, line 22: `explicit` would be nice.

jlm/llvm/opt/cne.cpp
```diff
@@ -16,11 +16,10 @@ namespace jlm::llvm
 class cnestat final : public util::Statistics
 {
 public:
-  virtual ~cnestat()
-  {}
+  ~cnestat() override = default;
 
-  cnestat()
-      : Statistics(Statistics::Id::CommonNodeElimination),
+  cnestat(util::filepath sourceFile)
```
phate, line 21: `explicit`

jlm/llvm/opt/inlining.cpp
```diff
@@ -16,11 +16,10 @@ namespace jlm::llvm
 class ilnstat final : public util::Statistics
 {
 public:
-  virtual ~ilnstat()
-  {}
+  ~ilnstat() override = default;
 
-  ilnstat()
-      : Statistics(Statistics::Id::FunctionInlining),
+  ilnstat(util::filepath sourceFile)
```
phate, line 21: `explicit`

jlm/util/Statistics.hpp
```diff
@@ -62,11 +74,123 @@ class Statistics
     return StatisticsId_;
   }
 
+  /**
+   * @return a string identifying the type of this Statistics instance
+   */
+  [[nodiscard]] std::string_view
```
phate, line 80: `const std::string_view &` as return type?

jlm/util/Statistics.cpp
```diff
@@ -1,15 +1,171 @@
...
+  return false;
+}
+
+Statistics::Measurement &
+Statistics::GetMeasurement(const std::string & name)
+{
+  for (auto & [mName, measurement] : Measurements_)
+    if (mName == name)
+      return measurement;
+  throw util::error(util::strfmt("Unknown measurement: ", name));
```
phate, line 115: I would almost like to replace this with `JLM_ASSERT()`. Nobody catches it anyway, and it is basically a bug to ask for a measurement that is not present.

jlm/util/Statistics.cpp
```diff
@@ -1,15 +1,171 @@
...
+  return false;
+}
+
+util::timer &
+Statistics::GetTimer(const std::string & name)
+{
+  for (auto & [mName, timer] : Timers_)
+    if (mName == name)
+      return timer;
+  throw util::error(util::strfmt("Unknown timer: ", name));
```
phate, line 145: `JLM_ASSERT()`

jlm/util/Statistics.hpp
```diff
@@ -62,11 +74,123 @@ class Statistics
...
+   * Retrieves the full list of timers
+   */
+  [[nodiscard]] util::iterator_range<TimerList::const_iterator>
+  GetTimers() const;
+
+protected:
+  /**
+   * Adds a measurement, identified by \p name, with the given value.
+   * Requires that the measurement doesn't already exist.
+   * Measurements are listed in insertion order.
```
phate, line 161: This is not really relevant for the documentation of this method, right?

jlm/util/Statistics.hpp
```diff
@@ -62,11 +74,123 @@ class Statistics
...
+   */
+  template<typename T>
+  void
+  AddMeasurement(std::string name, T value)
+  {
+    JLM_ASSERT(!HasMeasurement(name));
+    Measurements_.emplace_back(std::make_pair(std::move(name), std::move(value)));
+  }
+
+  // Mutable version of @see GetMeasurement
```
phate, line 172: Does doxygen recognize `//`?

jlm/util/Statistics.hpp
```diff
@@ -62,11 +74,123 @@ class Statistics
...
+
+  /**
+   * Creates a new timer with the given \p name.
+   * Requires that the timer does not already exist.
+   * @return a reference to the timer
+   */
+  util::timer &
+  AddTimer(std::string name);
+
+  // Mutable version of @see GetTimer
```
phate, line 184: Does doxygen recognize `//`?

jlm/util/Statistics.hpp
```diff
@@ -62,11 +74,123 @@ class Statistics
...
+  [[nodiscard]] util::timer &
+  GetTimer(const std::string & name);
 
 private:
   Statistics::Id StatisticsId_;
+  jlm::util::filepath SourceFile_;
```
phate, line 190: `util::filepath SourceFile_`

tests/jlm/util/TestStatistics.cpp
```diff
@@ -14,36 +14,58 @@
...
+  filepath path("file.ll");
+  MyTestStatistics statistics(Statistics::Id::Aggregation, path);
+
+  statistics.Start(10, 6.0);
+  statistics.Stop(-400, "poor");
+
+  assert(statistics.GetMeasurementValue<uint64_t>("count") == 10);
+  assert(statistics.GetMeasurementValue<double>("weight") == 6.0);
+  assert(statistics.GetMeasurementValue<int64_t>("bankAccount") == -400);
+  assert(statistics.GetMeasurementValue<std::string>("count") == "poor");
```
phate, line 51: I think the key should be `"state"` instead of `"count"`

phate (dismissed):
Some minor issues, but nothing critical. Up to you whether you want to fix them.

I guess the next steps are (?):
1. Convert all statistics to use the new interface
2. Clean up the Serialize()/ToString() methods of the Statistics class

If you want, I can take care of the (mundane) first point such that you can continue working on Andersen.

haved:
Thanks for the comments, I fixed almost all of them. I elected to go for `const util::filepath & sourceFile`, and remove a bunch of `std::move`.

I also modified the set of methods in `Statistics.hpp`, since it turns out C++ does overload resolution before checking visibility. Having public-but-const and protected non-const versions of the same method worked surprisingly poorly.

The only comment you had that I didn't act on was returning `string_view` vs. `const string_view &`, as I think the first is more flexible, and doesn't involve any copying.

Regarding converting classes, I was not planning on doing that work now. I want to get Andersen's statistics over to the `AddMeasurement` functionality, so that it becomes easier to add my new worklist implementation side-by-side.

phate (approved):
LGTM

PR #391, opened 2024-02-10
Convert statistics classes to use new interface

phate:
This PR refactors statistics classes to use the new Statistics class interface.


phate:
@haved I did not touch the Andersen statistics class. It is the only one that I did not convert.

haved (approved):
Looks great, the Label struct for repeated statistics labels is a good idea, makes it easier to maintain parity between different statistics classes

PR #395, opened 2024-02-12
[Andersen] move Statistics to the new system, add PassConfig, rename Constraints

haved:
This comes on top of #393. Ignore those files for now.

`Andersen::Statistics` has been modified to use the new system. A lot of the RVSDG-related labels are shared with Steensgaard, so I can move those strings to `Labels` if desired.

The `PassConfiguration` is what allows the choice of multiple solver implementation, and eventually specifying what kinds of offline preprocessing should be done.

I renamed the two horribly named constraints to `LoadConstraint` and `StoreConstraint`, since that is the only situation they are used, and it aligns better with the literature.


jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -898,6 +795,18 @@ Andersen::AnalyzeRvsdg(const rvsdg::graph & graph)
   }
 }
 
+void
+Andersen::SetConfiguration(PassConfiguration config)
+{
+  Config_ = config;
```
phate, line 801: `std::move(config)`. It does not make much of a difference now, but might going forward.

jlm/llvm/opt/alias-analyses/Andersen.hpp
```diff
@@ -25,10 +25,41 @@ class Andersen final : public AliasAnalysis
   class Statistics;
 
 public:
-  Andersen() = default;
+  /**
+   * class for configuring the Andersen pass, such as what solver to use.
+   */
+  class PassConfiguration
```
phate, line 31: Just a suggestion, but maybe `Configuration` is enough? It would then be `Andersen::Configuration`, which is quite readable.

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -350,59 +361,171 @@ class SupersetConstraint final
...
+   * @param superset the new PointerObject that should point to everything the subset points to
+   */
+  void
+  SetSuperset(PointerObject::Index superset)
+  {
+    Superset_ = superset;
+  }
+
+  /**
+   * @return the PointerObject whose points to-set should be contained within the superset
```
phate, line 382: "points-to set"

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -44,15 +44,15 @@ class PointerObject final
...
+  // May point to a memory object defined outside the module, or any escaped memory object
   uint8_t PointsToExternal_ : 1;
 
   // This memory object's address is known outside the module.
   // If set and Kind_ is Register, it only means the pointees of the Reigister have escaped.
   uint8_t HasEscaped_ : 1;
 
 public:
-  using Index = size_t;
+  using Index = std::uint32_t;
```
phate, line 55: Any particular reason why you changed that? The `size_t` guaranteed before that it is always the size of the register. Why was it important to know the exact size here? Considering also that you are actually hiding it behind `Index` again.

phate (commented):
Some minor suggestions, but nothing major. LGTM.

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
haved: This is due to the tendency of running out of memory while analyzing. Almost all our memory usage is sets containing `PointerObject::Index`, so we really save a lot of memory by doing this.

At the risk of making a "Bill Gates 640K statement", I think 4 billion variables + registers is going to be enough for any program.

All the points-to-sets need to be in memory while creating the final PointsToGraph, which becomes a bottleneck. It should also help speed, as the memory bandwidth hopefully decreases.

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
haved: I notice I have mixed "points to-set" and "points-to-set", but I agree that in English, points-to set makes the most sense.

haved:
I applied the changes you suggested, and merged with the changes to #393.

PR #396, opened 2024-02-13
Added Verilator tests for the HLS backend

sjalander:
Added --enable-verilator to configure.sh, which makes it possible to run checks that are dependent on verilator.


phate (commented):
What is the reasoning for moving all the tests from the jlm-eval-suite to jlm?

We just removed the verilator dependency from jlm (except in the  HLS workflow). All in a sudden we are reintroducing it to jlm again.

phate:
There is a lot to unpack here.

> One reason is to remove the dependency of jlm on jlm-eval-suite. Currently, we need to stage PR requests between jlm-eval-suite and jlm in a certain way to assure that the CI doesn't break for either of them when making updates. This has caused hassles on a number of occasions already.
The current CI has a dependency for HLS and the llvm-test-suite on jlm-eval-suite. Polybench already clones the polybench repository directly, and this change removes the dependency for HLS.

I do not think that you will get rid off the staging process. You have two requirements, one in jlm and one in jlm-eval-suite, which creates a cyclic dependency:
1. We check for every change in jlm that all the tests in jlm-eval-suite compile and run.
2. We check for every change to the tests in jlm-eval-suite that jlm still compiles with all tests.

The problem arises now if you make changes in jlm to the public interface, i.e., the command line interface of a tool. All in a sudden one might require also a change in jlm-eval-suite to incorporate this update. The same staging process is done for LLVM and all other tools, just that it is called versioning there, i.e., the public interface for a major version cannot be changed. We do not do versioning in jlm, because it is a lot of effort and does no square with the little time we have. Thus, we rather have breaking changes, but for that increase our velocity (towards papers).

The current state of this staging process is as follows:
1. We use the HEAD of master in the jlm CI
2. We use a specific commit, i.e.,version, of jlm in the jlm-eval-suite CI via git submodule.

I think that this should be reversed. We should use a specific jlm-eval-suite version in the jlm CI and the HEAD of master from jlm in the jlm-eval-suite CI. The reasoning is that jlm changes significantly faster than the jlm-eval-suite and therefore require more often updates to the version in jlm-eval-suite submodule/CI.

Moreover, why do we have jlm included as a submodule in jlm-eval-suite? I fail to recall the reasoning. I suspect it was to have jlm always in the right version readily available, as we do not have a configure script. If we would change the dependency as above, then we could just check out the master of jlm in the jlm-eval-suite CI.

> Another reason is that these tests are equivalent to the c-tests, which test end-to-end functionality, but for the HLS backend. So it seems logical that they are part of jlm.

The c-tests are superseded by the llvm-test-suite. They are from a time when we did not have llvm-test-suite yet. They should in my opinion either be moved in the jlm-eval-suite or completely removed. The end-to-end tests of jlm should not be in the same repository as jlm. These tests are only dependent on the jlm tools and nothing else. LLVM has also a separate llvm-test-suite repository.

In summary, what I think should happen:
1. Reverse the dependency between the jlm and jlm-eval-suite CI as described above.
2. Either remove the c tests from jlm or move them to jlm-eval-suite
3. Remove the jlm submodule from the jlm-eval-suite and rather introduce a configure script in jlm-eval-suite for configuring the Makefiles with the jlm tool.

sjalander:
@phate 
If the c-tests are in principle to be removed, then it's consistent.

PR #397, opened 2024-02-13
Improvements to build-circt.sh

sjalander:
Better handling of build and install paths, as well as the possibility to rerun the script if the build process was stopped or modified for some reason.


scripts/build-circt.sh
```diff
@@ -3,6 +3,20 @@ set -eu
 
 GIT_COMMIT=093cdfe482530623fea01e1d3242af93e533ba54
 
+# Get the absolute path to this script and set default build and install paths
+SCRIPT_DIR="$(dirname "$(realpath "$0")")"
+CIRCT_BUILD=${SCRIPT_DIR}/../build-circt
+CIRCT_INSTALL=${SCRIPT_DIR}/../usr
```
phate, line 9: Do I understand this correctly in that the installation is in the end in `.../jlm/usr` (if the below statements do not kick in)?

scripts/build-circt.sh
```diff
@@ -3,6 +3,20 @@ set -eu
...
+
+# Check if Makefile.config exists and use it to set the installation path
+CONFIG=${SCRIPT_DIR}/../Makefile.config
+if test -f "$CONFIG"; then
+	CIRCT_PATH=$(sed -n '/CIRCT_PATH=/s/^.*=//p' ${CONFIG})
+	if [ -n "${CIRCT_PATH-}" ]; then
+		CIRCT_INSTALL=${SCRIPT_DIR}/../${CIRCT_PATH}
+	fi
+fi
+
```
phate, line 19: This will tightly couple this build script to the Makefile.config. Even worse, it will couple it to the variable name `CIRCT_INSTALL` being used in Makefile.config. Why is this necessary? What is wrong with just providing these arguments on the command line?

scripts/build-circt.sh
sjalander: Yes, the default would be to place it in jlm/usr

scripts/build-circt.sh
sjalander: As I principle, I am in favor of having a single set of configurations that can then be used to set up the environment, such that it is coherent for the user.

configure.sh already provides a number of configuration variables that defines where things must be placed, so why not use existing information to assure that the setup is correct, i.e., **one set of ground truths**.

I agree that depending on a specific variable name, i.e., CIRCT_PATH in this case, is not ideal. But this path is a requirement for being able to locate where CIRCT exists when compiling the HLS backend, so a path variable is a requirement anyway.
Having one file that contains the configuration for how to build things makes sense to me, but maybe Makefile.config should be renamed to reflect that it contains configurations that are used by more than make.

An alternative is to instead create a 'build-circt' target in 'Makefile', which calls the './scripts/build-circt.sh' with the configured CIRCT_PATH. This would keep the use of variable names within the make echo system and decouple the build script(s). I'm more in favor of this solution, but you have earlier rejected such a make target.

It's also a question of usability when using the script. If a relative path is given, then you get different results depending on if you run `./scripts/build-circt.sh --enable-hls ./usr` or `cd scripts ; ./build-circt.sh --enable-hls ./usr`. Absolute paths are therefore required when using the script. 

I see your purist view on this, but it hampers ease of use.

phate: Generally, a user must do the following to compile jlm:

1. Build and install dependencies. Let's assume that we only have CIRCT as that is the dependency that is relevant for this PR. Thus, the user would need to run `build-circt.sh` in this step, providing a build-path and an install-path to the script.
2. Compile jlm. This can be split in two substeps:
  a. Configure the build by invoking `configure.sh`, which in turn creates `Makefile.config`. In this step, the user would need to provide the install-path to CIRCT from the step above such that jlm would be able to built with the HLS dialect.
  b. Call `make all`, which uses `Makefile.config` to build jlm.

As I understand it, you would like to automate the transition from step 1 to step 2a by automatically handing over the CIRCT install-path to `configure.sh` instead of letting the user manually hand it over. You want to go about it by extracting the CIRCT_PATH variable from `Makefile.config` in `build-circt.sh`.

The consequence of this is that the order of the compilation process is all in a sudden step 2a, step 1, step 2b from above. In other words, the user would need to configure the jlm build by running `configure.sh` in order to create the `Makefile.config` with a (potentially) not-yet-existent CIRCT install-path. Then the user would need to build and install the CIRCT dependency by running `build-circt.sh`, which would extract the CIRCT install-path from `Makefile.config`. Then the user would be able to compile jlm.

I really think that this is the wrong way to go about this as this is not intuitive. The normal process is build/install dependencies, then configure and build jlm.

If you really want to optimize this, then I think the right approach is to provide another script that performs step 1 and step 2a above (in this order) and automatically hands over the CIRCT install-path between these two steps. In other words, a script that calls `build-circt.sh` followed by `configure.sh`.

The other question that I have not addressed yet is whether it is worth it. You remove the handing over of the CIRCT install-path in the process, but how often do you need to hand over this install-path anyway? At least for me, I have installed the CIRCT dependency once, configured jlm, and since then never invoked `configure.sh` again. We are on the same CIRCT commit since months, i.e., the dependency is not changing (very often). If we would move to another CIRCT commit, then I would need to do it again, but I am not sure whether it is worth optimizing for that rather than spend the time on something else.

sjalander:
@phate 
Removed the dependency on Makefile.config

scripts/build-circt.sh
```diff
@@ -43,12 +48,15 @@ while [[ "$#" -ge 1 ]] ; do
...
+	git clone https://github.com/EECS-NTNU/circt.git ${CIRCT_BUILD}/circt.git
+fi
+cd ${CIRCT_BUILD}/circt.git
 git checkout ${GIT_COMMIT}
-mkdir build
-cd build
-cmake -G Ninja .. \
+mkdir -p  ../build
+cd ../build
+cmake -G Ninja ../circt.git \
```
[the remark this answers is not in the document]
phate: Morever, if I am not mistaken, cmake supports the `-B` command line option, which sets (or creates the build directory). This means you should be able to remove the `cd ../build` above and also the `mkdir -p` above (assuming my comment above was incorrect).

scripts/build-circt.sh
```diff
@@ -24,12 +29,12 @@ while [[ "$#" -ge 1 ]] ; do
 	case "$1" in
 		--build-path)
 			shift
-			CIRCT_BUILD="$1"
+			CIRCT_BUILD="${PWD}/$1"
```
phate, line 32: Should the default not be `${CIRCT_BUILD_DIR}`

scripts/build-circt.sh
```diff
@@ -24,12 +29,12 @@ while [[ "$#" -ge 1 ]] ; do
...
 			shift
 			;;
 		--install-path)
 			shift
-			CIRCT_INSTALL="$1"
+			CIRCT_INSTALL="${PWD}/$1"
```
phate, line 37: Should the default not be `${CIRCT_INSTALL_DIR}`

phate (changes requested):
If it is possible, we should avoid the necessity for mentally walking the folder structure in the script to understand it.

scripts/build-circt.sh
```diff
@@ -43,12 +48,15 @@ while [[ "$#" -ge 1 ]] ; do
 	esac
 done
 
-git clone https://github.com/EECS-NTNU/circt.git ${CIRCT_BUILD}
-cd ${CIRCT_BUILD}
+if [ ! -d "$CIRCT_BUILD/circt.git" ] ;
+then
+	git clone https://github.com/EECS-NTNU/circt.git ${CIRCT_BUILD}/circt.git
```
phate, line 53: I have not tested it, but what happens if you try to `git clone` something into a repository that is already there? In other words, is it possible to remove the `if [.....` statement and let `git clone` just do its thing?

scripts/build-circt.sh
```diff
@@ -3,6 +3,11 @@ set -eu
 
 GIT_COMMIT=093cdfe482530623fea01e1d3242af93e533ba54
 
+# Get the absolute path to this script and set default build and install paths
+SCRIPT_DIR="$(dirname "$(realpath "$0")")"
+CIRCT_BUILD=${SCRIPT_DIR}/../build-circt
+CIRCT_INSTALL=${SCRIPT_DIR}/../usr
```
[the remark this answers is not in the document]
sjalander: Changed

scripts/build-circt.sh
```diff
@@ -43,12 +48,15 @@ while [[ "$#" -ge 1 ]] ; do
 	esac
 done
 
-git clone https://github.com/EECS-NTNU/circt.git ${CIRCT_BUILD}
-cd ${CIRCT_BUILD}
+if [ ! -d "$CIRCT_BUILD/circt.git" ] ;
```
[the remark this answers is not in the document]
sjalander: Changed

scripts/build-circt.sh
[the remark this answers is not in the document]
sjalander: Changed

scripts/build-circt.sh
```diff
@@ -43,12 +48,15 @@ while [[ "$#" -ge 1 ]] ; do
...
+then
+	git clone https://github.com/EECS-NTNU/circt.git ${CIRCT_BUILD}/circt.git
+fi
+cd ${CIRCT_BUILD}/circt.git
```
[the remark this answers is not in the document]
sjalander: Keeping this line makes it possible to use conventional git syntax instead of relying on git commands that is likely not used by many and adds complexity.

scripts/build-circt.sh
```diff
@@ -43,12 +48,15 @@ while [[ "$#" -ge 1 ]] ; do
...
 git checkout ${GIT_COMMIT}
-mkdir build
-cd build
-cmake -G Ninja .. \
+mkdir -p  ../build
```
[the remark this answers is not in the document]
sjalander: A separate build directory is used for building outside the source directory. 
Same as jlm is build in build-release/build-debug.
I've opted for the following directory structure
```
CIRCT_BUILD
    build     # CIRCT_BUILD_DIR
    circt.git # CIRCT_GIT_DIR
```

An option would be to clone into CIRCT_BUILD and then only have a build subdirectory, CIRCT_BUILD/build.

I find the current setup to me more clear, since a user would expect to find the build in CIRCT_BUILD, but we also need to place the source somewhere. So now when somebody looks into the build directory, it should be rather obvious what is what.

scripts/build-circt.sh
```diff
@@ -43,12 +48,15 @@ while [[ "$#" -ge 1 ]] ; do
...
 done
 
-git clone https://github.com/EECS-NTNU/circt.git ${CIRCT_BUILD}
-cd ${CIRCT_BUILD}
+if [ ! -d "$CIRCT_BUILD/circt.git" ] ;
+then
+	git clone https://github.com/EECS-NTNU/circt.git ${CIRCT_BUILD}/circt.git
+fi
+cd ${CIRCT_BUILD}/circt.git
 git checkout ${GIT_COMMIT}
```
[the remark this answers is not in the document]
sjalander: Yes, but this is syntax I've never seen before and simply adds complexity to the script in order to save a single `cd` command

scripts/build-circt.sh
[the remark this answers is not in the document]
sjalander: Improved

scripts/build-circt.sh
sjalander: This is user input and not the default.

scripts/build-circt.sh
sjalander: This is user input and not the default.

scripts/build-circt.sh
sjalander: git clone fails if the directory already exists.

phate (approved):

PR #402, opened 2024-02-20
[Andersen] Add and utilize jlm build composite action

phate:
This PR adds a composite action for building jlm. This PR lays the foundation for the restructuring of our CI tests.

### DIFF
--- a/.github/actions/BuildJlm/action.yml
+++ b/.github/actions/BuildJlm/action.yml
@@ -65,7 +65,7 @@ runs:
 
     - name: "Configure jlm"
       run: |
-        echo "Configure arguments: ${JLM_CONFIGURE_ARGUMENTS}"
+        echo "Configure arguments: $JLM_CONFIGURE_ARGUMENTS"
         ./configure.sh ${JLM_CONFIGURE_ARGUMENTS}
       shell: bash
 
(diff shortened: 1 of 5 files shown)


.github/actions/BuildJlm/action.yml
```diff
@@ -0,0 +1,83 @@
...
+
+    - name: "Install MLIR dialect dependencies"
+      if: ${{inputs.enable-mlir == 'true'}}
+      uses: ./.github/actions/BuildMlirDialect
+
+    - name: "Collect command line arguments"
+      id: collect
+      run: |
+        export JLM_CONFIGURE_ARGUMENTS="--target release --enable-asserts"
+        if [[ "${{inputs.enable-hls}}" == "true" ]]; then
```
haved, line 60: Using bash if statements here and GitHub yaml if statements above is perhaps not ideal, but I can't come up with an alternative that avoids it, without creating a bunch of step outputs.

haved (approved):
Looks very clean, nice improvement!

.github/actions/BuildJlm/action.yml
phate: I had the same problem. The ugliness is very contained and local. So I did not worry too much about it.

PR #408, opened 2024-02-27
Improve sanity checks in Steensgaard

phate:
This PR does the following:

1. Improves the sanity checks in Steensgaard
2. Minor refactoring for better readability


haved (approved):
Very nice changes.
Now that Steensgaard tracks pointers in `IsOrContains<PointerType>` types, it also makes it much easier to compare the PointsToGraphs produced by Andersen and Steensgard, since the set of `rvsdg::output`s being mapped to `PointsToGraph::RegisterNode`s should be the same.

PR #409, opened 2024-02-27
[Andersen] Refactor InstallLlvmDependencies composite action

phate:
This PR does the following:

1. Renames InstallLlvmDependencies to InstallPackages
2. Introduces inputs to the composite action to enable the installation of specific packages


haved (approved):
Looks good, my only question is if repeated invocations of `apt-get update` take up any substantial amount of time, or if it is fast enough to ignore. Otherwise there could be a single step that always does `apt-get update` after setting up the llvm apt repo, before installing anything

phate:
I had auto-merge enabled. I will address it in another PR.

PR #410, opened 2024-02-29
Clean up Steensgaard context

phate:


haved (approved):
I like the renaming, and it's nice to be explicit about when you are getting and when you are inserting locations.

PR #412, opened 2024-02-29
Introduce CIRCT_LDFLAGS

### LINKED ISSUE #398: Introduce CIRCT_LDFLAGS
We currently add the necessary libraries for CIRCT individually. It is better to introduce a CIRCT_LDFLAGS into configure.sh in order to set them there and then just use them throughout the makefiles.

See [here](https://github.com/phate/jlm/blob/beff174d657d4df27a51a5e4371ef8171664f4fb/tools/jlm-hls/Makefile.sub#L18C1-L30C18) for the flags.

phate:
Closes #398


haved (approved):
LGTM

PR #413, opened 2024-02-29
Clean up serialization interface of Statistics class

### LINKED ISSUE #361: Improve statistics serialization

phate:
This is the last PR in series of PRs that refactored the Statistics class serialization.

The other PRs are #384 and #382. Closes #361.

### DIFF
--- a/jlm/util/Statistics.cpp
+++ b/jlm/util/Statistics.cpp
@@ -64,15 +64,13 @@ Statistics::GetSourceFile() const
 }
 
 std::string
-Statistics::ToString() const
-{
-  return util::strfmt(GetName(), " ", GetSourceFile().to_str(), " ", Serialize());
-}
-
-std::string
 Statistics::Serialize() const
 {
   std::ostringstream ss;
+
+  ss << GetName() << " ";
+  ss << GetSourceFile().to_str() << " ";
+
   for (const auto & [mName, measurement] : Measurements_)
   {
     if (ss.tellp() != 0)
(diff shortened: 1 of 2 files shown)


haved (approved):
Nice, I like the idea of allowing custom separators

PR #415, opened 2024-03-01
Update InstallPackages composite action

phate:
The PR does the following:

1. Removes `sudo apt-get update`. It was unnecessary
2. Uses pip install lit instead of setup.py. Using setup.py is deprecated.
3. Only adds the llvm repository when it is not already present (multiple invocations of the script can lead to it)
4. Adds the possibility for installing doxygen and verilator to the script.


haved (approved):
Looks good, I'm surprised you don't have to do `apt-get update` after adding a repository, but I guess `add-apt-repository` does it for you

PR #418, opened 2024-03-02
Add IsSupergraphOf to PointsToGraph

haved:
This is used as part of double checking PointsToGraphs created by the naive constraint solver and more sophisticated solvers. If you think some of the lambdas should be available as separate public methods (or private methods) let me know.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
+++ b/jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
   return *tmp;
 }
 
+bool
+PointsToGraph::IsSupergraphOf(const jlm::llvm::aa::PointsToGraph & subgraph) const
+{
+  // Given a memory node representing a memory object in an RVSDG module, this function finds
+  // a memory node representing the same memory object in a different PointsToGraph.
+  // If no corresponding memory node exists in the graph, nullptr is returned
+  const auto GetCorrespondingMemoryNode =
+      [](const PointsToGraph::MemoryNode & node,
+         const PointsToGraph & graph) -> const PointsToGraph::MemoryNode *
+  {
+    if (auto allocaNode = dynamic_cast<const AllocaNode *>(&node))
+    {
+      if (auto it = graph.AllocaNodes_.find(&allocaNode->GetAllocaNode());
+          it != graph.AllocaNodes_.end())
+        return it->second.get();
+    }
+    else if (auto deltaNode = dynamic_cast<const DeltaNode *>(&node))
+    {
+      if (auto it = graph.DeltaNodes_.find(&deltaNode->GetDeltaNode());
+          it != graph.DeltaNodes_.end())
+        return it->second.get();
+    }
+    else if (auto importNode = dynamic_cast<const ImportNode *>(&node))
+    {
+      if (auto it = graph.ImportNodes_.find(&importNode->GetArgument());
+          it != graph.ImportNodes_.end())
+        return it->second.get();
+    }
+    else if (auto lambdaNode = dynamic_cast<const LambdaNode *>(&node))
+    {
+      if (auto it = graph.LambdaNodes_.find(&lambdaNode->GetLambdaNode());
+          it != graph.LambdaNodes_.end())
+        return it->second.get();
+    }
+    else if (auto mallocNode = dynamic_cast<const MallocNode *>(&node))
+    {
+      if (auto it = graph.MallocNodes_.find(&mallocNode->GetMallocNode());
+          it != graph.MallocNodes_.end())
+        return it->second.get();
+    }
+    else if (MemoryNode::Is<UnknownMemoryNode>(node))
+    {
+      return &graph.GetUnknownMemoryNode();
+    }
+    else if (MemoryNode::Is<ExternalMemoryNode>(node))
+    {
+      return &graph.GetExternalMemoryNode();
+    }
+    else
+      JLM_UNREACHABLE("Unknown type of MemoryNode");
+
+    return nullptr;
+  };
+
+  // Given two nodes, checks if the first node points to everything the second points to.
+  // The nodes can belong to different PointsToGraphs.
+  const auto HasSupersetOfPointees = [&GetCorrespondingMemoryNode](
+                                         const PointsToGraph::Node & superset,
+                                         const PointsToGraph::Node & subset)
+  {
+    for (const auto & subsetTarget : subset.Targets())
+    {
+      auto correspondingTarget = GetCorrespondingMemoryNode(subsetTarget, superset.Graph());
+
+      // Check if the superset is pointing to the target
+      if (correspondingTarget == nullptr || !superset.HasTarget(*correspondingTarget))
+        return false;
+    }
+    return true;
+  };
+
+  // Given a memory node from the subgraph, check that a corresponding node exists in this graph.
+  // All edges the other node has, must have corresponding edges in this graph.
+  // If the other node is marked as escaping, the corresponding node must be as well.
+  const auto HasSuperOfMemoryNode = [&](const PointsToGraph::MemoryNode & subNode)
+  {
+    const auto thisNode = GetCorrespondingMemoryNode(subNode, *this);
+    if (thisNode == nullptr)
+      return false;
+
+    if (subNode.IsModuleEscaping() && !thisNode->IsModuleEscaping())
+      return false;
+
+    return HasSupersetOfPointees(*thisNode, subNode);
+  };
+
+  // Iterate through all memory nodes in the subgraph, and make sure we have corresponding nodes
+  for (const auto & node : subgraph.AllocaNodes())
+  {
+    if (!HasSuperOfMemoryNode(node))
+      return false;
+  }
+  for (const auto & node : subgraph.DeltaNodes())
+  {
+    if (!HasSuperOfMemoryNode(node))
+      return false;
+  }
+  for (const auto & node : subgraph.ImportNodes())
+  {
+    if (!HasSuperOfMemoryNode(node))
+      return false;
+  }
+  for (const auto & node : subgraph.LambdaNodes())
+  {
+    if (!HasSuperOfMemoryNode(node))
+      return false;
+  }
+  for (const auto & node : subgraph.MallocNodes())
+  {
+    if (!HasSuperOfMemoryNode(node))
+      return false;
+  }
+
+  // For each register mapped to a RegisterNode in the subgraph, this graph must also have mapped
+  // the same register to be a supergraph. The RegisterNode must point to a superset of what
+  // the subgraph's RegisterNode points to.
+  for (const auto [rvsdgOutput, subNode] : subgraph.RegisterNodeMap_)
+  {
+    auto correspondingRegisterNode = RegisterNodeMap_.find(rvsdgOutput);
+    if (correspondingRegisterNode == RegisterNodeMap_.end())
+      return false;
+
+    if (!HasSupersetOfPointees(*correspondingRegisterNode->second, *subNode))
+      return false;
+  }
+
+  return true;
+}
+
 std::string
 PointsToGraph::ToDot(
     const PointsToGraph & pointsToGraph,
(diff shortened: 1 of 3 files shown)


jlm/llvm/opt/alias-analyses/PointsToGraph.hpp
```diff
@@ -350,6 +350,17 @@ class PointsToGraph final
   PointsToGraph::ImportNode &
   AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node);
 
+  /**
+   * Checks if this PointsToGraph is a supergraph of the given PointsToGraph.
```
phate, line 354: "is a supergraph of \p subgraph"

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -305,6 +440,12 @@ PointsToGraph::Node::Sources() const
   return { SourceConstIterator(Sources_.begin()), SourceConstIterator(Sources_.end()) };
 }
 
+bool
+PointsToGraph::Node::HasSource(const PointsToGraph::Node & source) const
+{
+  return Sources_.find(const_cast<Node *>(&source)) != Sources_.end();
```
phate, line 446: Is this const_cast necessary to get it compiling?

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -293,6 +422,12 @@ PointsToGraph::Node::Targets() const
   return { TargetConstIterator(Targets_.begin()), TargetConstIterator(Targets_.end()) };
 }
 
+bool
+PointsToGraph::Node::HasTarget(const PointsToGraph::MemoryNode & target) const
+{
+  return Targets_.find(const_cast<MemoryNode *>(&target)) != Targets_.end();
```
phate, line 428: Is this const_cast necessary to get it compiling?

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
...
+      return &graph.GetUnknownMemoryNode();
+    }
+    else if (MemoryNode::Is<ExternalMemoryNode>(node))
+    {
+      return &graph.GetExternalMemoryNode();
+    }
+    else
+      JLM_UNREACHABLE("Unknown type of MemoryNode");
+
+    return nullptr;
```
phate, line 215: This could just be `JLM_UNREACHABLE(...)`, no? Remove the else (and the return nullptr), and just put the UNREACHABLE at the end?

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
...
+    for (const auto & subsetTarget : subset.Targets())
+    {
+      auto correspondingTarget = GetCorrespondingMemoryNode(subsetTarget, superset.Graph());
+
+      // Check if the superset is pointing to the target
+      if (correspondingTarget == nullptr || !superset.HasTarget(*correspondingTarget))
+        return false;
+    }
+    return true;
+  };
```
phate, line 233: You might want to do the cheap check using the number of targets first, before you iterate through all of them?

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
...
+
+  // Given a memory node from the subgraph, check that a corresponding node exists in this graph.
+  // All edges the other node has, must have corresponding edges in this graph.
+  // If the other node is marked as escaping, the corresponding node must be as well.
+  const auto HasSuperOfMemoryNode = [&](const PointsToGraph::MemoryNode & subNode)
+  {
+    const auto thisNode = GetCorrespondingMemoryNode(subNode, *this);
```
phate, line 240: `auto thisNode =`

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
...
+    return nullptr;
+  };
+
+  // Given two nodes, checks if the first node points to everything the second points to.
+  // The nodes can belong to different PointsToGraphs.
+  const auto HasSupersetOfPointees = [&GetCorrespondingMemoryNode](
+                                         const PointsToGraph::Node & superset,
+                                         const PointsToGraph::Node & subset)
+  {
+    for (const auto & subsetTarget : subset.Targets())
```
phate, line 224: `auto &`

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
   return *tmp;
 }
 
+bool
+PointsToGraph::IsSupergraphOf(const jlm::llvm::aa::PointsToGraph & subgraph) const
+{
+  // Given a memory node representing a memory object in an RVSDG module, this function finds
+  // a memory node representing the same memory object in a different PointsToGraph.
+  // If no corresponding memory node exists in the graph, nullptr is returned
+  const auto GetCorrespondingMemoryNode =
```
phate, line 170: `auto ...`

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
...
+    }
+    else
+      JLM_UNREACHABLE("Unknown type of MemoryNode");
+
+    return nullptr;
+  };
+
+  // Given two nodes, checks if the first node points to everything the second points to.
+  // The nodes can belong to different PointsToGraphs.
+  const auto HasSupersetOfPointees = [&GetCorrespondingMemoryNode](
```
phate, line 220: `auto ...`

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
...
+      if (correspondingTarget == nullptr || !superset.HasTarget(*correspondingTarget))
+        return false;
+    }
+    return true;
+  };
+
+  // Given a memory node from the subgraph, check that a corresponding node exists in this graph.
+  // All edges the other node has, must have corresponding edges in this graph.
+  // If the other node is marked as escaping, the corresponding node must be as well.
+  const auto HasSuperOfMemoryNode = [&](const PointsToGraph::MemoryNode & subNode)
```
phate, line 238: `auto ...`

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
...
+  for (const auto & node : subgraph.MallocNodes())
+  {
+    if (!HasSuperOfMemoryNode(node))
+      return false;
+  }
+
+  // For each register mapped to a RegisterNode in the subgraph, this graph must also have mapped
+  // the same register to be a supergraph. The RegisterNode must point to a superset of what
+  // the subgraph's RegisterNode points to.
+  for (const auto [rvsdgOutput, subNode] : subgraph.RegisterNodeMap_)
```
phate, line 280: `auto ...`

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
...
+      return false;
+
+    if (subNode.IsModuleEscaping() && !thisNode->IsModuleEscaping())
+      return false;
+
+    return HasSupersetOfPointees(*thisNode, subNode);
+  };
+
+  // Iterate through all memory nodes in the subgraph, and make sure we have corresponding nodes
+  for (const auto & node : subgraph.AllocaNodes())
```
phate, line 251: `auto ...`

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
...
+  {
+    if (!HasSuperOfMemoryNode(node))
+      return false;
+  }
+  for (const auto & node : subgraph.DeltaNodes())
```
phate, line 256: `auto ...`

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
...
+  {
+    if (!HasSuperOfMemoryNode(node))
+      return false;
+  }
+  for (const auto & node : subgraph.ImportNodes())
```
phate, line 261: `auto ...`

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
...
+  {
+    if (!HasSuperOfMemoryNode(node))
+      return false;
+  }
+  for (const auto & node : subgraph.LambdaNodes())
```
phate, line 266: `auto ...`

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
...
+  {
+    if (!HasSuperOfMemoryNode(node))
+      return false;
+  }
+  for (const auto & node : subgraph.MallocNodes())
```
phate, line 271: `auto ...`

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
```diff
@@ -161,6 +161,135 @@ PointsToGraph::AddImportNode(std::unique_ptr<PointsToGraph::ImportNode> node)
...
+    if (thisNode == nullptr)
+      return false;
+
+    if (subNode.IsModuleEscaping() && !thisNode->IsModuleEscaping())
+      return false;
+
+    return HasSupersetOfPointees(*thisNode, subNode);
+  };
+
+  // Iterate through all memory nodes in the subgraph, and make sure we have corresponding nodes
```
phate, line 250: Maybe compare the number of various nodes between the two graphs first before you iterate through everything?

phate (approved):
LGTM. I added some minor suggestions.

The biggest question I have is why `IsSupergraphOf()` and not `IsEquivalentTo()`? Should the graphs not be equivalent between the different solvers, and the only difference should be the time-to-solution?

haved:
It would be relatively easy to convert this function to an `IsEquivalentTo` by checking that there is an equal number of memory nodes, registers and edges while also checking that one is a subset. `IsEquivalentTo` can also be implemented as calling `IsSupergrafOf` both ways, with around double the runtime, but still linear, and it's not a hot function.

The reason I want IsSupergrafOf is that some of the offline processing might make the analysis ever so slightly less precise, in programs that contain reads and writes from variables that always point to NULL. These programs are not necessarily I'll formed, if the reads and writes never execute, but our static Analysis may have issues.

IsSupergrafOf can also be used to compare the output of Andersen and Steensgaard.

In general it felt more flexible to have the option to compare <= and not always ==

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
haved: Yes, sadly. The alternative is to set up a "transparent comparator", which would modify the type of `Sources_` with extra template stuff. See this [Stack Overflow Answer](https://stackoverflow.com/questions/61192917/c-stdunordered-set-find-count-contains-const-key-reference)

jlm/llvm/opt/alias-analyses/PointsToGraph.cpp
haved: No, all the other if bodies are supposed to return `nullptr` if they don't find a match. I can write `return nullptr` in each of them if it is clearer.

phate: You are absolutely right. I misread the code. Thanks for clarifying. LGTM.

phate (approved):
Thanks for the clarifications.

PR #419, opened 2024-03-03
[Andersen] add PointerObjectConstraintSet solver using worklist algorithm

haved:
This PR comes on top of #418, I will move its base once that is merged.

This PR modifies the existing naive solver slightly to enable code re-use using some templated functions.
It is also designed with future expansions in mind (e.g. difference propagation and unification), which might explain some peculiar choices.

`HandleEscapingFunctionConstraint` is removed as a constraint, as it was previously added for every function in the program. It made more sense to make it a global constraint, and not something you individually add per function.

The worklist solver is made the default in Andersen's configuration.
Currently, double-checking against the naive solver is controlled by an environment variable, as it is purely a debug thing, and feels like a nice mix between a compiler directive and a command line flag added to jlm. Let me know if you prefer it in a different way.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/Andersen.cpp
+++ b/jlm/llvm/opt/alias-analyses/Andersen.cpp
@@ -861,11 +861,24 @@ Andersen::Analyze(const RvsdgModule & module, util::StatisticsCollector & statis
   if (std::getenv(CHECK_AGAINST_NAIVE_SOLVER) != nullptr
       && Config_ != Configuration::DefaultNaiveConfiguration())
   {
+    std::cerr << "Comparing Andersen's PointsToGraph to a naivley solved PointsToGraph" << std::endl;
+
     // Temporarily switch to the default naive configuration
     auto customConfig = Config_;
     Config_ = Configuration::DefaultNaiveConfiguration();
 
     auto naivePointsToGraph = AnalyzeModule(module, statisticsCollector);
+
+    bool error = false;
+    if (!naivePointsToGraph->IsSupergraphOf(*pointsToGraph)) {
+      std::cerr << "The naive PointsToGraph is NOT a supergraph of the PointsToGraph" << std::endl;
+      error = true;
+    }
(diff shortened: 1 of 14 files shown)


jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -845,6 +852,27 @@ Andersen::Analyze(const RvsdgModule & module, util::StatisticsCollector & statis
...
+    Config_ = Configuration::DefaultNaiveConfiguration();
+
+    auto naivePointsToGraph = AnalyzeModule(module, statisticsCollector);
+
+    Config_ = customConfig;
+  }
+
+  return pointsToGraph;
+}
+
```
phate, line 875: Did you forget the code for raising an error if the two points-to graphs diverge? Right now, the computed `naivePointsToGraph` is not used for anything (as far as I can see).

phate (changes requested):
The only reason why I request changes is because I do not see that an actual error is raised when the points-to graph differ.

I am not happy with the environment variable approach, but since this is only meant for testing and will only ever be used by you/me it is absolutely fine. It is a quick hack that does the trick without having to touch too many files/code lines.

Otherwise, LGTM.

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -463,6 +592,172 @@ PointerObjectConstraintSet::GetConstraints() const noexcept
   return Constraints_;
 }
 
+size_t
+PointerObjectConstraintSet::SolveUsingWorklist()
+{
+  size_t numWorkItems = 0;
```
phate, line 598: Could you move this directly over the main loop. It is used there for the first time. Currently, you have ca. 100 lines of code between its definition and the first usage.

haved:
I tested on the subset of `llvm-test-suite` we use for Andersen, and all tests reported identical PointsToGraphs, so that's nice :)

phate (approved):

PR #420, opened 2024-03-16
[Andersen] change the PointerObjectSet interface to enable unification

haved:
This is the first step in enabling both offline and online variable unification. You no longer have access to `PointerObeject`s directly, so all access goes through the set.

The global `constexpr bool ENABLE_UNIFICATION` allows quickly disabling unification logic, to check how much performance is lost due to extra lookups associated with the disjoint-set data structure, when unification isn't being used. I'm fairly confident that unification enables net performance gains. (Solvers that maintain their own unification invariants can get direct access to points-to sets to avoid the penalty, if it is too great).

Variables are not fully unified, they only start sharing points-to set. The old variables still exist, since other points-to sets may still contain only some of the variables in the unification, or only a subset of the variables have the escaped flag.

Variables where `CanPoint()` is false may not be part of unifications, since they have special meaning. (I'll probably want to rename `CanPoint()` to `TracksPointees()`, as that is a more descriptive name. If you have a variable with `TracksPointees()=false` in your SCC I have other ideas for handling that)

### DIFF
--- a/jlm/llvm/opt/alias-analyses/Andersen.cpp
+++ b/jlm/llvm/opt/alias-analyses/Andersen.cpp
@@ -779,11 +779,11 @@ Andersen::AnalyzeRvsdg(const rvsdg::graph & graph)
 
     // Create a memory PointerObject representing the target of the external symbol
     // We can assume that two external symbols don't alias, clang does.
-    const PointerObject::Index importObject = Set_->CreateImportMemoryObject(argument);
+    const auto importObjectPO = Set_->CreateImportMemoryObject(argument);
 
     // Create a register PointerObject representing the address value itself
-    const PointerObject::Index importRegister = Set_->CreateRegisterPointerObject(argument);
-    Constraints_->AddPointerPointeeConstraint(importRegister, importObject);
+    const auto importRegisterPO = Set_->CreateRegisterPointerObject(argument);
+    Constraints_->AddPointerPointeeConstraint(importRegisterPO, importObjectPO);
   }
 
   AnalyzeRegion(rootRegion);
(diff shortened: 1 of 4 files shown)


haved:
It seems all builds fail in CI due to CIRCT regression tests being unable to run.

I'm a bit puzzled by the path mentioned in the error:
```
/usr/bin/python3.10: can't open file '/home/runner/work/jlm/jlm/home/runner/work/jlm/jlm/build-circt/build/bin/llvm-lit': [Errno 2] No such file or directory
```
As it seems circt has been cloned in an accidentally deep path, but I don't think that is the actual issue. Is the clone pinned to a specific commit? I haven't really touched anything in this PR that should affect CIRCT.

I'm also curious if we actually intend to run CIRCTs own tests when building jlm in CI. Also building CIRED is the same in all the workflows(?), so we could maybe do it once, cache it, and send it as a job artifact to the other jobs?

phate:
@sjalander Could you take a look, please?

sjalander:
I guess this is the problem

  LLVM_EXTERNAL_LIT set to /usr/local/bin/lit, but the path does not exist.

Not sure why it seems like lit is no longer being installed correctly.

phate (approved):
LGTM. Could elaborate where unification comes into play in the papers, please?

haved:
I'm afraid I introduced some bug that llvm-test-suite-andersen luckily picks up. Will investigate

haved:
I fixed the issue, it was a mistake in the way import nodes are set to always be marked as escaped.

It is annoying to make structural changes, as it affects both the naive solver and any other solvers, so any differences it introduces don't get detected by comparing against naive. Luckily there won't be much more structural changes, only adding solvers and cycle detectors.

haved:
Regarding why Unification was added in the first place, it's about what to do when a cycle is detected in the inclusion constraint graph. Multiple variables will turn into one logical variable in the constraint graph, but the new variable doesn't actually exist, so it can't be a part of points-to sets. Individual members of the new unification may still be a part of points-to sets, however. The number of constraint variables is reduced, since unifications have a shared set and shared constraints, but the number of program variables is the same.

phate:
All right. It makes sense. When there is a cycle in the graph, we just collapse it using unification (equivalent to Steensgaard). Thanks for the clarification.

phate (approved):

PR #421, opened 2024-03-17
Ensure correct flags for escaping registers in Steensgaard

phate:
The contents of registers that escape the module can point to external and all escaped memory, but they were not marked as points-to-external and points-to-escaped so far. This PR fixes this problem.


haved (approved):
MarkAsEscaped looks good, great that it adds the flags to the pointee location. I'm also wondering if maybe the pointee location should be marked as escaped here.

Otherwise, doing escape propagation can also be done as a linear pass after everything else is done.

PR #424, opened 2024-03-19
Add IsInvariant() to gamma output class

phate:
This PR does the following:

1. Adds IsInvariant() method to gamma output class.
2. Replaces all usages of old is_invariant() function.
3. Removes is_invariant() function.


haved (approved):
LGTM. It took me a second to realize that `dynamic_cast<const rvsdg::argument *>` returns `nullptr` unless the origin of the result is actually an argument.

PR #425, opened 2024-03-20
Make unified register PointerObjects share RegisterNode in the PointsToGraph

haved:
Since `RegisterNode`s support mapping multiple RVSDG outputs to a single `PointsToGraph` node, when they all share pointees, it makes sense to use this for unified `PointerObject`s of Register kind.

This PR also adds a test of `Andersen::ConstructPointsToGraphFromPointerObjectSet`, which was missing.

Just for fun, here is the PointsToGraph being built in the test:

ptg

### DIFF
--- a/jlm/llvm/opt/alias-analyses/Andersen.cpp
+++ b/jlm/llvm/opt/alias-analyses/Andersen.cpp
@@ -841,7 +841,7 @@ Andersen::AnalyzeModule(const RvsdgModule & module, util::StatisticsCollector & 
     JLM_UNREACHABLE("Unknown solver");
 
   statistics->StartPointsToGraphConstructionStatistics();
-  auto result = ConstructPointsToGraphFromPointerObjectSet(*Set_, *statistics);
+  auto result = ConstructPointsToGraphFromPointerObjectSet(*Set_, statistics.get());
   statistics->StopPointsToGraphConstructionStatistics(*result);
 
   statistics->StopAndersenStatistics();
@@ -899,7 +899,7 @@ Andersen::Analyze(const RvsdgModule & module)
 std::unique_ptr<PointsToGraph>
 Andersen::ConstructPointsToGraphFromPointerObjectSet(
     const PointerObjectSet & set,
-    Statistics & statistics)
+    Statistics * statistics)
 {
   auto pointsToGraph = PointsToGraph::Create();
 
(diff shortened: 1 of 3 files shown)


jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -899,7 +899,7 @@ Andersen::Analyze(const RvsdgModule & module)
 std::unique_ptr<PointsToGraph>
 Andersen::ConstructPointsToGraphFromPointerObjectSet(
     const PointerObjectSet & set,
-    Statistics & statistics)
+    Statistics * statistics)
```
phate, line 902: Why did you change this from a reference to a pointer?

phate: Is it because of the unit test? Could you not just create one there?

phate (approved):
Only a minor issue.

jlm/llvm/opt/alias-analyses/Andersen.cpp
haved: That is what I started doing, but I realized the Statistics class is only forward declared in the header, while the definition is in `Andersen.cpp`.   Without moving it, or declaring a public `Andersen:: CreateStatistics()`, I can't make instances of Statistics from the test.

I don't know if it is possible to forward declare `Andersen::Statistics::Create()`, but googling doesn't give me any examples of it being allowed.

phate: Could you provide a `ConstructPointsToGraphFromPointerObjectSet()` method that does not take a `Statistics` and internally is defined in the `cpp` file and calls the method version with `Statistics` parameter internally?

haved: Good point, that works

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -1003,4 +1006,12 @@ Andersen::ConstructPointsToGraphFromPointerObjectSet(
   return pointsToGraph;
 }
 
+std::unique_ptr<PointsToGraph>
+Andersen::ConstructPointsToGraphFromPointerObjectSet(const PointerObjectSet & set)
+{
+  // Create a throwaway instance of statistics
+  auto statistics = Statistics::Create(jlm::util::filepath(""));
```
phate, line 1013: There is no need to create an instance on the heap here. It could be on the stack, no?

phate (dismissed):
I would rather create the statistics on the stack instead of the heap here.

phate (approved):

PR #426, opened 2024-03-24
Mark variadic arguments of external calls as escaped in Steensgaard

phate:
The variadic arguments of external function calls were so far not handled and marked as escaped in Steensgaard. This PR fixes this.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/Steensgaard.cpp
+++ b/jlm/llvm/opt/alias-analyses/Steensgaard.cpp
@@ -1004,9 +1004,11 @@ Steensgaard::AnalyzeSimpleNode(const jlm::rvsdg::simple_node & node)
   {
     AnalyzeExtractValue(node);
   }
-  else if (is<FreeOperation>(&node) || is<ptrcmp_op>(&node))
+  else if (is<FreeOperation>(&node) || is<ptrcmp_op>(&node) || is<valist_op>(&node))
   {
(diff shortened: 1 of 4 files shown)


phate:
@haved : Looking at [the implementation of Andersen for this](https://github.com/phate/jlm/blob/ec233dda0c80389c5393f0b83ccbee220e89ed88/jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp#L412C1-L438C2), it does not seem to handle this either.

sjalander (approved):

haved:
@phate I had forgotten how I handled this, but looking at it now in `Andersen.cpp` I can see that /anything/ that enters a `valist` node is marked as escaping, no matter how direct the function call is.

The reason is explained in a comment in `AnalyzeValist`: The valist is read using the `va_start` macros and friends, which involves external function calls. I never added special handling of this, which means everything extracted from the valist points to external. At callsite it is therefore enough to mark all values entering the va_list as escaped.

phate:
@haved : Thanks for the clarification. I think that should do the trick even though variadic arguments are handled weird on x86. I will create another unit test for it and also have a look at what Andersen gives me on it.

haved:
@phate I didn't really think too hard about how variadic arguments are handled in the ABI, I only looked at the IR produced by clang with -O0. I'm hoping no optimization pass lowers `va_start` into pure stack manipulation, as such lowering has to be a part of the backend as far as I know.

PR #427, opened 2024-03-30
Fix handling of variadic arguments in Steensgaard

phate:
This PR does the following:

1. Fixes the handling of variadic arguments in Steensgaard for good.
2. Adds some RVSDG creators for different operations such that they are easier to create in the added unit test
3. Fixes the LLVM print function in test-util.hpp
4. Adds a variadic argument unit test.

### DIFF
--- a/jlm/llvm/ir/operators/operators.hpp
+++ b/jlm/llvm/ir/operators/operators.hpp
@@ -1798,6 +1798,22 @@ public:
     ConstantArray op(*vt, elements.size());
     return tac::create(op, elements);
   }
+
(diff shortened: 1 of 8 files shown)


haved (approved):
Looks great, and should be fully consistent with how Andersen does it. One could imagine some kind of special handling of the `va_start` function to not have to escape the valist, but there are lots of other functions that could benefit from special handling, so that's nothing new.

It's very nice to have a test rvsdg module that reads the va_list. I will port that test to Andersen as well.

PR #428, opened 2024-04-02
More expressive bitvalue comparison error messages

ProgrammingLouis:
Really small pull request that changes error messages in `rvsdg/bitstring/value-representation.hpp` to specify operation name in "Unequal number of bits" errors.

### DIFF
--- a/jlm/rvsdg/bitstring/value-representation.hpp
+++ b/jlm/rvsdg/bitstring/value-representation.hpp
@@ -169,7 +169,7 @@ private:
     JLM_ASSERT(remainder == 0);
 
     if (divisor.nbits() != nbits())
-      throw jlm::util::error("Unequal number of bits.");
+      throw jlm::util::error("Unequal number of bits in udiv.");
 
     /*
       FIXME: This should check whether divisor is zero, not whether nbits() is zero.
@@ -382,7 +382,7 @@ public:
   ult(const bitvalue_repr & other) const
   {
     if (nbits() != other.nbits())
-      throw jlm::util::error("Unequal number of bits.");
+      throw jlm::util::error("Unequal number of bits in ult.");
 
     char v = land(lnot(data_[0]), other[0]);
     for (size_t n = 1; n < nbits(); n++)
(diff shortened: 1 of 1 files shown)


jlm/rvsdg/bitstring/value-representation.hpp
```diff
@@ -426,7 +426,11 @@ class bitvalue_repr
   ne(const bitvalue_repr & other) const
   {
     if (nbits() != other.nbits())
-      throw jlm::util::error("Unequal number of bits.");
+    {
+      const std::string & msg = "Unequal number of bits in ne, " + std::to_string(nbits())
+                              + " != " + std::to_string(other.nbits());
```
phate, line 431: Rather use `strfmt`, please:
`jlm::util::strfmt("Unequal number of bits in ne: ", nbits(), " != ", other.nbits());`

See [strfmt.hpp](https://github.com/phate/jlm/blob/master/jlm/util/strfmt.hpp)

phate (changes requested):

jlm/rvsdg/bitstring/value-representation.hpp
ProgrammingLouis: Resolved in next commit

phate (changes requested):
I think you misunderstood my comment. I would like you to use jlm's [strfmt](https://github.com/phate/jlm/blob/48f628cef71a9e5a39c4959a8086e5fbff8c62ac/jlm/util/strfmt.hpp#L35C1-L35C21) to assemble the exception message. In other words, something like this: `jlm::util::error(util::strmft("Unequal number of bits in udiv", divisor.nbits(), " != ", nbits()))`. This has also the benefit of getting rid of all the `std::to_string()`s in the code. Moreover, it will make it easier to convert to std::format() once we move over to C++20.

phate (approved):

ProgrammingLouis:
> Missing header....

I'm really sorry for that, I don't want to take to much of your time for that.
I think this is fixed.

sjalander (approved):

PR #429, opened 2024-04-04
Fix various issues in Steensgaard

phate:
The PR fixes the following issues:

1. Added handling of missing ptr2bits operation
2. Fix handling of ConstantArray and ConstantStruct. The insertion of the output location was performed in the loop for the nodes' inputs. This would lead to multiple creation attempts in case of multiple pointer inputs and therefore fail. Moreover, the handling of the nodes was changed by analyzing types instead of checking for previously created locations. This is more intuitive and in line with the handling of all other nodes.
3. Utilize GetOrInsertRegisterLocation() everywhere instead of InsertRegisterLocation(). The opposite was done in a recent PR, i.e., utilize InsertRegisterLocation() instead of the more general GetOrInsertRegisterLocation(), but this was simply incorrect and leads to bugs in case of recursive functions. I also added some documentation of why simply using InsertRegisterLocation() is incorrect such that I won't forget it again and make the same mistake in future.

I provided no unit tests for this PR as these problems were discovered with the LLVM test suite. I will provide another PR soon where I add a check for the SteengaardAgnostic pass to the CI. The above issues would be discovered in this check if they would reappear.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/Steensgaard.cpp
+++ b/jlm/llvm/opt/alias-analyses/Steensgaard.cpp
@@ -623,36 +623,58 @@ public:
     return *location;
   }
 
-  RegisterLocation &
-  InsertRegisterLocation(const jlm::rvsdg::output & output, PointsToFlags pointsToFlags)
-  {
-    JLM_ASSERT(!HasRegisterLocation(output));
-
-    auto registerLocation = RegisterLocation::Create(output, pointsToFlags);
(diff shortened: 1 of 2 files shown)


haved (approved):
Looks good, the comment nicely explains how relying on outputs being visited doesn't always work in phi nodes. In Andersen I dodged this issue by not handling direct calls while analyzing the module, and instead connecting arguments and parameters, and return values and call node outputs, once solving starts.

PR #430, opened 2024-04-05
Clean up RegionAwareMemoryNodeProvider

phate:
1. Moves the RegionAwareMemoryNodeProvider statistics class from the hpp to the cpp file.
2. Changes jump table usages to if-then-else cascades for node handling.
3. Some other minor syntactic changes, such as removing superfluous namespace qualifications, removing dead headers, changing comments from `/* */` to `//`, etc.


haved (approved):
Looks good to me

PR #433, opened 2024-04-09
Ensure correct creation of node type when load/store/call nodes are copied

### LINKED ISSUE #193: -Jpsh -JAASteensgaardRegionAware  segfaults
jlc -Jpsh -JAASteensgaardRegionAware psh.c

[push.zip](https://github.com/phate/jlm/files/10610713/push.zip)

phate:
This PR does the following:

1. Overloads the copy() methods for the LoadNode, StoreNode, and CallNode class. This was necessary to ensure that the right node is actually created.
2. Provides unit tests for this issue for all three nodes.
3. Minor refactoring of existing unit tests files for these three nodes.

Closes #193


haved (approved):
LGTM. I'm guessing there wasn't a need for a `CreateNode` static method on the load node, since only the two others got it?

phate:
@haved I added the missing `CreateNode` method on the `LoadNode` class.

haved (approved):
LGTM

PR #434, opened 2024-04-09
Use clone of Andersen's variable and constraint set when double checking

haved:
I had some issues where the order nodes were traversed seemed different in the "main" analysis and the "double check" analysis. To make debugging easier, the double check now makes use of a copy of the constraint set, instead of creating the constraint set again.

@phate I tried looking into the `rvsdg::topdown_traverser`, and realized I might have misunderstood it. I initially thought it traversed the region in a topological order (all predecessors of X are visited before X), but now I'm not quite sure, as it kind of looks like "at least one predecessor of X has been visited before X". My analysis has however assumed the order to be topological, without any issues. Do you know if that's just been luck, or if it is somehow guaranteed?

### DIFF
--- a/jlm/llvm/opt/alias-analyses/Andersen.cpp
+++ b/jlm/llvm/opt/alias-analyses/Andersen.cpp
@@ -36,11 +36,16 @@ class Andersen::Statistics final : public util::Statistics
   inline static const char * NumLoadConstraints_ = "#LoadConstraints";
   inline static const char * NumFunctionCallConstraints_ = "#FunctionCallConstraints";
 
+  inline static const char * NumUnificationsOVS_ = "#Unifications(OVS)";
+  inline static const char * NumConstraintsRemovedOfflineNorm_ = "#ConstraintsRemoved(OfflineNorm)";
+
   inline static const char * NumNaiveSolverIterations_ = "#NaiveSolverIterations";
   inline static const char * NumWorklistSolverWorkItems_ = "#WorklistSolverWorkItems";
 
   inline static const char * AnalysisTimer_ = "AnalysisTimer";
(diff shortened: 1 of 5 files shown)


phate (approved):
LGTM

phate:
@haved : The `rvsdg::topdown_traverser` should traverse the nodes in a region in topological order. However, the underlying data structure for keeping track of the respective tracker state of the nodes (behind, frontier, ... ) and ultimately the nodes is a [set](https://github.com/phate/jlm/blob/ec9c7b6035606e40106bb104cc47d75d100a8d85/jlm/rvsdg/tracker.cpp#L156C1-L156C83). My suspicion is now that this could mean that the order of nodes between two traversals of the same region might not be the same for nodes that are independent from each other (due to internal rehashing in the set). It could be that for two independent nodes `n1` and `n2`, you would visit in traversal one `n1` first (and create a constraint), while in traversal two, you would visit `n2` first (and create a constraint). Both traverses are a topological order (as `n1` and `n2` are independent), but would lead to a different order of adding constraints etc. This is the best explanation I can come up why there might be a different traversal order. If you really think that we are traversing wrongly, i.e., not in topological order, then we would need to dig deeper as this would be a huge issue. It still baffles me that for the *same* RVSDG (as in the same in memory), two consecutive traversals would lead to a different traversal orders. @caleridas might have something to say here as well as he implemented the original piece of code.

phate:
I was thinking a little bit more about it. It should be no problem to verify whether the topdown traverser produces a different node order for the two consecutive traversals. The RVSDG is the same for both traversals and should also be the same in memory (as it is not modified). Thus a simple printing of the node pointer for both traversals and then comparing the lists with each other should do the trick, no?

PR #435, opened 2024-04-10
[Andersen] Split HasEscaped into two flags

haved:
The flag `HasEscaped` is now reserved for memory objects only. It implies the `PointeesEscaping` flag. The latter flag can be applied to registers, and it is also shared between all PointerObjects in a unification.
This allows the worklist to properly handle unification.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
+++ b/jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
@@ -193,17 +193,17 @@ PointerObjectSet::GetPointerObjectKind(PointerObjectIndex index) const noexcept
 }
 
 bool
-PointerObjectSet::CanPointerObjectPoint(PointerObjectIndex index) const noexcept
+PointerObjectSet::ShouldTrackPointees(PointerObjectIndex index) const noexcept
 {
   JLM_ASSERT(index < NumPointerObjects());
-  return PointerObjects_[index].CanPoint();
+  return PointerObjects_[index].ShouldTrackPointees();
 }
 
 bool
-PointerObjectSet::CanPointerObjectBePointee(PointerObjectIndex index) const noexcept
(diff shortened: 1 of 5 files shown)


haved:
It's worth noting that the "hack" `CanPoint()` has been replaced by a nicer solution called `ShouldTrackPointees()`. This PR hasn't fully implemented the idea of not tracking pointees using the combination of the PointsToExternal and PointeesEscaping flags, but this only affects performance, and not the final PointsToGraph.

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -721,61 +749,66 @@ size_t
...
     }
   }
 
-  // The worklist, initialized with every object
+  // The worklist, initialized with every unification root
   util::LrfWorklist<PointerObjectIndex> worklist;
   for (PointerObjectIndex i = 0; i < Set_.NumPointerObjects(); i++)
-    worklist.PushWorkItem(i);
+  {
+    if (Set_.GetUnificationRoot(i) == i)
```
phate, line 801: `Set_.IsUnificationRoot(i)` and then you implement this method with exactly the expression from above in its body.

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -788,47 +821,62 @@ PointerObjectConstraintSet::SolveUsingWorklist()
...
+  // - It has been marked as escaping all pointees since last time it was fired
+  // All work items are unification roots, or were unification roots when added
   while (worklist.HasMoreWorkItems())
   {
     const auto n = worklist.PopWorkItem();
     numWorkItems++;
 
+    // Only visit unification roots
+    const auto root = Set_.GetUnificationRoot(n);
+    if (n != root)
```
phate, line 860: `if (!Set_.IsUnificationRoot(n))`

phate (dismissed):

PR #436, opened 2024-04-11
Remove load-store-alloca reduction

### LINKED ISSUE #341: Using -JAASteensgaardAgnostic -JNodeReduction throws type error for BitfieldTest
**Recreate**
make

**Output**
jlc -w -Wno-implicit-int -JAASteensgaardAgnostic -JNodeReduction 2010-05-24-BitfieldTest.c
  what():  Type error - expected : bit8, received : bit64

[2010-05-24-BitfieldTest.zip](https://github.com/phate/jlm/files/13848576/2010-05-24-BitfieldTest.zip)

phate:
This PR removes the load-store-alloca reduction.

The reduction was buggy (it failed when the stored and loaded value was of a different type) and it is also just a more specific version of the [load-store reduction](https://github.com/phate/jlm/blob/86d9af20fef46935ab31ae4fbcb840eed9f50180/jlm/llvm/ir/operators/load.cpp#L205C1-L213C25). Since the load-store reduction is more general, we can just remove this reduction instead of fixing it.

Closes #341

### DIFF
--- a/jlm/llvm/ir/operators/load.cpp
+++ b/jlm/llvm/ir/operators/load.cpp
@@ -100,48 +100,6 @@ is_load_alloca_reducible(const std::vector<rvsdg::output *> & operands)
   return false;
 }
 
-/*
-  a s = alloca_op
-  ss = store_op a ... s
-  v sl = load_op a ss
-  =>
-  v s
-*/
-static bool
-is_load_store_alloca_reducible(const std::vector<rvsdg::output *> & operands)
-{
-  if (operands.size() != 2)
-    return false;
-
-  auto address = operands[0];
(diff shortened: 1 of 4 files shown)


phate:
@haved The node reduction pass performs simple graph rewrites, i.e. peephole optimizations. Both, the load-store reduction and the load-store-alloca reduction, were rerouting the stored value as the new origin of the loaded value (given that certain conditions are satisfied). Neither of the two rewrites were actually removing the alloca node. This needs to be done eventually by dead node elimination, but we are not there yet as we still need to be able to render the memory state as dead (which is currently not happening). We are inching closer though.

PR #437, opened 2024-04-11
Add method for dumping the constraint set as a subset graph

haved:
Also adds some utility functions to the GraphWriter classes, to make it nicer to use. At first I was surprised that `arrowtail` is completely ignored when edges have their default `dir=forward`, so it now selects a `dir` that makes manually specified arrow heads visible.

The subset graphs produced look like
image

Regular edges are superset constraints, dashed edges are complex constraints (dereference the side with the circle to convert it into a regular superset constraint edge).
The + means points to escaped, e means all the pointees in the set should escape. NOTRACK is for PointerObjects where the final PointsToGraph node doesn't get any pointees. Function call constraints don't use edges, so they are a bit harder to read.

For this PR, I have added another temporary environment variable `JLM_ANDERSEN_DUMP_SUBSET_GRAPH`, to enable dumping of the subset graph before and after solving. I don't imagine it will ever make sense to make it a proper command line option, so taking the ENV approach seemed the most sensible.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
+++ b/jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
@@ -764,13 +764,27 @@ PointerObjectConstraintSet::DrawSubsetGraph(jlm::util::GraphWriter & writer) con
     auto & node = graph.CreateNode();
     nodes[i] = &node;
 
-    if (Set_.IsPointerObjectRegister(i))
+    std::ostringstream label;
+    label << i;
+
+    auto kind = Set_.GetPointerObjectKind(i);
+    if (kind == PointerObjectKind::Register)
       node.SetShape(util::Node::Shape::Oval);
     else
       node.SetShape(util::Node::Shape::Rectangle);
 
-    std::ostringstream label;
-    label << i << "\n";
+    if (kind == PointerObjectKind::AllocaMemoryObject)
+      label << " A";
+    else if (kind == PointerObjectKind::MallocMemoryObject)
+      label << " M";
+    else if (kind == PointerObjectKind::FunctionMemoryObject)
+      label << " F";
+    else if (kind == PointerObjectKind::GlobalMemoryObject)
+      label << " G";
+    else if (kind == PointerObjectKind::ImportMemoryObject)
+      label << " I";
+
+    label << "\n";
 
     if (Set_.IsUnificationRoot(i))
     {
@@ -789,7 +803,7 @@ PointerObjectConstraintSet::DrawSubsetGraph(jlm::util::GraphWriter & writer) con
       label << "}";
 
       if (Set_.HasPointeesEscaping(i))
-        label << "\nescaping";
+        label << "e";
     }
     else
     {
@@ -797,7 +811,7 @@ PointerObjectConstraintSet::DrawSubsetGraph(jlm::util::GraphWriter & writer) con
     }
 
     if (!Set_.ShouldTrackPointees(i))
-      label << "\nNOT TRACKING";
+      label << "\nNOTRACK";
 
     node.SetLabel(label.str());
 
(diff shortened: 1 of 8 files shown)


jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -751,6 +751,157 @@ PointerObjectConstraintSet::GetConstraints() const noexcept
...
+      label << " A";
+    else if (kind == PointerObjectKind::MallocMemoryObject)
+      label << " M";
+    else if (kind == PointerObjectKind::FunctionMemoryObject)
+      label << " F";
+    else if (kind == PointerObjectKind::GlobalMemoryObject)
+      label << " G";
+    else if (kind == PointerObjectKind::ImportMemoryObject)
+      label << " I";
+
```
phate, line 786: Could we also add a default case with `JLM_UNREACHABLE()` here?

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -752,6 +752,13 @@ class PointerObjectConstraintSet final
   [[nodiscard]] const std::vector<ConstraintVariant> &
   GetConstraints() const noexcept;
 
+  /**
+   * Creates a subset graph containing all PointerObjects, their current points-to sets,
+   * and edges representing the current set of constraints.
+   */
+  jlm::util::Graph &
+  DrawSubsetGraph(jlm::util::GraphWriter & writer) const;
```
phate, line 760: Could we get a unit test on this method? That is one of the big benefits of the internal dot representation. Let's make use of it.

phate (changes requested):
Looks overall good, but I would still like to have a unit test on the `DrawSubsetGraph()` method as well as some splitting up in individual functions maybe.

tests/jlm/util/TestGraphWriter.cpp
```diff
@@ -32,16 +32,29 @@ TestGraphElement()
...
   graph.SetAttribute("color", "\"brown\"");
+  graph.SetAttribute("colour", "\"british brown\"");
   graph.SetAttributeGraphElement("graph", graph);
   graph.SetAttributeObject("another graph", &myInt);
 
+  assert(graph.HasAttribute("color"));
+  assert(graph.RemoveAttribute("colour"));
+  assert(!graph.HasAttribute("colour"));
+  // Removing the attribute again returns false
+  assert(!graph.RemoveAttribute("colour"));
```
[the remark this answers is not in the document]
haved: No, just me doing different spellings on purpose. I will make it clearer that they are supposed to be different.

haved:
I think all your comments should be addressed now, @phate. I had to add some methods to the GraphWriter, but I agree that being able to programatically inspect the graph is helpful for tests.

phate:
@haved Could it be that you forgot to push your changes? I do not see any changes after my comments except a merge from master.

haved:
@phate You are absolutely correct, It should be fixed now

phate (approved):
Approved as I do not want to delay this PR any longer.

The one thing I noticed is that magic constants are distributed throughout the code. Generally, they should be assigned to a static variable and the variable should then be used. In particular, if they are used at multiple places such as "arrowhead" (GraphWriter.cpp:985 and 1004), "arrowtail" (GraphWriter.cpp:992, 1005). I sadly did not catch this in the first review.

haved:
True, you bring up a good point, there are quite a few string literals, many of which are repeated. I will merge this for now, but I would like to clean some of it up later. I originally felt like adding a GetStyle() method for extracting the "style" attribute could be seen as overkill, but maybe you think it's actually fine @phate?

phate:
@haved I do not see a `GetStyle()` method. Did you mean `SetStyle()`? If so, I do not see a problem with this method. Am I missing something?

We currently hand also all the arguments as strings in. I would rather have enumerations and then internally translate them, but that is another issue.

haved:
@phate No I meant `GetStyle()`, which currently does not exist. The tests do `GetAttribute("style")` to extract the style, but the existence of `SetStyle()` maybe implies there should be a corresponding `GetStyle()`.

PR #441, opened 2024-04-16
CI HLS workflow uses specific jlm-eval-suite commit

sjalander:
The HLS workflow performs a checkout of a specific jlm-eval-suite commit. This makes it possible to make updates to the jlm-eval-suite independently of jlm, and that later can be used by the jlm CI. This avoids breaking the CI until pull requests have been approved for both the jlm-eval-suite and jlm repositories.

### DIFF
--- a/.github/workflows/hls.yml
+++ b/.github/workflows/hls.yml
@@ -28,6 +28,10 @@ jobs:
         install-verilator: true
     - name: "Clone jlm-test-suite"
       run: git clone https://github.com/phate/jlm-eval-suite.git
+    - name: "Checkout specific version of jlm-eval-suite"
+      run: |
+        cd jlm-eval-suite
+        git checkout d91d42359ae7cfb261b76a36da9c532ac65e89a5
     - name: "Link the build output directory to where it is expected by jlm-eval-suite"
       run: ln -s ${{ github.workspace }}/build jlm-eval-suite/jlm/build
     - name: "Run hls-test-suite"


phate:
@sjalander I am generally in favor of this, but I think it should go hand in hand with:

1. Checking out jlm master in the jlm-eval-suite CI run to ensure that every change to the eval suite actually compiles with the newest version of jlm.
2. Getting rid of the jlm submodule in the jlm-eval-suite. This is rather optional, but should happen to avoid confusion with the master checkout in the CI.

What do you think?

sjalander:
"1. Checking out jlm master in the jlm-eval-suite CI run to ensure that every change to the eval suite actually compiles with the newest version of jlm."

I don't see how we avoid breaking the eval CI regardless of how we organize this. The current PR in eval would fail regardless if it uses the jlm submodule or the head of master version since it needs new functionality in jlm, but updating jlm first would mean that its CI would fail due to missing functionality in eval.

Take the current case that initiated this PR, the HLS backend is generating one more file that must be compiled when running the hls-test-suite. The only way I could avoid temporarily breaking one of the CIs would be to add functionality in either eval or jlm that makes it possible to handle the cases with and without this extra file. But this would be additional logic in one of the repos that one then want to get rid of as soon as the transition has been made for both repositories.

So, is the eval CI needed in the first place. The benefit is that we check that things work when we don't make breaking changes (like above), which is relatively rare, and we know when it is expected and could simply ignore the CI run in those cases.


"2. Getting rid of the jlm submodule in the jlm-eval-suite. This is rather optional, but should happen to avoid confusion with the master checkout in the CI."

I see two benefits of using a submodule:
1. jlm has a dedicated location, such that the whole setup works out of the box (this is minor as it could be fixed by having a configuration script or setting an environment variable, e.g., JLM_BIN)
2. If an older version of jlm is needed to be run, then one have a setup that works, e.g., when porting the HLS I could go back and get a version of jlm, circt, and HLS suite that runs together.


One option could be to deprecate jlm-eval-suite and instead only use individual repositories, e.g., spec, polybench, llvm-suite, hls-suite, etc. This is already the case for polybench, where the jlm CI doesn't rely on jlm-eval-suite to run it.
There are a couple of drawbacks with deprecating eval:
1. One would need to check out each repository individually and create one's own setup such that all pieces runs together. This is minor if one use a single computer and set it up once, but I tend to use multiple machines, and it's not that rare for me to grab a new version to verify that things are working or try something new. It also makes the setup for new people (e.g., master students) more cumbersome.
2. Missing version control
3. I like the fact that I know can be in jlm-eval-suite and run everything from there without having to move to a specific directory, e.g., compile jlm, run polybench, spec, llvm-suite, hls-suite, etc.

1. and 2. could be handled by more (build) scripts in jlm like we have for CIRCT and MLIR that use a form of version control.
I have an idea how 3 could be achieved but I leave that for another discussion.

phate (approved):

PR #442, opened 2024-04-17
Add missing tests to Makefiles

phate:
We had test files missing in the Makefiles. I went through all the test listings in the Makefiles and compared to the files in the test folder to ensure that everything was added.


haved (approved):
Nice catch!

PR #443, opened 2024-04-18
Add ascii output format option to jlm-opt

phate:
Adds an option to jlm-opt for printing an RVSDG in ASCII format.


haved (dismissed):
Look good, very nice to have the option.

phate:
@haved I simplified the `ToCommandLineArgument()` method. Could you have a look again, please?

haved (approved):
Look good, sensible use of `unordered_map::at()` to not have to throw our own exceptions.

PR #447, opened 2024-04-21
Add valgrind to install packages composite action

phate:
The valgrind jobs are currently failing in several PRs as fetching the mirrors fails. This is due to a missing `apt-get update` before installing the package. This PR does the following:

1. Adds the valgrind installation to the InstallPackages composite action
2. Ensures that an `apt-get update` is performed before packages are installed

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,10 +32,6 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
-      - name: "Install valgrind"
-        uses: ./.github/actions/InstallPackages
-        with:
-          install-valgrind: true
       - name: "Build jlm"
         uses: ./.github/actions/BuildJlm
         with:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,8 +32,10 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
-      - name: Install valgrind
-        run: sudo apt-get install valgrind
+      - name: "Install valgrind"
+        uses: ./.github/actions/InstallPackages
+        with:
+          install-valgrind: true
       - name: "Build jlm"
         uses: ./.github/actions/BuildJlm
         with:
--- a/.github/actions/BuildJlm/action.yml
+++ b/.github/actions/BuildJlm/action.yml
@@ -42,11 +42,12 @@ runs:
         exit 1
       shell: bash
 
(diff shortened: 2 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -32,8 +32,10 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
-      - name: Install valgrind
-        run: sudo apt-get install valgrind
+      - name: "Install valgrind"
```
sjalander, line 35: The other tests simply call the BuildJlm action, which in turn installs the necessary packages.
So why not remove this and instead use the valgrind-check in the BuildJlm action to pass 'install-valgrind' on to the InstallPackages actions from the BuildJlm action.

I'm guessing that the following could be changed

    - name: "Install LLVM dependencies"
      uses: ./.github/actions/InstallPackages
      with:
        install-llvm: true
        install-clang: true

to

    - name: "Install dependencies"
      uses: ./.github/actions/InstallPackages
      with:
        install-llvm: true
        install-clang: true
        install-valgrind: valgrind-check

haved: I agree that calling into `./.github/actions/InstallPackages` from both the top level workflow, and from within the `BuildJlm` action is a bit annoying. It is however nice to not have to mention valgrind from within `BuildJlm`, since valgrind doesn't have anything to do with building.

Currently, the `BuildCirct` and `BuildMlirDialect` actions both call into `InstallPackages`, in addition to `BuildJlm` itself, bringing us up to at least three invocations (four if you include the valgrind install) in a single job. Looking at the log timestamps, `sudo apt-get update` takes about 2 seconds, so we "waste" 6 seconds of extra update calls. Trying to install a package that is already installed takes about 1 second for `apt-get` to realize and tell us, which happens about 5 times in the job. In total I would say 10 seconds are lost due to duplicate work in InstallPackages.

phate: I am not worried about the extra seconds. They are not on the critical path, nothing is lost.

I will move the installation in to the BuildJlm script. This is the script that executes the valgrind-check target, which is dependent on valgrind. So it also makes sense that the dependency is installed there.

haved: Ah true, I had forgotten that valgrind is a separate make target that effectively replaces building

phate: I pushed a fix. We are installing valgrind unconditionally now when we are building jlm. Just to be done with it.

sjalander (approved):

Issue #446, opened 2024-04-20
Handling of volatile memory accesses


phate:
We currently do not support `volatile` memory accesses, which was not a problem until the introduction of memory state encoding. The problem originates from the loss of the original ordering of volatile memory accesses. The following example illustrates the problem:

```c
extern void exit (int);                                                                                
extern void abort (void);                                                                              
                                                                                                       
volatile int a = 1;                                                                                    
volatile int b = 0;                                                                                    
volatile int x = 2;                                                                                    
volatile signed int r = 8;                                                                             
                                                                                                       
void __attribute__((noinline))                                                                         
foo (void)                                                                                             
{                                                                                                      
  exit (0);                                                                                            
}                                                                                                      
                                                                                                       
int                                                                                                    
main (void)                                                                                            
{                                                                                                      
  int si1 = a;                                                                                         
  int si2 = b;                                                                                         
  int i;                                                                                               
                                                                                                       
  for (i = 0; i < 100; ++i) {                                                                          
      foo ();                                                                                          
      if (x == 8)                                                                                      
          i++;                                                                                                 
      r += i + si1 % si2;                                                                              
  }                                                                                                    
  abort ();                                                                                            
}   
```

This is test program `pr38819.c` from the LLVM test suite. The unoptimized LLVM IR is as follows:

```
; ModuleID = 'pr38819.c'                                                                               
source_filename = "pr38819.c"                                                                          
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"           
target triple = "x86_64-pc-linux-gnu"                                                                  
                                                                                                       
@a = dso_local global i32 1, align 4                                                                   
@b = dso_local global i32 0, align 4                                                                   
@x = dso_local global i32 2, align 4                                                                   
@r = dso_local global i32 8, align 4                                                                   
                                                                                                       
; Function Attrs: noinline nounwind optnone uwtable                                                    
define dso_local void @foo() #0 {                                                                      
  call void @exit(i32 noundef 0) #2                                                                    
  unreachable                                                                                          
}                                                                                                      
                                                                                                       
; Function Attrs: noreturn                                                                             
declare void @exit(i32 noundef) #1                                                                     
                                                                                                       
; Function Attrs: noinline nounwind optnone uwtable                                                    
define dso_local i32 @main() #0 {                                                                      
  %1 = alloca i32, align 4                                                                             
  %2 = alloca i32, align 4                                                                             
[...]

PR #448, opened 2024-04-21
Add support for volatile memory accesses

phate:
This PR adds support for volatile memory accesses. Concretely, it does the following:

1. Add volatile support to load operations.
2. Add volatile support to store operations.
3. Adjust load reductions to handle volatile.
4. Adjust store reductions to handle volatile.

We do not need to do anything for the memcpy operation as it already supported volatile.

### DIFF
--- a/jlm/llvm/ir/operators/store.cpp
+++ b/jlm/llvm/ir/operators/store.cpp
@@ -41,6 +41,13 @@ StoreNode::copy(rvsdg::region * region, const std::vector<rvsdg::output *> & ope
 
 /* store normal form */
 
+// so3 = memoryStateMerge si1 si2
+// so4 = store a v so3
+// ... = anyOp so4
+// =>
+// so1 so2 = store a v si1 si2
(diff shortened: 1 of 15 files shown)


jlm/llvm/ir/operators/store.cpp
```diff
@@ -107,9 +137,18 @@ is_store_alloca_reducible(const std::vector<jlm::rvsdg::output *> & operands)
   if (alloca->output(1)->nusers() != 1)
     return false;
 
+  // It should NOT happen that a store node with an address directly from an alloca node is
```
haved, line 140: Why not? The following is legal c, however a bit strange for sure: 
```c
int main() {
    volatile int a;
    a = 10;
    return a;
}
```

haved (commented):
All in all this looks good, but there are some comments in the code I'm not entirely sure about.

phate:
I will close this PR. It is inadequate to address the volatile issue as explained in #446

PR #450, opened 2024-04-21
[Andersen] Add Offline Variable Substitution

haved:
Adds OVS, using an offline version of the subset graph, finding SCCs using `util/TarjanSCC`, assigning equivalence set labels (which also allows "direct" nodes to re-use their predecessors' label), and unifying them.

By doing this, I also found a bug in `PointerObjectSet::UnifyPointerObjects` code, which is fixed.

OVS only does unification, but afterwards you can call `PointerObjectConstraintSet::NormalizeConstraints()` to redirect constraints to unification roots directly, and remove duplicates. The worklist algorithm kind of does this already, but it's nice to get a statistic about how many constraints can be removed by doing OVS.

The default for Andersen is now OVS+Normalization+Worklist solver.

The Andersen tests had to be changed, as enabling OVS causes the produced `PointsToGraph` to have fewer `RegisterNode`s, since more registers are represented by the same node.

Tests-wise, only `util/TarjanSCC.hpp` has a test added currently. I will write some tests for the OVS as well.

**However:** The entire `llvm-test-suite` subset we use, gives PointsToGraphs isomorphic to the Naive solver for every single source file!


jlm/util/TarjanSCC.hpp
```diff
@@ -0,0 +1,133 @@
...
+ * @tparam SuccessorFunctor a functor with the signature (size_t) -> iterable<size_t>
+ * @param numNodes the number of nodes. Nodes are indexed from 0 to numNodes-1
+ * @param successors an instance of the SucessorFunctor
+ * @param sccIndex output vector to be filled with the index of the SCC each node ends up in
+ * @param topologicalOrder output vector to be filled with the nodes in a weak topological order
+ * @return the number of SCCs in the graph. One more than the largest SCC index
+ */
+template<typename SuccessorFunctor>
+size_t
+FindStronglyConnectedComponents(
```
phate, line 38: I am quite excited about this, but I am worried whether this method is not general enough to replace other implementations:
1. https://github.com/phate/jlm/blob/f9cc3456d60c9df68b0f55c66cfb9cdf560eef2d/jlm/llvm/ir/ipgraph.cpp#L12C1-L52C2
2. https://github.com/phate/jlm/blob/f9cc3456d60c9df68b0f55c66cfb9cdf560eef2d/jlm/llvm/ir/cfg-structure.cpp#L82C1-L128C2

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -958,6 +972,273 @@ PointerObjectConstraintSet::DrawSubsetGraph(util::GraphWriter & writer) const
...
+  // Output vectors from Tarjan's SCC algorithm
+  std::vector<size_t> sccIndex;
+  std::vector<size_t> topologicalOrder;
+  auto numSccs = util::FindStronglyConnectedComponents(
+      totalNodeCount,
+      GetSuccessors,
+      sccIndex,
+      topologicalOrder);
+
+  // Find out which SCCs only contain direct nodes
```
phate, line 1138: What are direct nodes?

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -958,6 +972,273 @@ PointerObjectConstraintSet::DrawSubsetGraph(util::GraphWriter & writer) const
...
+      unificationRoot[equivalenceSetLabel] = i;
+  }
+
+  return numUnifications;
+}
+
+size_t
+PointerObjectConstraintSet::NormalizeConstraints()
+{
+  // The new list of constraint, preserving the order of constraints that are not deleted.
```
phate, line 1181: "The new list of constraint**s**, ..."

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -958,6 +972,273 @@ PointerObjectConstraintSet::DrawSubsetGraph(util::GraphWriter & writer) const
...
+  std::vector<ConstraintVariant> newConstraints;
+
+  // Sets used to avoid adding duplicates of any constraints
+  std::set<std::pair<PointerObjectIndex, PointerObjectIndex>> addedSupersetConstraints;
+  std::set<std::pair<PointerObjectIndex, PointerObjectIndex>> addedStoreConstraints;
+  std::set<std::pair<PointerObjectIndex, PointerObjectIndex>> addedLoadConstraints;
+  std::set<std::pair<PointerObjectIndex, const CallNode *>> addedCallConstraints;
```
phate, line 1188: Why do you use `std::set` instead of `std::unordered_set` or `HashSet`?

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -716,32 +720,42 @@ EscapedFunctionConstraint::PropagateEscapedFunctionsDirectly(PointerObjectSet &
   return modified;
 }
 
+bool
+PointerObjectConstraintSet::IsFrozen()
+{
+  return ConstraintSetFrozen_;
+}
+
```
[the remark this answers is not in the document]
phate: Ignore the comment above. The documentation is on the declaration (obviously).

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -759,6 +769,41 @@ class PointerObjectConstraintSet final
...
+   * the same pointees, and unifies them.
+   * It is a version of Rountev and Chandra, 2000: "Off-line variable substitution",
+   * modified to support SSA based constraint graphs.
+   *
+   * The algorithm uses an offline constraint graph, consisting of two nodes per PointerObject v:
+   *  n(v) represents the points-to set of v
+   *  n(*v) represents the union of points-to sets of all pointees of v
+   *  Edges in the graph represent points-to set inclusion.
+   *
+   * SCCs are collapsed into single equivalence sets. n(*v) nodes are not included.
```
phate, line 783: Maybe spell it SCCs out here for the first time as well.

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -759,6 +769,41 @@ class PointerObjectConstraintSet final
...
+   *
+   * If an SCC consists of only "direct" nodes, and all predecessors share equivalence
```
phate, line 785: Again, what are direct nodes?

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -787,11 +832,33 @@ class PointerObjectConstraintSet final
...
 private:
+  /**
+   * Creates the special subset graph containing both regular nodes and dereference nodes,
+   * used by DoOfflineVariableSubstitution().
+   *
+   * Some nodes are marked as direct, when all subset predecessors are known offline.
+   * They can:
+   *  - only be n(v) nodes, where v is a register
+   *  - not be the return value of a function call
+   *  - not be an argument of a function body
```
phate, line 843: Right, here is the definition of it. Maybe add some breadcrumbs in the other comments?

tests/jlm/util/TestTarjanSCC.cpp
```diff
@@ -0,0 +1,204 @@
...
+TestTarjanSCC()
+{
+  TestDAG();
+  TestCycles();
+  TestDiamondChain(false);
+  TestDiamondChain(true);
+  return 0;
+}
+
+JLM_UNIT_TEST_REGISTER("jlm/util/TestTarjanSCC", TestTarjanSCC)
```
phate, line 204: I know we have (mostly) not done it so far, but I would like to move away from the single test function per file and move to multiple tests per file such that we have the chance to invoke them individually. In other words, it should look like this:

```
static int
TestDag()
{
...
}
JLM_UNIT_TEST_REGISTER("jlm/util/TestTarjanSCC-TestDag", TestDAG)

static int
TestCycles()
{
...
}
JLM_UNIT_TEST_REGISTER("jlm/util/TestTarjanSCC-TestCycles", TestCycles)
...
```

phate (changes requested):
Very impressive progress!

I have mostly commented on minor things(naming, clarifications, etc.) as I am honestly a little bit lost on the details due to me not having studied the papers in detail. However, the code is all very readable and I can follow the big picture even though the details elude me.

I am very excited about the Tarjan implementation and would like to see it being used in other places as well (instead of the specific implementations). Do you think it is possible to generalize it? If possible, maybe we should merge the Tarjan implementation in its own PR?

haved:
Hopefully the code makes a lot more sense when you also read the corresponding section of the thesis. Regarding using the SCC, I originally tried to make it not general using more templates (e.g. not requiring nodes to be indexed from 0 to N-1, but be general values insteas) but it became hard to support both vector and unordered_map for the output types. If you want to use the code, you can of course create a mapping externally.

I have, however, generalized it enough to be usable in at least one more situation: Pearce07 replaces the worklist by instead doing SCC followed by a topological order sweep of the DAG. This can easily be tested once online unification is known to work through lazy and hybrid cycle detection.

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
haved: Good point. Originally I tried to make all the Constraint variants, e.g. `StoreConstraint`, implement `operator <`, which would allow me to put them directly into a set. Implementing the comparison operator feels nicer than overloading `std::hash`. I ended up not liking it that much and went over to using `std::pair`, which does overload `std::hash`, so I will switch to `unordered_set`, i.e. `util::HashSet`.

jlm/util/TarjanSCC.hpp
haved: I agree it could be even more general, by not using a continuous range of indices. I will try to make a light wrapper and see how it ends up.

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
haved: Actually, I did try using `HashSet`. It turns out `std::pair` doesn't have an overload for `std::hash`, which makes a lot of people very angry and is widely regarded as a bad move. I will create a separate PR with a workaround, if we think it is worth it.

haved:
@phate I looked into generalizing Tarjan more for the cases you suggested. The main issue was that those discover nodes while working, while my code needs to have the set of nodes ready at the start, the way it's currently written. 

Instead of modifying it, I would like to wait until I have the scaffolding in place for running Andersen on SPEC source files, and timing the processing steps. That's what I'm working on now, and it would make it easier to check if generalizations have any negative performance impact.

I think I fixed all other comments you had, and made some comments clearer. I also moved to individual test registration for Tarjan, and strengthened the tests a bit.

This PR should be merged after #452, if you accept it.

phate:
Let's wait with the generalization then. I have not problem with that. Your work has priority!

phate (approved):

PR #452, opened 2024-04-24
Add custom Hash functor for HashSet, with overload for std::pair

haved:
@phate's comment about using `std::set` instead of `util::HashSet` made me remember why I picked `set` in the first place: `std::pair` doesn't have a specialization of `std::hash`. Since we have our own `HashSet` class that we like to use, I thought it would be fun to see what it would look like to add support for `std::pair`without requiring the user to specify a custom hash function for it.

I also added some constructors to be able to initialize the set with sets that do not have the same hash functor.

If we prefer to not go this route that is of course completely OK. I do enjoy being able to place `std::pair` in HashSets without any extra hassle, though.

### DIFF
--- a/tests/jlm/util/TestHashSet.cpp
+++ b/tests/jlm/util/TestHashSet.cpp
@@ -63,6 +63,30 @@ TestUniquePointer()
 }
 
 static void
+TestPair()
+{
+  jlm::util::HashSet<std::pair<int, int>> hashSet{ { 1, 10 }, { 5, 50 } };
+
+  // Inserting new value
+  auto result = hashSet.Insert({ 7, 70 });
+  assert(result && hashSet.Size() == 3);
+
+  // Try inserting already inserted
+  result = hashSet.Insert({ 1, 10 });
+  assert(!result && hashSet.Size() == 3);
+
+  // Contains is only true in the correct order
+  assert(hashSet.Contains({ 5, 50 }));
+  assert(!hashSet.Contains({ 50, 5 }));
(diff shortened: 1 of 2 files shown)


tests/jlm/util/TestHashSet.cpp
```diff
@@ -62,6 +62,30 @@ TestUniquePointer()
...
+  assert(hashSet.Contains({ 5, 50 }));
+  assert(!hashSet.Contains({ 50, 5 }));
+
+  // Removing works
+  result = hashSet.Remove({ 5, 50 });
+  assert(result && hashSet.Size() == 2);
+  result = hashSet.Remove({ 5, 50 });
+  assert(!result && hashSet.Size() == 2);
+}
+
```
phate, line 88: Could you make this its own unit test again, i.e., put a `JLM_REGISTER_UNIT_TEST()` for it instead of adding it in the `Test()` method?

phate (approved):
LGTM

tests/jlm/util/TestHashSet.cpp
haved: Sure, I assume it's fine to do it for all then

phate: It absolutely is. Thanks a lot!

haved:
I moved each test to be registered individually. Clang format likes to put an extra blank line between the test function and macro for registering

phate (approved):
LGTM

PR #453, opened 2024-04-25
Remove loop state

phate:
The loop state was a quick idea that turned out to be a dead end. By now it also became a burden. Let's remove it.


haved (approved):
A herculean effort, everything looks sensible, though you are of course modifying quite a few files I've never looked at before. Nice cleanups of `jlm::` and multiline comments.

PR #457, opened 2024-04-29
Add support for volatile stores

phate:
This PR does the following:

    Add StoreOperation and StoreNode class for volatile load operations.
    Adds conversion support for LLVM -> RVSDG
    Adds conversion support for RVSDG -> LLVM

### DIFF
--- a/jlm/llvm/backend/jlm2llvm/instruction.cpp
+++ b/jlm/llvm/backend/jlm2llvm/instruction.cpp
@@ -324,18 +324,45 @@ convert(
       ctx);
 }
 
-static inline ::llvm::Value *
-convert_store(
-    const rvsdg::simple_op & op,
-    const std::vector<const variable *> & args,
+static void
+CreateStoreInstruction(
+    const variable * address,
+    const variable * value,
+    bool isVolatile,
+    size_t alignment,
     ::llvm::IRBuilder<> & builder,
     context & ctx)
 {
-  JLM_ASSERT(is<StoreOperation>(op) && args.size() >= 2);
-  auto store = static_cast<const StoreOperation *>(&op);
+  auto storeInstruction = builder.CreateStore(ctx.value(value), ctx.value(address), isVolatile);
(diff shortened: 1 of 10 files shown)


jlm/llvm/backend/jlm2llvm/instruction.cpp
```diff
@@ -324,18 +324,45 @@ convert(
...
+      storeOperation->GetAlignment(),
+      builder,
+      ctx);
+  return nullptr;
+}
 
-  auto i = builder.CreateStore(ctx.value(args[1]), ctx.value(args[0]));
-  i->setAlignment(::llvm::Align(store->GetAlignment()));
+static ::llvm::Value *
+convert(
```
haved, line 359: I assume there is a historic reason why the above function is called `convert_store` and takes an untyped `simple_op`, while this one doesn't?

jlm/llvm/ir/operators/StoreVolatile.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+public:
+  [[nodiscard]] const StoreVolatileOperation &
+  GetOperation() const noexcept
+  {
+    return *util::AssertedCast<const StoreVolatileOperation>(&operation());
+  }
+
+  [[nodiscard]] size_t
+  NumMemoryStates() const noexcept
```
haved, line 134: Are there no helper functions for `GetMemoryStateInput(index)` and `GetMemoryStateOutput(index)`?

jlm/llvm/backend/jlm2llvm/instruction.cpp
phate: Yes, some cleanup and refactoring is necessary in this pass. However, that is for another PR.

jlm/llvm/ir/operators/StoreVolatile.hpp
phate: Meh, I forgot them. It was also not needed when writing the tests. I will add them once they are required. I want to get this merged and move on.

PR #458, opened 2024-05-01
Add support for volatile memcpy

phate:
This PR does the following:
1. Move MemCpy operation class into its own file.
2. Rename MemCpy class to MemCpyOperation to be aligned with the other classes.
3. Adds a MemCpyVolatileOperation class
4. Adds conversion support for LLVM -> RVSDG
5. Adds conversion support for RVSDG -> LLVM


jlm/llvm/ir/operators/MemCpy.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+namespace jlm::llvm
+{
+
+MemCpyOperation::~MemCpyOperation() = default;
+
+bool
+MemCpyOperation::operator==(const operation & other) const noexcept
+{
+  // Avoid common node elimination for memcpy operator
```
haved, line 16: Why is this important?

jlm/llvm/ir/operators/MemCpy.cpp
```diff
@@ -0,0 +1,53 @@
...
+{
+  return std::unique_ptr<rvsdg::operation>(new MemCpyOperation(*this));
+}
+
+MemCpyVolatileOperation::~MemCpyVolatileOperation() noexcept = default;
+
+bool
+MemCpyVolatileOperation::operator==(const operation & other) const noexcept
+{
+  // Avoid common node elimination for memcpy operator
```
haved, line 37: Avoiding common node elimination makes sense for volatile, so I get it here.

jlm/llvm/ir/operators/MemCpy.hpp
```diff
@@ -0,0 +1,188 @@
...
+      const std::vector<const variable *> & memoryStates)
+  {
+    auto operandPorts = CheckAndCreateOperandPorts(length->type(), memoryStates.size());
+    auto resultPorts = CreateResultPorts(memoryStates.size());
+
+    std::vector<const variable *> operands = { destination, source, length };
+    operands.insert(operands.end(), memoryStates.begin(), memoryStates.end());
+
+    MemCpyOperation op(operandPorts, resultPorts);
+    return tac::create(op, operands);
```
haved, line 56: Would we save a copy by constructing the `MemCpyOperation` directly into the`tac::create()`? I never quite know when the C++ compiler elides calling copy constructors.

jlm/llvm/ir/operators/MemCpy.hpp
```diff
@@ -0,0 +1,188 @@
...
+  operator==(const operation & other) const noexcept override;
+
+  [[nodiscard]] std::string
+  debug_string() const override;
+
+  [[nodiscard]] std::unique_ptr<rvsdg::operation>
+  copy() const override;
+
+  static std::unique_ptr<llvm::tac>
+  CreateThreeAddressCode(
```
haved, line 132: I assume there will be a future cleanup changing the names of the corresponding methods in `MemCpyOperation`. You have probably already confirmed this when I've asked about it previously.

haved (approved):
Looks good,  as usual I only really have some questions

jlm/llvm/ir/operators/MemCpy.hpp
phate: Yes, we could save a copy here. We create the operation and then it is copied in the tac constructor:

```
tac::tac(const jlm::rvsdg::simple_op & operation, const std::vector<const variable *> & operands)
    : operands_(operands),
      operation_(operation.copy())
{
  check_operands(operation, operands);

  auto names = create_names(operation.nresults());
  create_results(operation, names);
}
```

However, I do not consider this change as part of this PR.

jlm/llvm/ir/operators/MemCpy.hpp
phate: A future cleanup will collect all the creator methods for three address codes and RVSDG nodes in a builder, respectively. This will delegate the responsible to a single entity instead of spreading it out all over as well as save some code (as checks and port creation etc. can be unified for all operations instead of each of them implementing their own).

jlm/llvm/frontend/LlvmInstructionConversion.cpp
```diff
@@ -737,18 +737,45 @@ convert_free_call(const ::llvm::CallInst * i, tacsvector_t & tacs, context & ctx
   return nullptr;
 }
 
+static bool
+IsVolatile(const ::llvm::Value & value)
```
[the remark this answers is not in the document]
phate: I added some documentation.

jlm/llvm/ir/operators/MemCpy.cpp
phate: It was (not) there already. I implemented now the equality operator.

phate:
@haved I believe to have addressed all your comments.

haved (approved):

PR #460, opened 2024-05-01
Rename LoadOperation and LoadNode classes

phate:
Renames the LoadOperation and LoadNode class to LoadNonVolatileOperation and LoadNonVolatileNode, respectively.

The reason for this renaming is that a common operation and node class will be introduced in order to:

1. Refactor out common properties between the volatile and non-volatile operations/nodes.
2. Provide a common base class for passes, such as alias analysis, that do not care about the volatile/non-volatile.


jlm/llvm/ir/operators/load.cpp
```diff
@@ -11,31 +11,32 @@
...
   return operation && operation->narguments() == narguments()
       && operation->GetLoadedType() == GetLoadedType()
       && operation->GetAlignment() == GetAlignment();
 }
 
 std::string
-LoadOperation::debug_string() const
+LoadNonVolatileOperation::debug_string() const
 {
   return "LOAD";
```
haved, line 28: I guess the debug string can remain as "LOAD", but would you want to change the file name of this file as a part of this PR?

haved (dismissed):
Looks good, only one question

jlm/llvm/ir/operators/load.cpp
phate: @haved I changed the debug string to "Load". Moreover, I changed the file names to `LoadNonVolatile.[ch]pp` as well as the test file name to `LoadNonVolatileTests.cpp`. Please have a look again.

haved (approved):

PR #461, opened 2024-05-01
Enable load-load-state reduction again

### LINKED ISSUE #307: nf->set_load_load_state_reducible(true); in -Jred throws a type_error when -Jred is applied twice
**Recreate**
Modify jlm/llvm/opt/reduction.cpp 

static void
{

enable_load_reductions(graph);
// enable_gamma_reductions(graph);
// enable_unary_reductions(graph);
// enable_binary_reductions(graph);

Then comment out all optimizations except for 
  nf->set_load_load_state_reducible(true);
in enable_gamma_reductions.

Then compile jlm with debug:
make jlm-debug -j nproc

**Output**
  what():  Type error - expected : float, received : bit32
Aborted (core dumped)

[uint64_to_float.zip](https://github.com/phate/jlm/files/13577582/uint64_to_float.zip)

phate:
I could not reproduce the problem from #307 any longer. Looking at the input program and the reported error, I suspect that this optimization was just the trigger, but not the actual cause. I believe that the cause was the missing loaded type in the equality operator of the LoadOperation. It was fixed in #456.

Closes #307

### DIFF
--- a/jlm/llvm/opt/reduction.cpp
+++ b/jlm/llvm/opt/reduction.cpp
@@ -76,9 +76,7 @@ enable_load_reductions(jlm::rvsdg::graph & graph)
   nf->set_load_alloca_reducible(true);
   nf->set_multiple_origin_reducible(true);
   nf->set_load_store_state_reducible(true);
-  // set_load_load_state_reducible throws a type_error
-  // github issue #307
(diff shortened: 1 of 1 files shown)


haved (approved):
CI is happy, and I'm always happy to see more potential memory optimizations

PR #462, opened 2024-05-01
New file for the HLS backend

sjalander:


sjalander:
@phate Was it like this that you talked about, or did you mean one PR for each new file?

phate:
@sjalander Yes, this is what I was talking about. However, maybe you could also add the respective entries in the Makefile so they at least get compiled. Moreover, the respective tests for them in case there are any.

sjalander:
@phate I've added the files to the Makefile such that they compile

sjalander:
@phate And now I've managed to sort out the HLS benchmarks so the CI doesn't fall and merged the latest changes related to renamed loads.

phate:
@sjalander Could we exclude the pass that gives valgrind a headache as well as the unit test for it?

phate:
@sjalander I merged the store file reordering first. This screwed it up. It should be a matter of simply changing header files (I hope) to get it to build again.

sjalander:
@phate Renamed

phate (approved):

PR #463, opened 2024-05-03
Enable Control Constant Reduction again

### LINKED ISSUE #303: enable_gamma_reductions(graph) in -Jred creates PHI node with different input types
**Recreate**
Modify jlm/llvm/opt/reduction.cpp (line 134)

static void
{

// enable_load_reductions(graph);
enable_gamma_reductions(graph);
// nable_unary_reductions(graph);
// nable_binary_reductions(graph);

Then compile jlm with debug:
make jlm-debug -j nproc

**Output**
jlc: /usr/lib/llvm-16/include/llvm/IR/Instructions.h:2800: void llvm::PHINode::setIncomingValue(unsigned int, Value *): Assertion `getType() == V->getType() && "All operands to PHI node must be the same type as the PHI node!"' failed.
Aborted (core dumped)

[20111208-1.zip](https://github.com/phate/jlm/files/13562243/20111208-1.zip)

phate:
This PR enables control constant reduction again. It does the following:

1. Improves debug output of the match and control constant operations
2. Handles the case were all the constants in every branch are the same specifically

The problem arises because match operations are either mapped to integer types of bit width 1 or 32 in the back-end, depending on the nature of the match operation. This all works fine as long as the match operation is directly consumed by an LLVM terminator instruction. If that is not the case any longer, then we might end up with the problem where we have input values of differing type to an operation (i32 vs. i1), such as a SSA phi. The control constant reduction is an optimization that can trigger this problem.

This PR does not resolve the underlying problem (mismatching type translation), but only provides a band-aid such that the problem is less likely to reappear. I currently have no idea how to satisfactorily resolve the underlying problem.

Closes #303

### DIFF
--- a/jlm/llvm/opt/reduction.cpp
+++ b/jlm/llvm/opt/reduction.cpp
@@ -85,9 +85,7 @@ enable_gamma_reductions(jlm::rvsdg::graph & graph)
   auto nf = jlm::rvsdg::gamma_op::normal_form(&graph);
   nf->set_mutable(true);
   nf->set_predicate_reduction(true);
-  // set_control_constante_reduction cause a PHI node input type error
-  // github issue #303
-  nf->set_control_constant_reduction(false);
+  nf->set_control_constant_reduction(true);
 }
 
 static void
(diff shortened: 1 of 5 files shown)


phate:
Okay, this clearly was not enough. The problem is still triggered. *sigh* Back to the drawing board...

PR #465, opened 2024-05-04
Rename StoreOperation and StoreNode class

phate:
Renames the StoreOperation and StoreNode class to StoreNonVolatileOperation
and StoreNonVolatileNode, respectively.

The reason for this renaming is that a common operation and node class
will be introduced in order to:

1. Refactor out common properties between the volatile and non-volatile
operations/nodes.
2. Provide a common base class for passes, such as alias analysis, that
do not care about the volatile/non-volatile.


jlm/llvm/ir/operators/StoreNonVolatile.cpp
```diff
@@ -5,36 +5,37 @@
...
   return operation && operation->narguments() == narguments()
       && operation->GetStoredType() == GetStoredType()
       && operation->GetAlignment() == GetAlignment();
 }
 
 std::string
-StoreOperation::debug_string() const
+StoreNonVolatileOperation::debug_string() const
 {
   return "STORE";
```
haved, line 27: Change this to "Store" like you did with "Load"?

haved (approved):
Looks good, just one comment

jlm/llvm/ir/operators/StoreNonVolatile.cpp
phate: Fixed. Good catch!

haved (approved):

PR #467, opened 2024-05-04
Introduce abstract MemCpyOperation class

phate:


haved (approved):
Nice to have a base class, looks good

PR #470, opened 2024-05-07
Introduce abstract base classes for store operations and nodes

phate:
The PR does the following:
1. Introduces an abstract base class for store operations
2. Introduces an abstract base class for store nodes
3. Factors out common functionality for the store classes

### DIFF
--- a/jlm/llvm/ir/operators/Store.cpp
+++ b/jlm/llvm/ir/operators/Store.cpp
@@ -33,6 +33,12 @@ StoreNonVolatileOperation::copy() const
   return std::unique_ptr<jlm::rvsdg::operation>(new StoreNonVolatileOperation(*this));
 }
 
+[[nodiscard]] size_t
+StoreNonVolatileOperation::NumMemoryStates() const noexcept
+{
+  return nresults();
+}
+
 rvsdg::node *
 StoreNonVolatileNode::copy(rvsdg::region * region, const std::vector<rvsdg::output *> & operands)
     const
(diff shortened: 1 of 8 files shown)


jlm/llvm/ir/operators/Store.cpp
```diff
@@ -63,6 +97,41 @@ StoreVolatileOperation::copy() const
...
+
+  return { MemoryStateInputIterator(input(3)), MemoryStateInputIterator(nullptr) };
+}
+
+[[nodiscard]] StoreNode::MemoryStateOutputRange
+StoreVolatileNode::MemoryStateOutputs() const noexcept
+{
+  if (NumMemoryStates())
+  {
+    return { MemoryStateOutputIterator(nullptr), MemoryStateOutputIterator(nullptr) };
```
haved, line 129: This is backwards, no? If there are memory states, the iterator pair should not start with nullptr?

jlm/llvm/ir/operators/Store.cpp
```diff
@@ -33,6 +33,40 @@ StoreNonVolatileOperation::copy() const
...
+
+  return { MemoryStateInputIterator(input(2)), MemoryStateInputIterator(nullptr) };
+}
+
+[[nodiscard]] StoreNode::MemoryStateOutputRange
+StoreNonVolatileNode::MemoryStateOutputs() const noexcept
+{
+  if (NumMemoryStates())
+  {
+    return { MemoryStateOutputIterator(nullptr), MemoryStateOutputIterator(nullptr) };
```
haved, line 64: This is backwards, no? If there are memory states, the iterator pair should not start with nullptr?

haved (changes requested):
I think there might be to if statements with inverted conditions

jlm/llvm/ir/operators/Store.cpp
phate: Really great catch!

jlm/llvm/ir/operators/Store.cpp
phate: Well spotted!

phate:
@haved I need another approval. I had a merge conflict.

haved (approved):
Looks good!

PR #472, opened 2024-05-08
Fix memory state operation equality

phate:
The equality operator of all memory state operations was not taking the number of operands/results into account. This PR fixes this issue. An upcoming PR will move them in the same file and overhaul their implementation.


haved (commented):
Looks good!

haved (approved):

PR #474, opened 2024-05-08
Adds HLS files that previously were missed

sjalander:


jlm/hls/Makefile.sub
```diff
@@ -47,15 +49,18 @@ libhls_HEADERS = \
...
 	jlm/hls/backend/rvsdg2rhls/GammaConversion.hpp \
 	jlm/hls/backend/rvsdg2rhls/mem-queue.hpp \
 	jlm/hls/backend/rvsdg2rhls/mem-sep.hpp \
 	jlm/hls/backend/rvsdg2rhls/memstate-conv.hpp \
 	jlm/hls/backend/rvsdg2rhls/remove-redundant-buf.hpp \
 	jlm/hls/backend/rvsdg2rhls/remove-unused-state.hpp \
+	jlm/hls/backend/rvsdg2rhls/rhls-dne.hpp \
 	jlm/hls/backend/rvsdg2rhls/rvsdg2rhls.hpp \
 	jlm/hls/backend/rvsdg2rhls/ThetaConversion.hpp \
+	jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.hpp \
```
phate, line 63: This file does not exist in the PR.

phate: Okay, this file already exists in the repository, but is just not mentioned here.

phate (changes requested):
Looking at the other HLS PR, I would have expected the following files in this PR:

1. dae-conv.[ch]pp
2. instrument-ref.[ch]pp
3. mem-conv.cpp. There exists no mem-conv.hpp in the other PR. However, the mem-conv.hpp is mentioned in the Makefile there.
4. rhls-dne.[ch]pp

Number 1 and 4 are here. Number 2 is missing. Any reason? Number 3 is also missing, as well as the hpp file from the other PR.

jlm/hls/Makefile.sub
phate: Ignore the above comments.

sjalander:
They require modifications to jlm/hls/ir/hls.[hc]pp
So this PR only contains files that does not require modifications of already committed files.

phate (approved):

PR #475, opened 2024-05-09
Overhaul MemoryStateEncoder.[ch]pp

phate:
This PR overhauls the Memory state encoder in anticipation of the upcoming changes necessary for volatile memory accesses.


haved (approved):
Nice cleanup!

PR #478, opened 2024-05-10
HLS: Special fork for constants

sjalander:
Adds support for forks that have a constant as input, provides unit tests, and documentation.


jlm/hls/ir/hls.hpp
```diff
@@ -101,13 +106,15 @@ class fork_op final : public jlm::rvsdg::simple_op
...
+  create(size_t nalternatives, jlm::rvsdg::output & value, bool constant = false)
   {
 
     auto region = value.region();
-    fork_op op(nalternatives, value.type());
+    fork_op op(nalternatives, value.type(), constant);
     return jlm::rvsdg::simple_node::create_normalized(region, op, { &value });
   }
+
+  bool Constant_;
```
phate, line 117: Could we call it `IsConstant_`?

jlm/hls/ir/hls.hpp
```diff
@@ -101,13 +106,15 @@ class fork_op final : public jlm::rvsdg::simple_op
   }
 
   static std::vector<jlm::rvsdg::output *>
-  create(size_t nalternatives, jlm::rvsdg::output & value)
+  create(size_t nalternatives, jlm::rvsdg::output & value, bool constant = false)
```
phate, line 109: `isConstant`

jlm/hls/ir/hls.hpp
```diff
@@ -80,6 +80,11 @@ class fork_op final : public jlm::rvsdg::simple_op
       : jlm::rvsdg::simple_op({ type }, std::vector<jlm::rvsdg::port>(nalternatives, type))
   {}
 
+  fork_op(size_t nalternatives, const rvsdg::type & type, bool constant)
```
phate, line 83: `isConstant`

jlm/hls/ir/hls.hpp
```diff
@@ -80,6 +80,11 @@ class fork_op final : public jlm::rvsdg::simple_op
...
+      : rvsdg::simple_op({ type }, std::vector<rvsdg::port>(nalternatives, type)),
+        Constant_(constant)
+  {}
+
   bool
   operator==(const jlm::rvsdg::operation & other) const noexcept override
```
phate, line 89: You added an attribute to the operation, but did not modify the equality operator. Is the equality operator still correct?

jlm/hls/backend/rvsdg2rhls/add-forks.cpp
```diff
@@ -36,14 +37,15 @@ add_forks(jlm::rvsdg::region * region)
         add_forks(structnode->subregion(n));
       }
     }
+    bool constant = dynamic_cast<const rvsdg::bitconstant_op *>(&node->operation())
```
phate, line 40: `bool isConstant = is<bitconstant_op>() || is<ctlconstant_op>();`

jlm/hls/ir/hls.hpp
phate, line 117: Make that property private and add an `IsConstant()` accessor to the class.

jlm/hls/backend/rhls2firrtl/RhlsToFirrtlConverter.cpp
```diff
@@ -461,6 +466,8 @@ RhlsToFirrtlConverter::MlirGenLoopConstBuffer(const jlm::rvsdg::simple_node * no
 circt::firrtl::FModuleOp
 RhlsToFirrtlConverter::MlirGenFork(const jlm::rvsdg::simple_node * node)
 {
+  auto op = dynamic_cast<const jlm::hls::fork_op *>(&node->operation());
+  bool constant = op->Constant_;
```
phate, line 470: `auto isConstant = op->IsConstant()`

phate (changes requested):

jlm/hls/ir/hls.hpp
```diff
@@ -101,13 +107,22 @@ class fork_op final : public jlm::rvsdg::simple_op
...
   }
+
+  bool
+  IsConstant() const
+  {
+    return IsConstant_;
+  }
+
+private:
+  bool IsConstant_;
```
phate, line 125:
```
[[nodiscard]] bool
IsConstant() const noexcept
{
  return IsConstant_;
}
```

phate (changes requested):
I thought about this a little bit more, and I am sorry, but I will block it again. I would like to see two more things:
1. Some documentation on the fork operation and the `IsConstant_` attribute. I had sort of(-ish) an idea what the fork operation might do beforehand, but I do not understand how this works with `IsConstant_` now.
2. For once, this is a rather small and contained HLS PR. Let's take this opportunity and start with some unit tests. It would be nice to see at least unit tests for the fork operation. You already introduced a bug in the first iteration. This should have already been a hint to add some unit tests.

sjalander:
@phate Should I also refactor the fork_op class in the same PR?

jlm/hls/ir/hls.hpp
```diff
@@ -100,14 +138,38 @@ class fork_op final : public jlm::rvsdg::simple_op
...
   }
+
+  /**
+   * Cechk if a fork is a constant fork (CFORK).
+   *
+   * /return True if the fork is a constant fork, i.e., the input of the fork is a constant, else
+   * false.
+   */
+  bool
+  IsConstant() const
```
phate, line 166:
```
[[nodiscard]] bool
IsConstant() const noexcept
{
...
}
```

jlm/hls/backend/rvsdg2rhls/add-forks.cpp
```diff
@@ -36,14 +37,15 @@ add_forks(jlm::rvsdg::region * region)
         add_forks(structnode->subregion(n));
       }
     }
+    bool isConstant = rvsdg::is<rvsdg::bitconstant_op>(node->operation())
+                   || rvsdg::is<rvsdg::ctlconstant_op>(node->operation());
```
phate, line 41: This will only work for those constants, but what about floating point constants or other constants?

I guess a better way is to check whether a node has no input to determine whether it is a constant.

jlm/hls/ir/hls.hpp
```diff
@@ -100,14 +138,38 @@ class fork_op final : public jlm::rvsdg::simple_op
...
+  {
+    return IsConstant_;
+  }
+
+private:
+  bool IsConstant_ = false;
```
phate, line 172: For the future: I am not sure whether there is a particular reason that I a do not see as to why this property is here, but it might be better to implement this when you lower the fork operation. It should be trivial to determine whether a constant is connected to a fork or not. It is  a simple tracing of the fork input to the next node and then checking whether the node is a constant or not.

phate (dismissed):
I left some minor comments. Some are more FYI, others are minor improvements, but nothing major. Thanks for adding the documentation. It helped me to better understand what is going on, and will definitely help future us.

@sjalander Let's do the refactoring in another PR.

jlm/hls/ir/hls.hpp
sjalander: Agreed, the minor benefit of having this could be during debug or viewing the graph as for example the debug_string() depends on it.
Could change IsConstant() to check the inputs of the source node.

jlm/hls/backend/rvsdg2rhls/add-forks.cpp
sjalander: Changed to checking if ninputs() == 0

jlm/hls/ir/hls.hpp
sjalander: Fixed

phate (approved):

jlm/hls/backend/rvsdg2rhls/add-forks.cpp
```diff
@@ -36,14 +37,15 @@ add_forks(jlm::rvsdg::region * region)
...
     {
       auto out = node->output(i);
       if (out->nusers() > 1)
       {
-        std::vector<jlm::rvsdg::input *> users;
-        users.insert(users.begin(), out->begin(), out->end());
-        auto fork = hls::fork_op::create(out->nusers(), *out);
+        std::vector<rvsdg::input *> users(out->begin(), out->end());
+        auto fork = hls::fork_op::create(out->nusers(), *out, isConstant);
         for (size_t j = 0; j < users.size(); j++)
```
phate, line 49: I just saw it now, but the[ `divert_users()` method of an output](https://github.com/phate/jlm/blob/feb7a6c60627a49bff98b42c5540eabb420a83fa/jlm/rvsdg/node.hpp#L354C1-L361C4) does exactly this.

phate: The same a few lines above. The entire add-forks pass could use some refactoring. Merge it for now, it is not part of this PR.

phate (approved):

jlm/hls/backend/rvsdg2rhls/add-forks.cpp
davidmetz: divert_users would not work here, since the fork is a user of the same output, and only the other users should be diverted and not it. And the other users are all diverted to separate outputs of the fork, since that's the whole point.

phate: Right, it also seems to create a 1-to-1 correspondence between a single user and a fork output.

phate:
@sjalander Please merge eventually so we get rid off some open PRs.

PR #479, opened 2024-05-10
CNE: Simple support for hls

sjalander:


phate (changes requested):
The library dependencies are:

libutil <-- librvsdg <-- libllvm <-- libhls

By introducing something from hls to libllvm, you are creating a cyclic dependency as libllvm all in a sudden is also dependent on libhls. This is not the right way to go about this.

If you want some CNE support for HLS, then you have currently two possibilities:
1. We factor out the mechanisms from CNE and abstract it away from the concrete operations such that it can also be used in HLS. This will require some thinking and work.
2. You implement your own CNE in libhls. You already went the same route with DNE. This means we are duplicating even more code though.

As it is, this is not going to be merged.

sjalander:
I guess the ideal would be to go with option 1 and also do the same for DNE.

phate:
@sjalander Yes, the best would be to go for option 1, but that would require some thinking and engineering time.

sjalander:
@phate
Went the easier route with duplicating code and created a HLS specific CNE optimization.

phate (changes requested):
Just asking to be sure: This PR does nothing else than cloning the CNE to the HLS namespace, right? You did not modify the CNE implementation in any?

Could you introduce a FIXME on the `cne` class that informs about it only being a copy from the LLVM cne (with explicit reference to the other implementation, please). If you made modifications, then please mention those as well. I will merge the PR with the FIXME.

I will also try to prioritize some engineering time in getting general support for common optimizations such that we can avoid the code duplication in the future.

sjalander:
@phate 
No, it also has these changes applied to it
https://github.com/phate/jlm/pull/479/commits/2bbab4a181f7c0e4b7db6e8b584ba515834d2a46

phate (approved):

PR #480, opened 2024-05-11
HLS: Configurable pointer size

sjalander:


jlm/hls/backend/rhls2firrtl/RhlsToFirrtlConverter.cpp
```diff
@@ -321,7 +321,7 @@ RhlsToFirrtlConverter::MlirGenSimpleNode(const jlm::rvsdg::simple_node * node)
       int bits = JlmSize(pointeeType);
       if (dynamic_cast<const jlm::rvsdg::bittype *>(pointeeType))
       {
-        ;
+        pointeeType = nullptr;
```
phate, line 324: I do not understand why you set `pointeeType` to `nullptr` here. This does not make sense to me at all. The only place where `pointeeType` is used is in the `JlmSize()` method, and this method does not really handle `nullptr` gracefully.

jlm/hls/backend/rhls2firrtl/RhlsToFirrtlConverter.cpp
```diff
@@ -3906,7 +3907,7 @@ RhlsToFirrtlConverter::GetModuleName(const jlm::rvsdg::node * node)
...
```
phate, line 3910: The same as above here.

jlm/hls/ir/hls.cpp
```diff
@@ -8,6 +8,8 @@
 namespace jlm::hls
 {
 
+size_t PointerSizeInBits = 64;
```
phate, line 11: Yeah no. Let's not introduce global variables. This should be handed in the respective methods as parameter.

phate (changes requested):
Introducing global variables is not the way to go.

PR #482, opened 2024-05-11
Fix bug in isOrContains type helper function

haved:
The diff should be self-explanatory :)


phate (dismissed):
Ouch, that is a big one. Great find! This might also fix some issues in the Steensgaard pipeline.

I will approve, but could you do me the favor and put some unit tests on this function if you are not in a hurry with this PR? This is a central function and I am not sure whether it actually has some tests on it.

phate (approved):
Thanks a lot!

PR #483, opened 2024-05-14
HLS: Initial conversion of llvm intrinsics generated by Clang

sjalander:
Currently only llvm.smin.i32 is handled

### DIFF
--- a/jlm/hls/backend/rvsdg2rhls/rvsdg2rhls.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/rvsdg2rhls.cpp
@@ -39,6 +39,7 @@
 #include <jlm/llvm/opt/InvariantValueRedirection.hpp>
 #include <jlm/llvm/opt/inversion.hpp>
 #include <jlm/llvm/opt/reduction.hpp>
+#include <jlm/rvsdg/bitstring/comparison.hpp>
 #include <jlm/rvsdg/traverser.hpp>
 #include <jlm/rvsdg/view.hpp>
 #include <jlm/util/Statistics.hpp>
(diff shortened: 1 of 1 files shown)


phate (changes requested):
I stopped reviewing after I saw that this is extremely limited and would only cover cases for the intrinsic being in the lambda subregion (and not even in any gamma/theta node of the lambda).

The absolutely right way to go about this would be to:
1. Introduce an smin operation similar like we did for the[ memcpy intrinsic](https://github.com/phate/jlm/blob/master/jlm/llvm/ir/operators/MemCpy.hpp)
2. Detect the smin intrinsic in the frontend when converting LLVM to RVSDG similarly like we do for [memcpy](https://github.com/phate/jlm/blob/56b8e7abb0351fcdd004e35df5444ddd2df51f93/jlm/llvm/frontend/LlvmInstructionConversion.cpp#L763C1-L792C2).
3. Add support for converting the smin operation/node back to an LLVM intrinsic in the backend.
4. Add a pass in the HLS part that checks for the smin/node and converts it to a gamma.

If you want to restrict it to the HLS part of the compiler (and not touch the other parts of the compiler), then you should:

1. Add an HLS pass that iterates through all the (sub-)regions of an RVSDG/lambda (you only have one lambda I guess). There is no need for a toplogical sorting, so taking the [nodes directly](https://github.com/phate/jlm/blob/56b8e7abb0351fcdd004e35df5444ddd2df51f93/jlm/rvsdg/region.hpp#L369C1-L369C27) should be enough. For every call node that you find, you :
2. [Compute the CallTypeClassifier](https://github.com/phate/jlm/blob/56b8e7abb0351fcdd004e35df5444ddd2df51f93/jlm/llvm/ir/operators/call.hpp#L455C1-L456C43)
3. If the call type classifier is an [external call](https://github.com/phate/jlm/blob/56b8e7abb0351fcdd004e35df5444ddd2df51f93/jlm/llvm/ir/operators/call.hpp#L161C1-L165C4) and the [argument ](https://github.com/phate/jlm/blob/56b8e7abb0351fcdd004e35df5444ddd2df51f93/jlm/llvm/ir/operators/call.hpp#L209C1-L214C4) is a smin intrisinc, then replace this node with a gamma node.

PR #484, opened 2024-05-17
Perf: Avoid visiting nodes for a "second" time multiple times in TarjanSCC

haved:
This does not affect correctness, but should boost performance by avoiding unnecessary visits of nodes that are already parts of finished SCCs.

When implementing DFS using a stack, I had forgotten that the same node can be added multiple times before it actually gets visited. In a graph like
```
A -------> B
 \         ^
  --> C --/
```
Visiting node A will push B and C. Since we then visit C first, it will also have to push B to process B before C can be popped. The edge `A->B` should then be ignored when popping B the second time.

### DIFF
--- a/jlm/util/TarjanScc.hpp
+++ b/jlm/util/TarjanScc.hpp
@@ -59,7 +59,8 @@ FindStronglyConnectedComponents(
   // The lowest order seen through any path of edges that do not enter finished SCCs
   std::vector<int64_t> lowLink(numNodes);
 
-  // The DFS stack is a regular DFS traversal stack
+  // The DFS stack is a regular DFS traversal stack.
+  // Note that the same element can be pushed many times, but will only be visited twice
   std::stack<size_t> dfsStack;
   // The SCC stack is only popped once an SCC is found
   std::stack<size_t> sccStack;
@@ -89,8 +90,9 @@ FindStronglyConnectedComponents(
             dfsStack.push(next);
         }
       }
-      else // This is the second time node is visited, all children have been processed
+      else if (order[node] != SCC_FINISHED)
       {
+        // This is the second time node is visited, all children have been processed
         dfsStack.pop();
         for (auto next : successors(node))
         {
--- a/jlm/util/TarjanScc.hpp
+++ b/jlm/util/TarjanScc.hpp
@@ -90,10 +90,14 @@ FindStronglyConnectedComponents(
             dfsStack.push(next);
         }
       }
-      else if (order[node] != SCC_FINISHED)
+      else
       {
-        // This is the second time node is visited, all children have been processed
+        // This node has been visited before, so all children have been processed
         dfsStack.pop();
+
+        if (order[node] == SCC_FINISHED)
+          continue;
+
         for (auto next : successors(node))
         {
           // Ignore edges to nodes that are already part of a finished SCC
--- a/jlm/util/TarjanScc.hpp
+++ b/jlm/util/TarjanScc.hpp
@@ -90,13 +90,13 @@ FindStronglyConnectedComponents(
             dfsStack.push(next);
         }
       }
-      else
+      else if (order[node] == SCC_FINISHED) // This node has already been fully processed
       {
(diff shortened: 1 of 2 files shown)


jlm/util/TarjanScc.hpp
```diff
@@ -89,9 +90,14 @@ FindStronglyConnectedComponents(
...
       }
-      else // This is the second time node is visited, all children have been processed
+      else
       {
+        // This node has been visited before, so all children have been processed
         dfsStack.pop();
+
+        if (order[node] == SCC_FINISHED)
+          continue;
+
```
phate, line 100: Maybe turn this around instead of using `continue`:

```
if (order[node] != SCC_FINISHED)
{
...
}
```

jlm/util/TarjanScc.hpp
```diff
@@ -89,9 +90,14 @@ FindStronglyConnectedComponents(
...
         for (auto next : successors(node))
         {
           // Ignore edges to nodes that are already part of a finished SCC
```
phate, line 103: Maybe turning it here around as well instead of using `continue`.

phate (approved):
I just have two absolutely minor suggestions, not sure whether they are worth it.

I suspect not, but just asking to be sure: Is there any easy(-ish) way to put a unit test on this problem such that it won't happen again?

haved:
The change you suggest is what I did first, but I still need to pop the element. One alternative would be to break it into if/elseif/else, which might be nicer.

Regarding unit testing it, it can be done with a custom callback, that checks that each node is only queried about its successor twice. I can absolutely do that

haved:
I added a test and changed the code to avoid `continue`

tests/jlm/util/TestTarjanScc.cpp
```diff
@@ -300,3 +300,50 @@ TestDiamondChainWithBackEdge()
...
+  std::vector<size_t> topologicalOrder;
+  auto numSccs = jlm::util::FindStronglyConnectedComponents(
+      numNodes,
+      GetSuccessors,
+      sccIndex,
+      topologicalOrder);
+
+  JLM_ASSERT(numSccs == 4);
+  for (size_t timesQueried : successorsQueried)
+    JLM_ASSERT(timesQueried <= 2);
```
phate, line 336: So I might be wrong here, as I do not fully know the details by heart, but should this not be `== 1` for each of the nodes, i.e., the successor for a node was only asked for once (and therefore the node was only visited once?

phate: If the different nodes have a different number in successorQueried, i.e., 0 or 1, or 2, maybe we can be more precise here and assert for all the nodes, e.g.,  `assert(successorsQueried[0] == 0)`

phate (approved):
Sorry that it took so long to give you a review. Approved with minor comment.

jlm/util/TarjanScc.hpp
```diff
@@ -89,9 +90,14 @@ FindStronglyConnectedComponents(
             dfsStack.push(next);
         }
       }
-      else // This is the second time node is visited, all children have been processed
+      else if (order[node] == SCC_FINISHED) // This node has already been fully processed
```
phate, line 93: I like that much better in terms of readability.

tests/jlm/util/TestTarjanScc.cpp
haved: Currently every node is queried for its children exactly twice. Once when it is visited for the first name (and needs to ensure all successors are visited), and once again after all successors have been visited. These two visits can be separated by visits of an arbitrary amount of other nodes, so it has to query at least some nodes twice. Anything more than that is unnecessary.

The usual call stack based dfs places the iterators on the call stack, but I assume that getting the list of successors again is cheaper than adding them to the explicit stack.

jlm/util/TarjanScc.hpp
haved: @phate it looks like this comment might have started a new review that is not accepted? I'm not able to merge it seems

PR #485, opened 2024-05-20
[Andersen] Make InterProceduralGraphConversion emit the _VolatileNode varieties

haved:
With this change, the `StoreVoltatileOperation` and `LoadVolatileOperation` are placed in `StoreVolatileNode` and `LoadVolatileNode`, respectively. I'm not sure how one would best test that you never place one of the operations in a regular `simple_node`.

Andersen is modified slightly, to use the superclasses.

I also had to add `#include <algorithm>` to `graph.cpp`, since `std::any_of` apparently isn't transitively included from any of the other headers anymore.

### DIFF
--- a/jlm/llvm/frontend/InterProceduralGraphConversion.cpp
+++ b/jlm/llvm/frontend/InterProceduralGraphConversion.cpp
@@ -538,8 +538,11 @@ ConvertThreeAddressCode(
             { typeid(CallOperation), Convert<CallNode, CallOperation> },
             { typeid(LoadNonVolatileOperation),
               Convert<LoadNonVolatileNode, LoadNonVolatileOperation> },
+            { typeid(LoadVolatileOperation), Convert<LoadVolatileNode, LoadVolatileOperation> },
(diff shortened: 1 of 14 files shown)


phate (approved):
LGTM

phate:
@haved Just a wild guess (I have not looked at the logs), but with you changing it in the Andersen, you added support for the volatile variants to it. However, the provisioning and the memory state encoder do not support it yet.

haved:
That is very true. I'm just surprised that it gets that far into the GCC torture tests before having an issue. Maybe most tests don't include volatile? I will see if I can pinpoint the issue, but the thesis results don't care about encoding, so at worst I'll only make the PtG and then discard it when necessary.

haved:
@phate I added handling of both volatile and non volatile to the encoder, and it seems to work now. I wasn't too sure how it should be implemented, so have a look. I think it is a bit weird that the InputIterator and OutputIterator give references as their values, despite the `value_type` being a pointer, but it doesn't really affect anything. I just had to use a for loop to convert between reference and pointer.

PR #486, opened 2024-05-23
Add support for volatile operations to alias analyses pipeline [Andersen] [SteensgaardAgnostic]

phate:
This PR does the following:
1. Adds support for volatile load, store, and memcpy nodes to memory state encoder
2. Adds support for volatile load, store, and memcpy nodes to Steensgaard alias analysis
3. Adds support for volatile load, store, and memcpy nodes to Andersen alias analysis
4. Adds support for volatile load, store, and memcpy nodes to region-aware memory node provider
5. Tighten the handling of simple nodes in the memory state encoder

### DIFF
--- a/jlm/llvm/ir/operators/Load.cpp
+++ b/jlm/llvm/ir/operators/Load.cpp
@@ -69,6 +69,17 @@ LoadNonVolatileNode::MemoryStateOutputs() const noexcept
   return { MemoryStateOutputIterator(output(1)), MemoryStateOutputIterator(nullptr) };
 }
 
+LoadNonVolatileNode &
+LoadNonVolatileNode::CopyWithNewMemoryStates(
+    const std::vector<rvsdg::output *> & memoryStates) const
+{
+  return CreateNode(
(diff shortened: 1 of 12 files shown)


phate:
@haved  Here is my version for handling volatile nodes in the memory state encoder. It was heavily inspired by your PR, but handles a few things slightly differently. What I do not have yet is unit tests for the volatile cases. Can you wait for another day?

phate:
The failing tests (for steensgaard agnostic) is due to the tightening and not having all nodes in the if-then-else statement. I will dig into it.

haved:
@phate yes I can wait for tests in this PR before it is merged

phate:
@haved The PR is now ready for review. I did not manage to add unit tests yet, but consider the successful runs of the steensgaard-agnostic and andersen-agnostic checks test enough. I opened issue #488 to not forget about the unit tests.

haved (approved):
Very nice, I like the separation between `EncodeLoad` and `ReplaceLoadNode`

PR #487, opened 2024-05-25
Fix three address code conversion for volatile loads and stores

phate:
The three address code conversion for volatile loads and stores did not create the correct LoadVolatileNode and StoreVolatileNode classes, but simple_node classes.

This PR does the following:
1. Fix the three address code conversion for volatile loads
2. Fix the three address code conversion for volatile stores
3. Refactors the conversion to be better readable
4. Adds unit tests for the conversions
5. Adds support functions to easier test other conversions in the future


phate:
Preferably I would liked to spell the mapping between operations and nodes out explicitly, but that was too big of a change for the current time constraints. This is a reasonable compromise with the added unit tests.

phate:
@haved I needed this PR first to properly test the memory state encoder PR (without proper conversion, nothing is testable).

haved (approved):
Nice, I like the rewritten handler. The tests are nice

PR #490, opened 2024-05-27
[Andersen] Add queue for new edges to the worklist, and enable picking worklist policy

haved:
This PR fixes a bug in the worklist solver where pointees could be added to `GetPointsToSet(node)`, while the points-to set was being iterated over.

Running the worklist is now divided into two functions, to convert runtime configuration into compile-time configuration through dispatching the corresponding templated version.

With this PR, the only configuration of the worklist solver is the worklist policy.

The default configuration checks an environment variable to be overridden. This is to make development easier, and not because I think environment variables are a good configuration interface :)

### DIFF
--- a/jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
+++ b/jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
@@ -731,16 +731,6 @@ public:
      * The number of items that were popped from the worklist before the solution converged.
      */
     size_t NumWorkItemsPopped{};
-
-    /**
-     * The number of cycles detected by online cycle detection, if enabled.
-     */
-    std::optional<size_t> NumOnlineCyclesDetected{};
-
-    /**
-     * The number of unifications made by online cycle detection, if enabled.
-     */
-    std::optional<size_t> NumOnlineCycleUnifications{};
   };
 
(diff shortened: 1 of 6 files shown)


jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -701,6 +703,36 @@ class PointerObjectConstraintSet final
   using ConstraintVariant =
       std::variant<SupersetConstraint, StoreConstraint, LoadConstraint, FunctionCallConstraint>;
 
+  enum class WorklistSolverPolicy
+  {
+    LRF,
```
phate, line 708: Could we spell them out? While I certainly know what Fifo and Lifo means, I had to think now a bit about LRF (least recently fired?) Also, could we put the papers references as documentation behind them such that we have some bread crumbs here.

phate (approved):
Minor comment

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
haved: @phate I changed them to be written out, and added a reference to the paper, and to the corresponding `jlm::util::___Worklist` classes.

phate (approved):
LGTM

PR #494, opened 2024-05-29
Move memory state operations to MemoryStateOperations.[hc]pp

phate:
This PR moves all operations that only work on memory states into the same header and implementation file. This ensures that they are logically grouped together instead of being split up over multiple files.

The following operations are affected:
1. MemStateOperator, MemStateMergeOperator, MemStateSplitOperator: Moved from operators.[hc]pp
2. Lambda[Entry/Exit]MemStateOperator and Call[Entry/Exit]MemStateOperator: Moved from alias-analysis/Operators.[hc]pp


haved (approved):
Nice change, the new structure makes sense

PR #498, opened 2024-05-31
Clean up memory state operations

phate:
The PR does the following:
1. Rename the operations to consistent names
2. Adds some documentation
3. Cleans up some minor code smells


haved (approved):
Very nice, the new comments on the Lambda/Call versions finally made me realize why they exist.

PR #499, opened 2024-06-02
Introduce shared ptr for types

caleridas:
This change introduces shared_ptr for types, uses them in most places
for storage, and starts putting in singleton objects for several primitive
types.

The singletons use the "untracked feature" for std::shared_ptr -- the
"object used for lifetime tracking" can be separate from the "pointer
that the shared_ptr logically points to". It is possible to make the
former a "null" object in which case we actually have a non-reference
counted "shared" object that is supposed to have eternal lifetime.

This small trick allows to skip the cost of refcount inc/dec for these
eternal primitive objects. According to measurements, "create + destroy"
of e.g. bittype(32) is at ~6 CPU clock cycles (including function call
overhead), so this is already cheaper than before.

### DIFF
--- a/jlm/hls/backend/rvsdg2rhls/instrument-ref.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/instrument-ref.cpp
@@ -90,7 +90,7 @@ instrument_ref(llvm::RvsdgModule & rm)
   //  addr, width, memstate
   jlm::llvm::FunctionType loadFunctionType(
       { jlm::llvm::PointerType::Create().get(),
-        &jlm::rvsdg::bit64,
+        &*jlm::rvsdg::bittype::Create(64),
         llvm::iostatetype::create().get(),
         llvm::MemoryStateType::Create().get() },
       { llvm::iostatetype::create().get(), llvm::MemoryStateType::Create().get() });
@@ -102,8 +102,8 @@ instrument_ref(llvm::RvsdgModule & rm)
   // addr, data, width, memstate
   jlm::llvm::FunctionType storeFunctionType(
       { jlm::llvm::PointerType::Create().get(),
-        &jlm::rvsdg::bit64,
-        &jlm::rvsdg::bit64,
+        &*jlm::rvsdg::bittype::Create(64),
+        &*jlm::rvsdg::bittype::Create(64),
         llvm::iostatetype::create().get(),
         jlm::llvm::MemoryStateType::Create().get() },
       { llvm::iostatetype::create().get(), jlm::llvm::MemoryStateType::Create().get() });
@@ -115,7 +115,7 @@ instrument_ref(llvm::RvsdgModule & rm)
   // addr, size, memstate
   jlm::llvm::FunctionType allocaFunctionType(
       { jlm::llvm::PointerType::Create().get(),
-        &jlm::rvsdg::bit64,
+        &*jlm::rvsdg::bittype::Create(64),
         llvm::iostatetype::create().get(),
         jlm::llvm::MemoryStateType::Create().get() },
       { llvm::iostatetype::create().get(), jlm::llvm::MemoryStateType::Create().get() });
@@ -246,7 +246,7 @@ instrument_ref(
       }
       auto data = node->input(1)->origin();
       auto dbt = dynamic_cast<const jlm::rvsdg::bittype *>(&data->type());
-      if (*dbt != jlm::rvsdg::bit64)
+      if (*dbt != *jlm::rvsdg::bittype::Create(64))
       {
         jlm::llvm::zext_op op(dbt->nbits(), 64);
         data = jlm::rvsdg::simple_node::create_normalized(data->region(), op, { data })[0];
--- a/jlm/hls/backend/rvsdg2rhls/instrument-ref.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/instrument-ref.cpp
@@ -91,9 +91,9 @@ instrument_ref(llvm::RvsdgModule & rm)
   jlm::llvm::FunctionType loadFunctionType(
       { jlm::llvm::PointerType::Create().get(),
         &*jlm::rvsdg::bittype::Create(64),
-        llvm::iostatetype::create().get(),
+        llvm::iostatetype::Create().get(),
         llvm::MemoryStateType::Create().get() },
-      { llvm::iostatetype::create().get(), llvm::MemoryStateType::Create().get() });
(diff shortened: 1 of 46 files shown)


jlm/hls/backend/rvsdg2rhls/instrument-ref.cpp
```diff
@@ -90,10 +90,10 @@ instrument_ref(llvm::RvsdgModule & rm)
   //  addr, width, memstate
   jlm::llvm::FunctionType loadFunctionType(
       { jlm::llvm::PointerType::Create().get(),
-        &jlm::rvsdg::bit64,
-        llvm::iostatetype::create().get(),
+        &*jlm::rvsdg::bittype::Create(64),
```
phate, line 93: Is this not equivalent to `jlm::rvsdg::bittype::Create(64).get()`?

The pattern is all over the place. I will only comment on it here.

jlm/rvsdg/control.cpp
```diff
@@ -41,13 +38,34 @@ ctltype::operator==(const jlm::rvsdg::type & other) const noexcept
...
+  if (nalternatives < 4)
+  {
+    return std::shared_ptr<const jlm::rvsdg::ctltype>(
+        std::shared_ptr<void>(),
+        &static_instances[nalternatives]);
+  }
+  else
+  {
+    return std::make_shared<ctltype>(nalternatives);
+  }
```
phate, line 67: Could we perform the check for `nalternaives == 0` here?

jlm/rvsdg/bitstring/type.cpp
```diff
@@ -28,16 +28,36 @@ bittype::operator==(const jlm::rvsdg::type & other) const noexcept
...
+
+  if (nbits <= 64)
+  {
+    return std::shared_ptr<const bittype>(std::shared_ptr<void>(), &static_instances[nbits]);
+  }
+  else
+  {
+    return std::make_shared<bittype>(nbits);
+  }
+}
```
phate, line 61: Could we perform the check for `nbits == 0` here?

jlm/rvsdg/operation.cpp
```diff
@@ -20,14 +20,14 @@ port::port(const jlm::rvsdg::type & type)
...
-port::port(std::unique_ptr<jlm::rvsdg::type> type)
+port::port(std::shared_ptr<const jlm::rvsdg::type> type)
     : type_(std::move(type))
 {}
 
 bool
 port::operator==(const port & other) const noexcept
 {
-  return *type_ == *other.type_;
+  return type_ == other.type_ || *type_ == *other.type_;
```
phate, line 30: This is a performance optimization, right? First checking for pointer in case of static instances, and then if that is false, invoking `operator==()` on the pointer. Could we put a comment on this expressing it?

jlm/rvsdg/operation.hpp
```diff
@@ -33,37 +33,17 @@ class port
 
   port(const jlm::rvsdg::type & type);
 
-  port(std::unique_ptr<jlm::rvsdg::type> type);
+  port(std::shared_ptr<const jlm::rvsdg::type> type);
 
-  inline port(const port & other)
-      : type_(other.type_->copy())
-  {}
+  inline port(const port & other) = default;
```
phate, line 38: Remove `inline`

jlm/rvsdg/operation.hpp
```diff
@@ -33,37 +33,17 @@ class port
...
 
-  inline port(port && other)
-      : type_(std::move(other.type_))
-  {}
+  inline port(port && other) = default;
```
phate, line 40: Remove `inline`

tests/test-types.hpp
```diff
@@ -26,7 +26,7 @@ class valuetype final : public rvsdg::valuetype
   virtual bool
   operator==(const rvsdg::type & other) const noexcept override;
 
-  virtual std::unique_ptr<rvsdg::type>
+  virtual std::shared_ptr<const jlm::rvsdg::type>
```
phate, line 29: Remove `virtual`

tests/test-types.hpp
```diff
@@ -45,7 +45,7 @@ class statetype final : public rvsdg::statetype
...
```
phate, line 48: Remove `virtual`

phate (changes requested):
Looks good to me. Only minor comments.

jlm/rvsdg/bitstring/type.cpp
caleridas: done

jlm/rvsdg/operation.hpp
caleridas: done

jlm/rvsdg/control.cpp
caleridas: done

jlm/hls/backend/rvsdg2rhls/instrument-ref.cpp
caleridas: Yes that is correct, and it is (mostly) to specific to functiontype which wants pointers. The "&*" pattern will go way when this is converted to also used shared_ptr -- and then "&*" is a bit easier to search&replace for than "something.get()"

phate: Got it. Thanks!

jlm/rvsdg/bitstring/type.cpp
```diff
@@ -28,16 +28,41 @@ bittype::operator==(const jlm::rvsdg::type & other) const noexcept
...
-const bittype bit1(1);
-const bittype bit8(8);
-const bittype bit16(16);
-const bittype bit32(32);
-const bittype bit64(64);
+std::shared_ptr<const bittype>
+bittype::Create(std::size_t nbits)
+{
+  static const bittype static_instances[65] = {
+    bittype(0),  bittype(1),  bittype(2),  bittype(3),  bittype(4),  bittype(5),  bittype(6),
```
phate, line 41: Could we leave a comment for `bittype(0)` here similarly to `ctltype(0)`?

phate (approved):
Great job! I had one more minor comment. Up to you whether you want to act on it.

jlm/rvsdg/bitstring/type.cpp
caleridas: Well, bitstring(0) is not technically illegal, it would just be an empty type that does not store any value (like "std::nullopt_t") -- I am more inclined of removing the throw below.

PR #500, opened 2024-06-03
Rework bitstring comparison ops using templates instead of macros

caleridas:
Remove all code generation macros for instantiating bitstring comparison operation classes. Replace this with normal C++ template classes.


caleridas:
Same as for the computation ops, somehow forgot that comparison ops exist...

phate (approved):
LGTM

PR #505, opened 2024-06-07
Add overloads to shared_ptr of type for various ops

caleridas:
Add overloaded constructors and/or creator functions that take "shared_ptr<const type>" instead of "const type &" arguments (and hence can take over reference from caller, instead of requiring a copy operation).

In some cases, the "type copy" methods can be removed as they have no callers anymore. For others, the superseded methods will be removed once all callers are converted.

This sets the groundwork for gradually converting all call sites to use shared type references.

### DIFF
--- a/jlm/llvm/ir/operators/Load.hpp
+++ b/jlm/llvm/ir/operators/Load.hpp
@@ -193,7 +193,19 @@ public:
       size_t alignment)
       : LoadOperation(
           CreateOperandPorts(numMemoryStates),
-          CreateResultPorts(loadedType, numMemoryStates),
+          CreateResultPorts(
+              std::static_pointer_cast<const rvsdg::valuetype>(loadedType.copy()),
+              numMemoryStates),
+          alignment)
+  {}
+
+  LoadVolatileOperation(
+      std::shared_ptr<const rvsdg::valuetype> loadedType,
+      size_t numMemoryStates,
+      size_t alignment)
+      : LoadOperation(
+          CreateOperandPorts(numMemoryStates),
+          CreateResultPorts(std::move(loadedType), numMemoryStates),
           alignment)
   {}
 
@@ -232,9 +244,10 @@ private:
   }
 
   static std::vector<rvsdg::port>
-  CreateResultPorts(const rvsdg::valuetype & loadedType, size_t numMemoryStates)
+  CreateResultPorts(std::shared_ptr<const rvsdg::valuetype> loadedType, size_t numMemoryStates)
   {
-    std::vector<rvsdg::port> ports({ loadedType, iostatetype() });
+    std::vector<rvsdg::port> ports(
+        { rvsdg::port(std::move(loadedType)), rvsdg::port(iostatetype::Create()) });
     std::vector<rvsdg::port> states(numMemoryStates, { MemoryStateType() });
     ports.insert(ports.end(), states.begin(), states.end());
     return ports;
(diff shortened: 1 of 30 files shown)


caleridas:
This adds several overloads to existing methods, with the intent of later superseding the methods that rely on "type copy". It is too cumbersome to do all conversions in a single pass, so I think I need to go through this intermediate step.

jlm/rvsdg/structural-node.hpp
```diff
@@ -77,20 +77,29 @@ class structural_input : public node_input
 public:
   virtual ~structural_input() noexcept;
 
-protected:
```
phate, line 80: Why did you remove it?

jlm/rvsdg/structural-node.hpp
```diff
@@ -77,20 +77,29 @@ class structural_input : public node_input
...
   structural_input(
       jlm::rvsdg::structural_node * node,
       jlm::rvsdg::output * origin,
       const jlm::rvsdg::port & port);
 
-public:
```
phate, line 86: See comment above.

jlm/rvsdg/structural-node.hpp
```diff
@@ -113,14 +122,19 @@ class structural_output : public node_output
 public:
   virtual ~structural_output() noexcept;
 
-protected:
```
phate, line 116: Why removed?

jlm/rvsdg/structural-node.hpp
```diff
@@ -113,14 +122,19 @@ class structural_output : public node_output
...
   structural_output(jlm::rvsdg::structural_node * node, const jlm::rvsdg::port & port);
 
-public:
```
phate, line 119: Same here

phate (changes requested):

jlm/rvsdg/structural-node.hpp
caleridas: make_shared (and make_unique) need a public constructor, otherwise they cannot construct the class. It is also essential to use make_shared<X>(...) instead of shared_ptr<X>(new X(...)) as the former will cause one allocation, the latter causes two

if the constructor needs to be made "really" private (but you generally do not need that, unless you have something like enabled_shared_from_this), it requires some construction like:

`class X {
private:
  struct private_tag {};
public:
  X(Arg1 arg1, Arg2 arg2, ..., private_tag);

  shared_ptr<X> Create(Arg1 arg1, Arg2 arg2, ...) { return make_shared<X>(arg1, arg2, ..., private_tag {}); }
`

but that is a bit much for now, also since I need to do some more cleanup later

jlm/llvm/ir/operators/Load.hpp
```diff
@@ -193,7 +193,19 @@ class LoadVolatileOperation final : public LoadOperation
       size_t alignment)
       : LoadOperation(
           CreateOperandPorts(numMemoryStates),
-          CreateResultPorts(loadedType, numMemoryStates),
+          CreateResultPorts(
+              std::dynamic_pointer_cast<const rvsdg::valuetype>(loadedType.copy()),
```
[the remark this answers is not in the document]
caleridas: right, changed it

jlm/rvsdg/structural-node.hpp
caleridas: as above

phate (approved):

PR #506, opened 2024-06-08
Change testnode to expect shared_ptr of type

caleridas:
Change the constructor of testnode to required shared_ptr<type> for input and output parameters. This is a large and purely mechanical change, but affects numerous files. This is the majority of node conversions required.


phate (approved):
LGTM

PR #507, opened 2024-06-08
Shared ptr6

caleridas:
This PR:

- converts constructors of rvsdg base nodes to only accept shared_ptr<type> instead of "const type &"
- adds further overloads to various llvm IR ops to accept shared_ptr<type> (for transition in addition to "const type&")
- removes the superseded "const type &" constructors from expport/impport and clean up all ripple effects

With this PR, the conversion to changed API in unit tests is nearly complete, so afterwards the "const type &"
API functions can be removed one-by-one.


jlm/rvsdg/nullary.hpp
```diff
@@ -25,8 +25,8 @@ class nullary_op : public simple_op
 public:
   virtual ~nullary_op() noexcept;
 
-  inline nullary_op(const jlm::rvsdg::port & result)
-      : simple_op({}, { result })
+  inline nullary_op(std::shared_ptr<const jlm::rvsdg::type> result)
```
phate, line 28: `explicit`

phate (approved):
LGTM. Only one minor comment.

jlm/rvsdg/nullary.hpp
caleridas: done

phate (approved):

PR #512, opened 2024-06-10
Add methods for retrieving memory states from lambda nodes [AndersenAgnostic][SteensgaardAgnostic]

phate:
This PR does the following:
1. Adds convenience methods to the lambda node for retrieving the memory state region argument or result.
2. Uses them to remove some FIXMEs in the memory state encoder.
3. Fixes an FreeNull RVSDG test where the memory state was not the last argument/result.

It is currently not enforced on lambda creation that the last argument/result of a lambda node needs to be a memory state. This is only enforced on the call nodes at the moment. The reason why we currently cannot enforce this is because the HLS dialect uses the same lambda, but without memory states. This tight coupling will hopefully be resolved in the near future and then we can start to enforce it on the lambda side as well.

### DIFF
--- a/jlm/llvm/ir/operators/lambda.cpp
+++ b/jlm/llvm/ir/operators/lambda.cpp
@@ -148,13 +148,28 @@ node::add_ctxvar(jlm::rvsdg::output * origin)
   return cvargument::create(subregion(), input);
 }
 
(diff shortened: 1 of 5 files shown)


haved (approved):
Nice cleanup

PR #513, opened 2024-06-10
Remove duplicate input_node() methods

phate:
We had three incarnations of the same method for retrieving the node connected to an input. Remove two, keep one.

### DIFF
--- a/jlm/llvm/opt/push.cpp
+++ b/jlm/llvm/opt/push.cpp
@@ -185,7 +185,7 @@ push(jlm::rvsdg::gamma_node * gamma)
       auto argument = region->argument(n);
       for (const auto & user : *argument)
       {
-        auto tmp = input_node(user);
+        auto tmp = jlm::rvsdg::input::GetNode(*user);
         if (tmp && tmp->depth() == 0)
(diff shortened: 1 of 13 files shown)


jlm/llvm/opt/push.cpp
```diff
@@ -185,7 +185,7 @@ push(jlm::rvsdg::gamma_node * gamma)
       auto argument = region->argument(n);
       for (const auto & user : *argument)
       {
-        auto tmp = input_node(user);
+        auto tmp = jlm::rvsdg::input::GetNode(*user);
```
haved, line 188: It seems only this file has the `jlm::` prefix included in all the replaced calls. Is there something special about this file or just a simple mistake?

haved (approved):
Looks good, just one comment

jlm/llvm/opt/push.cpp
phate: The functions are not in the jlm namespace (yet). This file has not been properly refactored yet. That is why the `jlm::` prefix was needed.

caleridas (approved):
Looks good

PR #514, opened 2024-06-11
Fix make target for Andersen with agnostic encoding [AndersenAgnostic]

phate:
We changed the make target for Andersen alias analysis with agnostic encoding in this [PR](https://github.com/phate/jlm-eval-suite/pull/31). Fix the CI to reflect this change.


phate:
The PR in jlm-eval-suite needs to be merged before this one can be merged.

haved (approved):

PR #515, opened 2024-06-12
Updates for compiling with LLVM/MLIR-17 [AndersenAgnostic][SteensgaardAgnostic]

sjalander:


configure.sh
```diff
@@ -141,6 +141,7 @@ MLIR_PATH=${MLIR_PATH}
 MLIR_LDFLAGS=${MLIR_LDFLAGS}
 LLVMCONFIG=${LLVM_CONFIG_BIN}
 ENABLE_COVERAGE=${ENABLE_COVERAGE}
+LD_LIBRARY_PATH=$(${LLVM_CONFIG_BIN} --libdir)
```
phate, line 144: Is this line really necessary? LD_LIBRARY_PATH was not defined before and is nowhere else used either now?

jlm/hls/backend/rhls2firrtl/RhlsToFirrtlConverter.cpp
```diff
@@ -4016,7 +4017,11 @@ RhlsToFirrtlConverter::WriteCircuitToFile(const circt::firrtl::CircuitOp circuit
   std::string fileName = name + extension();
   std::error_code EC;
   ::llvm::raw_fd_ostream output(fileName, EC);
-  auto status = circt::firrtl::exportFIRFile(module, output);
+  auto status = circt::firrtl::exportFIRFile(
+      module,
+      output,
+      100,
```
phate, line 4023: Could we assign this magic number to a variable with a descriptive name?

jlm/hls/backend/rhls2firrtl/RhlsToFirrtlConverter.cpp
```diff
@@ -4045,7 +4050,11 @@ RhlsToFirrtlConverter::toString(const circt::firrtl::CircuitOp circuit)
   // Export FIRRTL to string
   std::string outputString;
   ::llvm::raw_string_ostream output(outputString);
-  auto status = circt::firrtl::exportFIRFile(module, output);
+  auto status = circt::firrtl::exportFIRFile(
+      module,
+      output,
+      100,
```
phate, line 4056: same here

jlm/llvm/backend/jlm2llvm/jlm2llvm.cpp
```diff
@@ -264,7 +264,8 @@ convert_attribute_kind(const attribute::kind & kind)
         { attribute::kind::VScaleRange, ak::VScaleRange },
         { attribute::kind::LastIntAttr, ak::LastIntAttr },
 
-        { attribute::kind::EndAttrKinds, ak::EndAttrKinds } });
+        { attribute::kind::EndAttrKinds, ak::EndAttrKinds },
+        { attribute::kind::NoFPClass, ak::NoFPClass } });
```
phate, line 268: The NoFPClass attribute should not go last I believe. You see that all these attribute blocks have First/Last pairs. Check where it is located in the the LLVM attribute list

jlm/llvm/frontend/LlvmModuleConversion.cpp
```diff
@@ -174,7 +174,8 @@ ConvertAttributeKind(const ::llvm::Attribute::AttrKind & kind)
         { ak::UWTable, attribute::kind::UWTable },
         { ak::VScaleRange, attribute::kind::VScaleRange },
         { ak::LastIntAttr, attribute::kind::LastIntAttr },
-        { ak::EndAttrKinds, attribute::kind::EndAttrKinds } });
+        { ak::EndAttrKinds, attribute::kind::EndAttrKinds },
+        { ak::NoFPClass, attribute::kind::NoFPClass } });
```
phate, line 178: same here

jlm/llvm/ir/attribute.hpp
```diff
@@ -117,7 +117,8 @@ class attribute
     VScaleRange,
     LastIntAttr,
 
-    EndAttrKinds ///< Sentinel value useful for loops
+    EndAttrKinds, ///< Sentinel value useful for loops
+    NoFPClass
```
phate, line 121: I doubt it that this comes after the sentinel value for loops.

scripts/build-circt.sh
```diff
@@ -69,15 +69,16 @@ git checkout ${GIT_COMMIT}
...
 	-B ${CIRCT_BUILD_DIR} \
-	-DCMAKE_C_COMPILER=clang-16 \
-	-DCMAKE_CXX_COMPILER=clang++-16 \
+	-DCMAKE_C_COMPILER=clang-17 \
+	-DCMAKE_CXX_COMPILER=clang++-17 \
 	-DCMAKE_BUILD_TYPE=RelWithDebInfo \
-	-DLLVM_DIR=/usr/lib/llvm-16/cmake/ \
-	-DMLIR_DIR=/usr/lib/llvm-16/lib/cmake/mlir \
+	-DLLVM_DIR=/usr/lib/llvm-17/cmake/ \
+	-DMLIR_DIR=/usr/lib/llvm-17/lib/cmake/mlir \
```
phate, line 76: Should those absolute paths not come from `llvm-config`` At least the path prefix?

scripts/build-mlir.sh
```diff
@@ -61,10 +61,10 @@ git checkout ${GIT_COMMIT}
...
 	${MLIR_GIT_DIR} \
 	-B ${MLIR_BUILD_DIR} \
-	-DCMAKE_C_COMPILER=clang-16 \
-	-DCMAKE_CXX_COMPILER=clang++-16 \
-	-DLLVM_DIR=/usr/lib/llvm-16/cmake/ \
-	-DMLIR_DIR=/usr/lib/llvm-16/lib/cmake/mlir \
+	-DCMAKE_C_COMPILER=clang-17 \
+	-DCMAKE_CXX_COMPILER=clang++-17 \
+	-DLLVM_DIR=/usr/lib/llvm-17/cmake/ \
+	-DMLIR_DIR=/usr/lib/llvm-17/lib/cmake/mlir \
```
phate, line 67: Same here, no?

phate (changes requested):

phate:
@sjalander We would also need to update the LLVM test suite to the 17 version, no?

haved:
> @haved Do you want us to hold back on this PR until you are done with your thesis?

I'm all good, my results are done, and the code used to make them are in the branch masters-results. I will not prioritize merging the techniques into master before the thesis is finished.

configure.sh
sjalander: The libraries are for some reason not copied/linked in the system default path when installing the MLIR packages. So, if one doesn't manually fix this on ones own machine, then the MLIR unit tests fail.

jlm/llvm/backend/jlm2llvm/jlm2llvm.cpp
sjalander: Moved

jlm/llvm/ir/attribute.hpp
sjalander: Moved

scripts/build-circt.sh
sjalander: Using llvm-config

scripts/build-mlir.sh
sjalander: Using llvm-config

jlm/hls/backend/rhls2firrtl/RhlsToFirrtlConverter.cpp
sjalander: Adde variable

configure.sh
caleridas: This is very weird, you are taking the pre-compiled ones from apt repository? I would like to take a look, I think that should also be reported because it sounds like a packaging bug.

caleridas:
> @haved Do you want us to hold back on this PR until you are done with your thesis?
> 
> @caleridas Any opinion about it?

no issues from my side

jlm/hls/backend/rhls2firrtl/RhlsToFirrtlConverter.cpp
sjalander: Changed to size_t targetLineLength = 100;

jlm/hls/backend/rhls2firrtl/RhlsToFirrtlConverter.cpp
sjalander: Changed to size_t targetLineLength = 100;

jlm/llvm/frontend/LlvmModuleConversion.cpp
sjalander: Moved

.github/workflows/ClangFormat.yml
```diff
@@ -16,4 +16,4 @@ jobs:
       - name: "Find source files"
         run: find . -name "*.cpp" -or -name "*.hpp" > sourceFiles.txt
       - name: "Check format"
-        run: clang-format-16 --Werror --dry-run --style="file:.clang-format" --verbose --files=sourceFiles.txt
\ No newline at end of file
+        run: clang-format-17 --Werror --dry-run --style="file:.clang-format" --verbose --files=sourceFiles.txt
```
phate, line 19: Since you touch this line anyway, would it be possible to use the respective `format-dry-run` make target?

phate: See my review message for additional details.

jlm/llvm/ir/attribute.hpp
phate: This was not moved, it is still after the sentinel.

phate (changes requested):
1. In the Makefile.rules, lines 175 - 187 need also to be changed to support LLVM 17. It might be beneficial to add the clang-format command also the `configure.sh` script such that it is written to the `Makefile.config`. Moreover, it might be even better to introduce an LLVM_VERSION variable from which all the commands are derived in the Makefiles in the end, such that we only have to change it at once place.

configure.sh
sjalander: This is also the case for MLIR-16

sjalander:
@phate I checked the llvm-test-suite version that we are using, and it already contains the "release/17.x" branch.

.github/workflows/ClangFormat.yml
sjalander: Changed to format-dry-run

jlm/llvm/ir/attribute.hpp
sjalander: Sorry, should be fixed now

.github/actions/BuildJlm/action.yml
```diff
@@ -19,7 +19,7 @@ inputs:
 
   cxx:
     description: "Set C++ compiler. Default is 'clang++'."
-    default: "clang++-16"
+    default: "clang++-17"
```
phate, line 22: I think it is possible to just say `clang++` here, avoiding the version change on the next LLVM update.

configure.sh
phate: I will just resolve this conversation and move on here. I have not really the appetite to figure this out.

phate (changes requested):
I only added one minor issue, which would be nice to get fixed. However, this is not the reason why I hold this PR back.

I would like to get this [llvm-test-suite PR](https://github.com/phate/llvm-test-suite/pull/3) merged, and then I would like you to take the commit hash and use it in the CI instead of the old one from the jlm-with-llvm-16 branch. The new jlm-with-llvm-17 branch is based on the jlm-with-llvm-16 branch, and it should just compile (famous last words).

.github/actions/BuildJlm/action.yml
sjalander: Changed to clang++

phate (approved):

PR #516, opened 2024-06-13
Adds support for custom make targets

sjalander:
The user can write their own make targets in Makefile.custom, which is not tracked by git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 Makefile.config
+Makefile.custom
 .idea
 docs/html
 build
--- a/Makefile
+++ b/Makefile
@@ -31,3 +31,8 @@ include jlm/mlir/Makefile.sub
 endif
 
 include Makefile.rules
+
+# Provide support for custom make targets
+ifneq ("$(wildcard Makefile.custom)","")
+include Makefile.custom
+endif


phate:
I am reluctant on this one, but not against it. What is the concrete use case here?

@caleridas Do you have an opinion here?

caleridas:
> I am reluctant on this one, but not against it. What is the concrete use case here?
> 
> @caleridas Do you have an opinion here?

Would say if 1. It does not hurt anyone and 2. It helps someone, then it is not objectionable. And I think both preconditions are satisfied.

That said, I find it hard to imagine a good use case for this, it might be good to know what it is if there may be a better solution than interfering at the build system level.

phate:
Well, that is why I asked for the concrete use case at hand in order to clarify point 2 (it helps someone). The PR only adds this feature, but not the use case. I am reluctant in "supporting" this just for the sake of having the option without any concrete use for it.

phate:
@sjalander See comments above.

sjalander:
My use case looks like the following: 

---- START Makefile.custom ----

# LLVM related variables
CLANG_BIN_DIR=$(shell $(LLVMCONFIG) --bindir)
CLANG=$(CLANG_BIN_DIR)/clang

JLM_ROOT=$(PWD)

export PATH := $(JLM_ROOT)/build/:$(PATH)

POLYBENCH_ROOT = $(JLM_ROOT)/usr/polybench
ifneq ("$(wildcard $(POLYBENCH_ROOT)/Makefile.sub)","")
JLC=jlc
include $(POLYBENCH_ROOT)/Makefile.sub
endif

HLS_TEST_ROOT=$(JLM_ROOT)/usr/hls-test-suite
ifneq ("$(wildcard $(HLS_TEST_ROOT)/Makefile.sub)","")
JHLS=jhls
FIRTOOL=./usr/bin/firtool
include $(HLS_TEST_ROOT)/Makefile.sub
endif

LLVM_TEST_ROOT=$(JLM_ROOT)/usr/llvm-test-suite/jlm
ifneq ("$(wildcard $(LLVM_TEST_ROOT)/Makefile.sub)","")
include $(LLVM_TEST_ROOT)/Makefile.sub
endif

.PHONY: help
help:
	@$(HELP_TEXT_POLYBENCH)
	@$(HELP_TEXT_LLVM_TEST)
	@$(HELP_TEXT_HLS_TEST)`

---- END Makefile.custom ----

The custom Makefile makes it possible to interact with the different test suites that we have, other than by the run scripts (two of which are as PRs right now).

Each run script is currently for running one specific test that is used for the CI, while the custom make files make it possible to interact with the test suites at a finer granularity, e.g., only running a part of the hls-test-suite.
Such functionality could be added to the run scripts. But that would require typing a lot to specify different arguments to the various scripts that we have.

[Makefile.custom.patch](https://github.com/user-attachments/files/15846199/Makefile.custom.patch)

phate (approved):
All right, I am not a big fan of it, but it is not very invasive.

PR #519, opened 2024-06-13
Script for running the HLS test suite and updated CI

sjalander:


scripts/run-hls-test.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
+set -eu
+
+# URL to the benchmark git repository and the commit to be used
+GIT_REPOSITORY=https://github.com/EECS-NTNU/hls-test-suite.git
```
phate, line 5: The same issue as for the llvm-test-suite script. Please use: `https://github.com/phate/hls-test-suite.git`

scripts/run-hls-test.sh
```diff
@@ -0,0 +1,86 @@
...
+BENCHMARK_RUN_TARGET=run
+
+# We assume that the firtool is in the PATH
+FIRTOOL=firtool
+
+# Check if verilator exists
+if ! command -v verilator &> /dev/null
+then
+	echo "No verilator in ${PATH}" 
+	echo "Consider installing the verilator package for you Linux distro."
```
phate, line 24: "for you**r** Linux distro."

scripts/run-hls-test.sh
```diff
@@ -0,0 +1,86 @@
...
+done
+
+# Check if firtool exists
+if ! command -v ${FIRTOOL} &> /dev/null
+then
+	echo "${FIRTOOL} is not found."
+	echo "Make sure to use '--firtool COMMAND' to specify which firtool to use."
+	echo "You can use './scripts/build-circt.sh' to build it, if needed."
+	exit 1
+fi
```
phate, line 75: Move this check up in the script to the verilator check so we have it all in the same place.

phate (changes requested):

scripts/run-hls-test.sh
sjalander: Fixed

scripts/run-hls-test.sh
sjalander: FIRRTOOL is a user parameter, so the test must be below the argument parsing.
I moved the verilator check instead.

phate (approved):

PR #520, opened 2024-06-13
Script for running the llvm test suite and udpated CI [AndersenAgnostic][SteensgaardAgnostic]

sjalander:


scripts/run-llvm-test.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+set -eu
+
+# URL to the benchmark git repository and the commit to be used
+GIT_REPOSITORY=https://github.com/EECS-NTNU/llvm-test-suite.git
```
phate, line 5: I am not a big fan of having a dependency on an NTNU repository in the jlm CI where I do not have any rights. I created a clone of this repository, please add the clone: `https://github.com/phate/llvm-test-suite.git`

phate (changes requested):
A few more points:

1. Could you please append to the PR title "[AndersenAgnostic][SteensgaardAgnostic]" such that these two CI changes are tested. I do not seem to be allowed to do it.
2. The move to a new repository resulted in us losing the entire CI setup for the test suite repository. That is not great. However, considering that we are checking out a specific version here, the loss of CI might not be as important. We would need to update the commit here in order to get changes in and then would catch discrepancies between the suite and jlm.

phate:
@haved Do you have an opinion here?

scripts/run-llvm-test.sh
```diff
@@ -0,0 +1,67 @@
...
+GIT_COMMIT=45d49e6626838ab340a195d9a5cf04e7bba09af0
```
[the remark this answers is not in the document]
sjalander: Not great to have a comment about a different repository state that should be updated next time the GIT_COMMIT is changed. But I added it.

scripts/run-llvm-test.sh
```diff
@@ -0,0 +1,67 @@
...
+BENCHMARK_RUN_TARGET=llvm-run-opt
+
+function commit()
+{
+	echo ${GIT_COMMIT}
+}
+
+function usage()
+{
+	echo "Usage: ./run-llvm-test.sh [OPTION] [VAR=VALUE]"
```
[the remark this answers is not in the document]
sjalander: Renamed

scripts/run-llvm-test.sh
sjalander: You could have easily been added to the existing repo.
But I've changed the URL.

Having everything related to RVSDG and jlm under phate is maybe also not ideal, since you have 22 repositories so they kind of get swallowed. One option would be to create an organization and place everything related under it, and it looks like https://github.com/rvsdg/ is available.

sjalander:
@phate 
1. You could have edited the heading of the PR here on github to add [AndersenAgnostic][SteensgaardAgnostic].
2. Yes, it is possible to add commits to the llvm-test-suite that doesn't work (e.g., the llvm-team does this as they are not aware of jlm). But to make a change that affects jlm, one is required to use a commit that passes jlm's CI. So I don't see this as a problem, and we avoid the circular dependency that we have with jlm-eval-suite.

scripts/run-llvm-test-suite.sh
```diff
@@ -0,0 +1,68 @@
...
+
+if [ ! -d "$BENCHMARK_DIR" ] ;
+then
+	git clone ${GIT_REPOSITORY} ${BENCHMARK_DIR}
+fi
+
+export PATH=${JLM_BIN_DIR}:${PATH}
+cd ${BENCHMARK_DIR}
+git checkout ${GIT_COMMIT}
+cd jlm
```
haved, line 66: Does this actually use the jlm compiled from the pull request branch, or does it use the jlm submodule within jlm-eval-suite, which is pinned to some other commit?

scripts/run-llvm-test-suite.sh
```diff
@@ -0,0 +1,68 @@
...
+make clean
+make ${BENCHMARK_RUN_TARGET}
```
haved, line 68: This should not run within the jlm folder, but the jlm-eval-suite root, right?

scripts/run-llvm-test-suite.sh
haved: Also `make submodule` and `make apply-llvm-git-patch` is gone? Am I missing something?

scripts/run-llvm-test.sh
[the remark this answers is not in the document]
haved: Was the previous version deleted from the branch? It looks like both have been added to me

scripts/run-llvm-test-suite.sh
sjalander: jlm-eval-suite is no longer used.

Instead, a jlm specific version of the llvm-test-suite repo has been created, which the new script clones and checks out a commit that already has the patch applied.

scripts/run-llvm-test-suite.sh
sjalander: If one runs the script without providing any arguments, then it ends up in a usr-subdirectory of the jlm root.

Again, this is to avoid the need for jlm-eval-suite.

scripts/run-llvm-test-suite.sh
haved: Ah, now I see. I will have a look at that repo

scripts/run-llvm-test.sh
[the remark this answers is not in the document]
phate: You definitely got a point. I think you are right, the comment will be forgotten to be updated. Your choice, either reverse this or leave it as it is. I am fine either way but tend now towards not having it actually.

phate (changes requested):
You have now the `run-llvm-test.sh` and `run-llvm-test-suite.sh` in this PR.

scripts/run-llvm-test.sh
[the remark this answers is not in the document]
sjalander: So I removed the comment.

sjalander:
run-llvm-test.sh has been removed.

haved:
@phate the conditional triggering of jobs based on PR title does not detect changed titles. I think you can manually trigger them somehow, otherwise I have often had something else to commit, triggering a new check

scripts/run-llvm-test.sh
phate: I will have a look at github organizations.

phate:
@havedIs everything clear now? Can we merge this PR in your opinion?

phate (approved):

haved (approved):
Looks good to me!

haved:
@sjalander all good to merge

PR #521, opened 2024-06-16
Convert remaining llvm operators

caleridas:
Convert all remaining llvm operators to operator on std::shared_ptr<type>, remove any method that requires copying types.


caleridas:
This is almost the last part. Only the following work remains:
- final cleanup/conversion of HLS parts & remove type::copy method
- API cleanup (remove residual methods, systematically make everything based on ptrs, consistent naming across different parts of the code, some consolidation of helper functions etc.)

phate (approved):
LGTM

PR #522, opened 2024-06-16
Finalize type shared_ptr support

caleridas:
Remove type copying from all remaining places, convert HLS ops. Remove
the "copy" member function from all type classes. This concludes the transition
to use shared references for types, although some API cleanup is stilll desired.

### DIFF
--- a/jlm/hls/ir/hls.cpp
+++ b/jlm/hls/ir/hls.cpp
@@ -8,6 +8,13 @@
 namespace jlm::hls
 {
 
+std::shared_ptr<const triggertype>
+triggertype::Create()
+{
+  static const triggertype instance;
+  return std::shared_ptr<const triggertype>(std::shared_ptr<void>(), &instance);
+}
+
 jlm::rvsdg::structural_output *
 loop_node::add_loopvar(jlm::rvsdg::output * origin, jlm::rvsdg::output ** buffer)
 {
@@ -15,7 +22,7 @@ loop_node::add_loopvar(jlm::rvsdg::output * origin, jlm::rvsdg::output ** buffer
   auto output = jlm::rvsdg::structural_output::create(this, origin->Type());
 
   auto argument_in = jlm::rvsdg::argument::create(subregion(), input, origin->Type());
-  auto argument_loop = add_backedge(origin->type());
+  auto argument_loop = add_backedge(origin->Type());
 
   auto mux =
       hls::mux_op::create(*predicate_buffer(), { argument_in, argument_loop }, false, true)[0];
@@ -69,7 +76,7 @@ loop_node::copy(jlm::rvsdg::region * region, jlm::rvsdg::substitution_map & smap
     auto arg = subregion()->argument(i);
     if (auto ba = dynamic_cast<backedge_argument *>(arg))
     {
-      auto na = loop->add_backedge(arg->type());
+      auto na = loop->add_backedge(arg->Type());
       smap.insert(ba, na);
     }
   }
@@ -98,9 +105,9 @@ loop_node::copy(jlm::rvsdg::region * region, jlm::rvsdg::substitution_map & smap
 }
 
 backedge_argument *
-loop_node::add_backedge(const jlm::rvsdg::type & type)
+loop_node::add_backedge(std::shared_ptr<const jlm::rvsdg::type> type)
 {
-  auto argument_loop = backedge_argument::create(subregion(), type);
+  auto argument_loop = backedge_argument::create(subregion(), std::move(type));
   auto result_loop = backedge_result::create(argument_loop);
   argument_loop->result_ = result_loop;
   result_loop->argument_ = argument_loop;
@@ -114,7 +121,7 @@ loop_node::create(jlm::rvsdg::region * parent, bool init)
   if (init)
   {
     auto predicate = jlm::rvsdg::control_false(ln->subregion());
-    auto pred_arg = ln->add_backedge(jlm::rvsdg::ctltype(2));
+    auto pred_arg = ln->add_backedge(jlm::rvsdg::ctltype::Create(2));
     pred_arg->result()->divert_to(predicate);
     // we need a buffer without pass-through behavior to avoid a combinatorial cycle of ready
     // signals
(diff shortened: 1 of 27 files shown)


jlm/hls/ir/hls.hpp
```diff
@@ -226,8 +230,8 @@ class sink_op final : public jlm::rvsdg::simple_op
   virtual ~sink_op()
   {}
 
-  sink_op(const jlm::rvsdg::type & type)
-      : jlm::rvsdg::simple_op({ type.copy() }, {})
+  sink_op(const std::shared_ptr<const jlm::rvsdg::type> & type)
```
phate, line 233: `explicit`

jlm/hls/ir/hls.hpp
```diff
@@ -264,8 +268,8 @@ class predicate_buffer_op final : public jlm::rvsdg::simple_op
   virtual ~predicate_buffer_op()
   {}
 
-  predicate_buffer_op(const jlm::rvsdg::ctltype & type)
-      : jlm::rvsdg::simple_op({ type.copy() }, { type.copy() })
+  predicate_buffer_op(const std::shared_ptr<const jlm::rvsdg::ctltype> & type)
```
phate, line 271: `explicit`

jlm/hls/ir/hls.hpp
```diff
@@ -410,25 +419,20 @@ class triggertype final : public jlm::rvsdg::statetype
...
-
 class trigger_op final : public jlm::rvsdg::simple_op
 {
 public:
   virtual ~trigger_op()
   {}
 
-  trigger_op(const jlm::rvsdg::type & type)
-      : jlm::rvsdg::simple_op({ trigger.copy(), type.copy() }, { type.copy() })
+  trigger_op(const std::shared_ptr<const jlm::rvsdg::type> & type)
```
phate, line 434: `explicit`

jlm/hls/ir/hls.hpp
```diff
@@ -473,8 +477,8 @@ class print_op final : public jlm::rvsdg::simple_op
   virtual ~print_op()
   {}
 
-  print_op(const jlm::rvsdg::type & type)
-      : jlm::rvsdg::simple_op({ type.copy() }, { type.copy() })
+  print_op(const std::shared_ptr<const jlm::rvsdg::type> & type)
```
phate, line 480: `explicit`

jlm/hls/ir/hls.hpp
```diff
@@ -1093,39 +1100,26 @@ class mem_req_op final : public jlm::rvsdg::simple_op
...
     LoadTypes_ = new std::vector<std::shared_ptr<const rvsdg::type>>();
     StoreTypes_ = new std::vector<std::shared_ptr<const rvsdg::type>>();
     for (auto loadType : load_types)
     {
       JLM_ASSERT(
-          dynamic_cast<const rvsdg::bittype *>(loadType)
-          || dynamic_cast<const llvm::PointerType *>(loadType));
-      LoadTypes_->emplace_back(loadType->copy());
+          dynamic_cast<const rvsdg::bittype *>(loadType.get())
+          || dynamic_cast<const llvm::PointerType *>(loadType.get()));
```
phate, line 1113: `JLM_ASSERT(is<rvsdg::bittype*>(loadType.get()) || is<llvm::PointerType>(loadedType.get()));`

jlm/hls/ir/hls.hpp
```diff
@@ -1320,8 +1316,8 @@ class local_mem_op final : public jlm::rvsdg::simple_op
   virtual ~local_mem_op()
   {}
 
-  local_mem_op(const llvm::arraytype & at)
-      : simple_op({}, CreateOutTypes(at))
+  local_mem_op(std::shared_ptr<const llvm::arraytype> at)
```
phate, line 1319: `explicit`

phate (changes requested):

jlm/hls/ir/hls.hpp
caleridas: that does not compile -- I cannot find definition for something named "is"

caleridas: oh right now I remember that I even touched it -- fixed

jlm/hls/ir/hls.cpp
```diff
@@ -135,28 +142,26 @@ loop_node::set_predicate(jlm::rvsdg::output * p)
...
 {
-  auto elements =
-      new std::vector<std::pair<std::string, std::shared_ptr<const jlm::rvsdg::type>>>();
-  elements->emplace_back("addr", llvm::PointerType::Create());
-  elements->emplace_back("size", std::make_unique<jlm::rvsdg::bittype>(4));
-  elements->emplace_back("id", std::make_unique<jlm::rvsdg::bittype>(8));
+  std::vector<std::pair<std::string, std::shared_ptr<const jlm::rvsdg::type>>> elements;
+  elements.emplace_back("addr", llvm::PointerType::Create());
+  elements.emplace_back("size", std::make_unique<jlm::rvsdg::bittype>(4));
+  elements.emplace_back("id", std::make_unique<jlm::rvsdg::bittype>(8));
```
phate, line 150: Should this not be `rvsdg::bittype::Create(4)` and `rvsdg::bittype::Create(8)`, respectively?

jlm/hls/ir/hls.hpp
```diff
@@ -455,11 +459,11 @@ class trigger_op final : public jlm::rvsdg::simple_op
   static std::vector<jlm::rvsdg::output *>
   create(jlm::rvsdg::output & tg, jlm::rvsdg::output & value)
   {
-    if (tg.type() != trigger)
+    if (tg.type() != *triggertype::Create())
```
phate, line 462: This can be replaced by `!rvsdg::is<triggertype>(tg.type())`, no?

jlm/hls/ir/hls.hpp
```diff
@@ -814,10 +816,10 @@ class load_op final : public jlm::rvsdg::simple_op
     return *util::AssertedCast<const llvm::PointerType>(&argument(0).type());
   }
 
-  [[nodiscard]] const rvsdg::valuetype &
+  [[nodiscard]] std::shared_ptr<const rvsdg::valuetype>
   GetLoadedType() const noexcept
   {
-    return *util::AssertedCast<const rvsdg::valuetype>(&result(0).type());
+    return std::dynamic_pointer_cast<const rvsdg::valuetype>(result(0).Type());
```
phate, line 822: I actually do not like it that you replace `AssertedCast()` with `dynamic_pointer_cast()`. The problem is that we do not have an `AssertedCast()` variant that takes a `shared_ptr`. Any chance we could introduce it and use it?

jlm/hls/ir/hls.hpp
```diff
@@ -1006,10 +1011,10 @@ class decoupled_load_op final : public jlm::rvsdg::simple_op
...
```
phate, line 1017: same here

jlm/hls/ir/hls.hpp
```diff
@@ -1019,7 +1024,7 @@ class mem_resp_op final : public jlm::rvsdg::simple_op
   virtual ~mem_resp_op()
   {}
 
-  mem_resp_op(const std::vector<const rvsdg::valuetype *> & output_types)
+  mem_resp_op(const std::vector<std::shared_ptr<const rvsdg::valuetype>> & output_types)
```
phate, line 1027: `explicit`

jlm/hls/ir/hls.hpp
```diff
@@ -1087,45 +1094,25 @@ class mem_req_op final : public jlm::rvsdg::simple_op
 {
 public:
   virtual ~mem_req_op()
-  {
-    delete LoadTypes_;
-    delete StoreTypes_;
-  }
+  {}
```
phate, line 1097: `~mem_req_op() override = default;`?

jlm/hls/ir/hls.hpp
```diff
@@ -1470,19 +1459,19 @@ class local_load_op final : public jlm::rvsdg::simple_op
...
     inputs.push_back(&load_result);
     return jlm::rvsdg::simple_node::create_normalized(region, op, inputs);
   }
 
-  [[nodiscard]] const rvsdg::valuetype &
+  [[nodiscard]] std::shared_ptr<const rvsdg::valuetype>
   GetLoadedType() const noexcept
   {
-    return *util::AssertedCast<const rvsdg::valuetype>(&result(0).type());
+    return std::dynamic_pointer_cast<const rvsdg::valuetype>(result(0).Type());
```
phate, line 1474: same here

phate (changes requested):
I am sorry, but I found some more minor issues.

jlm/hls/ir/hls.cpp
caleridas: good catch! I find several of these, but missed this one

jlm/hls/ir/hls.hpp
caleridas: I think we should rather get rid of the cast here entirely and store this thing as shared_ptr<const valuetype> to begin with. This requires a bit more, particularly "simple_node" is a bit too simplistic in that it squashes down everything to shared_ptr<type> from which you have to upcast again. I would like to refactor the API such that the respective classes store the "correct" types themselves and should overload a ResultType(index) instead to expose the generic form.

jlm/hls/ir/hls.hpp
caleridas: yes, and generally definitions for any virtual methods should not be in header -- it slows down compilation & linking (and causes version coupling issues for shared libraries, but that's a bit premature to be concerned about this).

jlm/hls/ir/hls.hpp
phate: You resolved the conversation, but the explicit is not there.

phate (approved):

jlm/hls/ir/hls.hpp
caleridas: sorry, forgotten cherry-pick on branch, resolved now

phate (approved):

PR #524, opened 2024-06-20
Adding default MLIR and CIRCT paths

sjalander:
This update makes it possible to enable the HLS and MLIR backend by simply providing --enable-hls and --enable-mlir without having to specify any paths if CIRCT and MLIR have been built using the scripts provided by jlm with default arguments.

./configure.sh --enable-mlir --enable-hls
instead of
./configure.sh --enable-mlir ./usr --enable-hls ./usr

The option to specify a different path is still available for the case where CIRCT or MLIR are located somewhere else.


phate (approved):

sjalander:
This is not working as intended.
Closing it for now while considering better options.

PR #528, opened 2024-06-27
Implement equality operator for attributeset class

### LINKED ISSUE #525: Missing lambda attributeset `==` operator
As discussed in #492, I'm working on implementing a graph equivalence function in jlm.

The issue is that the lambda `==` operator calls this attributeset `==` operator which is not implemented

In `jlm/llvm/ir/attribute.hpp` :
```c++
class attributeset final

...

bool
  operator==(const attributeset & other) const noexcept
  {
    /*
      FIXME: Ah, since this is not a real set, we cannot cheaply implement a comparison.
    */
    return false;
  }
```
Also, if I want to implement it myself there need to be a choice on the attributeset implementation here which has not been addressed.

```c++
class attributeset final

...

private:
  /*
    FIXME: Implement a proper set. Elements are not unique here.
  */
  std::vector<std::unique_ptr<attribute>> attributes_;
```

@phate @sjalander

phate:
This PR reimplements the internals of the attributeset class such that we can properly support `operator==()`.

Closes #525


jlm/llvm/ir/attribute.hpp
```diff
@@ -267,181 +220,148 @@ class int_attribute final : public enum_attribute
...
+{
+  std::size_t
+  operator()(const jlm::llvm::type_attribute & attribute) const noexcept
   {
-    /*
-      FIXME: Ah, since this is not a real set, we cannot cheaply implement a comparison.
-    */
-    return false;
+    // FIXME: I have no better idea of how to handle this currently
+    return std::hash<ptrdiff_t>()(reinterpret_cast<ptrdiff_t>(&attribute));
```
phate, line 287: @haved @caleridas Any better idea?

phate:
FYI @ProgrammingLouis

jlm/llvm/ir/attribute.hpp
haved: This might be problematic as two `type_attribute` can compare equal with `operator==` yet have different hashes.

Would hashing the kind and type and xoring them together work?

haved: @phate I checked the definition of std::hash, and they require == equality to imply identical hashes

phate: Yeah, this won't work properly unless we implement hashing capabilities for `rvsdg::type`. I will take a stab at it in a new PR, then we can merge this one.

phate: Fixed.

phate:
@caleridas @haved I finally found some time to fix this. The changes since last review:
1. Fixed hashing of types (and improved the combining of the other hashes)
2. Added a unit test for attribute set (in-)equality

FYI @ProgrammingLouis

phate:
@haved @caleridas Anyone, please?

haved (dismissed):
Looks good, I only have a question about one of the casts.

Also in your PR description you say `operator=()`, I assume you mean `operator==()`

jlm/llvm/backend/jlm2llvm/jlm2llvm.cpp
```diff
@@ -271,42 +271,54 @@ convert_attribute_kind(const attribute::kind & kind)
...
-static ::llvm::AttributeSet
-convert_attributes(const attributeset & as, context & ctx)
+static ::llvm::Attribute
+ConvertEnumAttribute(const llvm::enum_attribute & attribute, context & ctx)
 {
-  auto convert_attribute = [](const llvm::attribute & attribute, context & ctx)
-  {
-    auto & llvmctx = ctx.llvm_module().getContext();
+  auto & llvmContext = ctx.llvm_module().getContext();
+  auto ea = dynamic_cast<const enum_attribute *>(&attribute);
```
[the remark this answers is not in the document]
phate: Ah, that is nonsense. I will fix it.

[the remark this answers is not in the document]
phate: Removed.

phate:
@haved I removed the `dynamic_cast<>()` and fixed the description. I need a new review though.

haved (approved):

PR #535, opened 2024-07-02
[WIP] Add hashing for types

phate:
What is still missing:
1. Provide documentation
2. Unit tests

### DIFF
--- a/jlm/hls/ir/hls.cpp
+++ b/jlm/hls/ir/hls.cpp
@@ -4,15 +4,40 @@
  */
 
 #include <jlm/hls/ir/hls.hpp>
+#include <jlm/util/HashSet.hpp>
 
 namespace jlm::hls
 {
 
+std::size_t
+triggertype::ComputeHash() const noexcept
+{
+  return std::hash<const triggertype *>()(GetInstance());
+}
+
 std::shared_ptr<const triggertype>
 triggertype::Create()
 {
+  return std::shared_ptr<const triggertype>(std::shared_ptr<void>(), GetInstance());
+}
+
+const triggertype *
+triggertype::GetInstance() noexcept
+{
   static const triggertype instance;
-  return std::shared_ptr<const triggertype>(std::shared_ptr<void>(), &instance);
+  return &instance;
+}
+
+std::size_t
+bundletype::ComputeHash() const noexcept
+{
+  std::size_t seed = 0;
+  for (auto & element : elements_)
+  {
+    util::CombineHash(seed, element.first, element.second->ComputeHash());
+  }
+
+  return seed;
 }
 
 jlm::rvsdg::structural_output *
--- a/jlm/util/HashSet.hpp
+++ b/jlm/util/HashSet.hpp
@@ -31,9 +31,18 @@ struct Hash<std::pair<First, Second>>
   }
 };
 
+// FIXME: add documentation
+template<typename T, typename... Args>
+void
+CombineHash(std::size_t & seed, const T & arg, Args... args)
+{
+  seed ^= std::hash<T>()(arg) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+  (CombineHash(seed, args), ...);
+}
+
 /**
- * Represents a set of values. A set is a collection that contains no duplicate elements, and whose
- * elements are in no particular order.
+ * Represents a set of values. A set is a collection that contains no duplicate elements, and
+ * whose elements are in no particular order.
  * @tparam ItemType The type of the items in the hash set.
  */
 template<typename ItemType, typename HashFunctor = Hash<ItemType>>
--- a/jlm/hls/ir/hls.hpp
+++ b/jlm/hls/ir/hls.hpp
@@ -419,10 +419,15 @@ public:
     return type;
   };
 
+  [[nodiscard]] std::size_t
+  ComputeHash() const noexcept override;
+
   static std::shared_ptr<const triggertype>
   Create();
 
 private:
+  static const triggertype *
+  GetInstance() noexcept;
 };
 
 class trigger_op final : public jlm::rvsdg::simple_op
@@ -704,6 +709,9 @@ public:
     }
     return true;
   };
+
+  [[nodiscard]] std::size_t
+  ComputeHash() const noexcept override;
 
   std::shared_ptr<const jlm::rvsdg::type>
   get_element_type(std::string element) const
--- a/jlm/llvm/ir/types.cpp
+++ b/jlm/llvm/ir/types.cpp
@@ -4,6 +4,7 @@
  */
 
 #include <jlm/llvm/ir/types.hpp>
+#include <jlm/util/HashSet.hpp>
 #include <jlm/util/strfmt.hpp>
 
 #include <unordered_map>
@@ -78,6 +79,22 @@ FunctionType::operator==(const jlm::rvsdg::type & _other) const noexcept
   return true;
 }
 
+std::size_t
+FunctionType::ComputeHash() const noexcept
+{
+  std::size_t seed = 0;
+  for (auto argumentType : ArgumentTypes_)
+  {
+    util::CombineHash(seed, argumentType->ComputeHash());
+  }
+  for (auto resultType : ResultTypes_)
+  {
+    util::CombineHash(seed, resultType->ComputeHash());
+  }
+
+  return seed;
+}
+
 FunctionType &
 FunctionType::operator=(const FunctionType & rhs) = default;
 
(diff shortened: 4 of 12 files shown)


phate:
@haved @caleridas I am not finished with it yet, but could I get a review to see whether this goes in the right direction, please?

jlm/hls/ir/hls.cpp
```diff
@@ -4,15 +4,40 @@
...
 #include <jlm/hls/ir/hls.hpp>
+#include <jlm/util/HashSet.hpp>
 
 namespace jlm::hls
 {
 
+std::size_t
+triggertype::ComputeHash() const noexcept
+{
+  return std::hash<const triggertype *>()(GetInstance());
```
caleridas, line 15: I would suggest not hashing on some address, but simply using a constant magic value

Something like this would also work (it gets compile-time evaluated):

`constexpr std::size_t const_hash_str(std::string_view s) {
	std::size_t v = 0;
	for (auto c : s) {
		v = v * 5 + c;
	}
	return v;
}

std::size_t triggertype::ComputeHash() const noexcept { return constr_str_hash("triggertype"); }
`

phate: I am fine with either, but why not hashing on a constant address? What is the rationale?

caleridas: a) determinism b) compile-time constants are always nice I would say

haved (commented):
I have no comments beyond things I assume are already covered by //FIXMEs, or just me marveling at C++ features. 

I had to read up on shared pointers and fold expressions again to understand some of it, but it makes sense.

The only slight annoyance is that sometimes values are passed to `CombineHash`, and other times a hash is passed. In the latter case you end up hashing a hash, which isn't really a problem, but uses some extra CPU and might increase the chance of hash collisions by an extremely small margin.

jlm/util/HashSet.hpp
```diff
@@ -31,9 +31,18 @@ struct Hash<std::pair<First, Second>>
   }
 };
 
+// FIXME: add documentation
+template<typename T, typename... Args>
+void
+CombineHash(std::size_t & seed, const T & arg, Args... args)
+{
+  seed ^= std::hash<T>()(arg) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
```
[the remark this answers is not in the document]
phate: Ultimately from boost (at some pointer).
Solution stolen from [here](https://stackoverflow.com/questions/2590677/how-do-i-combine-hash-values-in-c0x)
Explanation can be found [here](https://stackoverflow.com/questions/35985960/c-why-is-boosthash-combine-the-best-way-to-combine-hash-values)
I generally do not have a strong opinion here, but need something that just works without me re-inventing the wheel.

jlm/hls/ir/hls.cpp
phate: Computing the hash at compile-time now where possible. I did not find a way to automatically get the qualified name of the type (`typeid(triggertype).name` did not do the trick), so I sadly had to provide manually the string.

phate:
@haved I separated hash computation and combination from each other. This way we avoid taking the hash of a hash value.

jlm/util/Hash.hpp
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/**
+ * Computes a hash value from string \p s at compile-time.
+ *
+ * @param s The string from which to compute the hash value.
+ * @return A hash value of \p s.
+ */
+constexpr std::size_t
+ComputeConstantHash(const std::string_view & s)
```
caleridas, line 77: just pass std::string_view as value (no const &)

jlm/util/Hash.hpp
```diff
@@ -0,0 +1,90 @@
...
+#include <string_view>
+
+namespace jlm::util
+{
+
+/**
+ * Our own version of std::hash that also supports hashing std::pair
+ */
+template<typename T>
+struct Hash : std::hash<T>
```
caleridas, line 19: TBH I am never sure which std:: classes you are allowed to derive from and which not. With rare exceptions, things in std:: should be treated as final. I am unclear about std::hash. Is hash-of-pair a recurring thing?

caleridas (approved):
Nothing much to say, does what it states that it does. Maybe name "ComputeHash" is a bit verbose... but does not really matter.

One thing that is going to be a footgun: `std::unordered_map<std::shared_ptr<const type>, ...>` might be something desirable to write, it even compiles and mostly works... except when it will not. I had considered introducing a wrapper type for std::shared_ptr<const type> that does not suffer from this problem, but I also do not like too many wrapping abstractions... not sure what is best.

jlm/util/Hash.hpp
haved: There is at least one place where HashSet of pair comes up, and I would think it is a common enough use case that it comes up again.

I'm noticing that the implementation currently does not support pairs of pairs, due to not using `util::Hash` in the implementation.

Regarding not inheriting from std::hash, I tried looking for information about it being illegal or discouraged. Since there is only one method, operator(), we could manually implement it and call `std::hash<T>()(value)` from there.

jlm/util/Hash.hpp
```diff
@@ -0,0 +1,90 @@
...
+{
+};
+
+template<typename First, typename Second>
+struct Hash<std::pair<First, Second>>
+{
+  std::size_t
+  operator()(const std::pair<First, Second> & value) const noexcept
+  {
+    return std::hash<First>()(value.first) ^ std::hash<Second>()(value.second) << 1;
```
haved, line 29: If you want to you can replace `std::hash` with `Hash` to enable hashing pairs of pairs as well.

haved (approved):
Looks good, just one comment where I realize my original implementation could be better

jlm/util/Hash.hpp
phate: I removed the function again.

phate:
@caleridas @haved I need a review again. The following things changed since last review:
1. I added unit tests for the `ComputeHash()` method for all type subclasses.
2. I realized that I need to add a subtype specific component to the hash, otherwise unrelated subtypes with the same attributes would produce the same hash. For example, `bittype` and `ctltype` might produce the same hash if no type component is added as both have only a size attribute (i.e., the number of bits and the number of alternatives) and if the value of both attributes was the same, then the hash was the same (for obvious reasons). The same problem occurred for `fixedvectortype`, `scalablevectortype`, and `arraytype`.
3. I realized that the C++ standard library provides [builtin hashes for types](https://en.cppreference.com/w/cpp/types/type_info/hash_code). I used this as the type specific component of the hash, avoiding all the home-grown strings. This in turn rendered the `ComputeConstantHash()` method superfluous and I removed it again. The problem is that the `hash_code()` is not computed at compilation time any longer. It seems to be the more reliable approach, but might not be as great in terms of determinism (quickly checking the result of `hash_code()` between program runs seems to result in the same hash value for `g++` and `clang++` though, but it does not seem to be guaranteed) and performance.
4. Changed `std::hash` to `Hash` for `std::pairs`.

tests/jlm/hls/ir/TriggerTypeTests.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+  // Arrange
+  triggertype triggerType1;
+  triggertype triggerType2;
+  jlm::llvm::iostatetype ioStateType;
+
+  // Act & Assert
+  assert(triggerType1.ComputeHash() == triggerType1.ComputeHash());
+  assert(triggerType2.ComputeHash() == triggerType2.ComputeHash());
+  assert(triggerType1.ComputeHash() != ioStateType.ComputeHash());
```
haved, line 25: It seems a bit off to assert that the hashes are strictly not the same, as hash collisions are possible. Since the hashes are based on `typeid().hash_code()` it might be the case that hash collisions are not possible (the standard recommends that implementations avoid this as much as possible)

haved (approved):
Looks good to me, my only concern is that the tests assert that hash collisions do not happen. With 64 bit hashes and using `typeid().hash_code()`, it should not be a problem, but the C++ standard does not strictly require different types to have different hash codes. In practice I think it is fine, as the asserts are only used in tests.

tests/jlm/hls/ir/TriggerTypeTests.cpp
phate: I agree with you, but I have no idea how to better check it. There is a chance for hash collisions, but it should hopefully be very unlikely. Before I introduced the type aspect to the hash for each implementation, we had guaranteed collisions for hashes from different types (which should be very unlikely and not guaranteed). These checks ensure that we will hopefully not introduce a bug again such that it becomes guaranteed.

Honestly, I have no idea of how to improve upon this. @caleridas Any idea how to improve from here (if it is even worth it).

phate: @haved @caleridas  I removed all the unit tests with the too strict asserts again.

caleridas (approved):
looks good

haved (approved):
I agree, looks good

PR #541, opened 2024-07-09
Test CIRCT cache

sjalander:


.github/workflows/hls.yml
```diff
@@ -6,8 +6,53 @@ on:
...
+      id: color-selector
+      run: echo "SELECTED_COLOR=green" >> "$GITHUB_OUTPUT"
+    - name: Get color
+      env:
+        SELECTED_COLOR: ${{ steps.color-selector.outputs.SELECTED_COLOR }}
+      run: echo "The selected color is $SELECTED_COLOR"
+    - name: "Get the commit used for building CIRCT and use it as the cache key"
+      id: get-circt-hash
+      run: |
+        echo "GIT_HASH=$(./scripts/build-circt.sh --get-commit-hash)" >> "$GITHUB_OUTPUT"
```
phate, line 29: The difference to the `SELECTED_COLOR=green` above is that you do not just assign a string statically here. You are trying to execute something. I suspect something silently fails there. What happens if you just do statically `"GIT_HASH=<copied-hash-key-from-script>"` and then try to run it again? If it works, then you know that something is wrong with `$(./scripts/build-circt.sh --get-commit-hash)`

phate (commented):
Maybe it would be better to use a two-step process as shown here: https://nitinsapru.hashnode.dev/using-githuboutput-to-write-output-parameters-from-github-action-jobs-steps

PR #542, opened 2024-07-10
Build scripts fix for --get-commit-hash

sjalander:
Changed --get-commit-hash to use 'exit 0' and print to stdout instead of stderr.


sjalander:
@phate It took time to realize that the build scripts print to stderr instead of stdout. This is a copy and paste mistake from how the --help option works, which I guess one want to make sure people see and not intended to continue afterward.
In the log it looked like the hash was the correct as one could read it, but in reality we shouldn't have seen any printout.

phate (approved):

PR #544, opened 2024-07-14
Removes --firtool option from run-hls-test.sh

sjalander:
With jlm-hls being able to generate Verilog there is no longer a need to use the firtool to convert FIRRTL to Verilog.


sjalander:
Premature as changes in the hls-test-suite is needed first.
Will close it for now and open it once changes to the hls-test-suite has been merged.

PR #545, opened 2024-07-14
jhls links llvm-IR files and creates single object file

sjalander:
This simplifies how HLS benchmarks are compiled and reduce their dependencies to only jlm binaries (jhls and jlm-hls) and Verilator.

### DIFF
--- a/jlm/tooling/CommandGraphGenerator.cpp
+++ b/jlm/tooling/CommandGraphGenerator.cpp
@@ -360,20 +360,15 @@ JhlsCommandGraphGenerator::GenerateCommandGraph(const JhlsCommandLineOptions & c
       commandLineOptions.UseCirct_);
   m2r2.AddEdge(hls);
 
-  auto inputFiles = util::filepath(commandLineOptions.OutputFile_.to_str() + ".re*.ll");
-  std::vector<util::filepath> linkerInputFiles;
-  linkerInputFiles.push_back(inputFiles);
-  auto & llvmLink = LlvmLinkCommand::Create(
-      *commandGraph,
-      linkerInputFiles,
-      util::filepath(commandLineOptions.OutputFile_.to_str() + ".merged.ll"),
-      true,
-      false);
+  auto linkerInputFiles = util::filepath(commandLineOptions.OutputFile_.to_str() + ".re*.ll");
+  auto mergedFile = util::filepath(commandLineOptions.OutputFile_.to_str() + ".merged.ll");
+  auto & llvmLink =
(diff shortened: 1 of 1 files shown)


jlm/tooling/CommandGraphGenerator.cpp
```diff
@@ -360,12 +360,24 @@ JhlsCommandGraphGenerator::GenerateCommandGraph(const JhlsCommandLineOptions & c
...
-    if (c.RequiresLinking())
-      lnkifiles.push_back(c.OutputFile());
-  }
+  auto inputFiles = util::filepath(commandLineOptions.OutputFile_.to_str() + ".re*.ll");
+  std::vector<util::filepath> linkerInputFiles;
+  linkerInputFiles.push_back(inputFiles);
+  auto & llvmLink = LlvmLinkCommand::Create(
+      *commandGraph,
+      linkerInputFiles,
+      util::filepath(commandLineOptions.OutputFile_.to_str() + ".merged.ll"),
```
phate, line 369: Assign this file path to a variable and use the variable at all use sides instead. This file path is currently constructed multiple times.

jlm/tooling/CommandGraphGenerator.cpp
```diff
@@ -360,12 +360,24 @@ JhlsCommandGraphGenerator::GenerateCommandGraph(const JhlsCommandLineOptions & c
...
-  {
-    if (c.RequiresLinking())
-      lnkifiles.push_back(c.OutputFile());
-  }
+  auto inputFiles = util::filepath(commandLineOptions.OutputFile_.to_str() + ".re*.ll");
+  std::vector<util::filepath> linkerInputFiles;
+  linkerInputFiles.push_back(inputFiles);
+  auto & llvmLink = LlvmLinkCommand::Create(
+      *commandGraph,
+      linkerInputFiles,
```
phate, line 368: Instead of using the `linkedInputFiles` vector here, could you not just use `{inputFiles}` instead, and then remove the `linkerInputFiles` vector?

phate (changes requested):
Two minor comments.

jlm/tooling/CommandGraphGenerator.cpp
sjalander: Done

jlm/tooling/CommandGraphGenerator.cpp
sjalander: Done

phate (approved):

PR #551, opened 2024-07-18
Add printing of rvsdg tree to jlm-opt

phate:
This PR adds the capability to print the region tree to `jlm-opt`. Here is an example output:

```
RootRegion
-DELTA[.str.3]
-DELTA[array2]
-DELTA[.str.2]
-DELTA[array]
-DELTA[__PRETTY_FUNCTION__.main]
-DELTA[.str.1]
-DELTA[.str]
-DELTA[points]
-LAMBDA[main]
--THETA
---GAMMA
----Region[0]
----Region[1]
--GAMMA
---Region[0]
---Region[1]
----GAMMA
-----Region[0]
-----Region[1]
------GAMMA
-------Region[0]
-------Region[1]
--------GAMMA
---------Region[0]
---------Region[1]
--------GAMMA
---------Region[0]
---------Region[1]
--GAMMA
---Region[0]
---Region[1]

```

### DIFF
--- a/jlm/tooling/Command.cpp
+++ b/jlm/tooling/Command.cpp
@@ -471,6 +471,28 @@ JlmOptCommand::PrintAsMlir(
 }
 
 void
+JlmOptCommand::PrintAsRvsdgTree(
+    const llvm::RvsdgModule & rvsdgModule,
+    const util::filepath & outputFile,
+    util::StatisticsCollector &)
+{
+  auto & rootRegion = *rvsdgModule.Rvsdg().root();
+  auto tree = rvsdg::region::ToTree(rootRegion);
+
+  if (outputFile == "")
+  {
+    std::cout << tree << std::flush;
+  }
+  else
+  {
+    std::ofstream fs;
+    fs.open(outputFile.to_str());
+    fs << tree;
+    fs.close();
+  }
+}
+
+void
 JlmOptCommand::PrintRvsdgModule(
     llvm::RvsdgModule & rvsdgModule,
     const util::filepath & outputFile,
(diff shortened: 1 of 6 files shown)


jlm/tooling/Command.cpp
```diff
@@ -470,6 +470,28 @@ JlmOptCommand::PrintAsMlir(
 #endif
 }
 
+void
+JlmOptCommand::PrintAsRvsdgTree(
+    const llvm::RvsdgModule & rvsdgModule,
+    const util::filepath & outputFile,
+    util::StatisticsCollector &)
+{
+  auto & rootRegion = *rvsdgModule.Rvsdg().root();
```
sjalander, line 479: rootRegion is only used once, so why not inline it?

jlm/tooling/CommandLine.hpp
```diff
@@ -54,6 +54,7 @@ class JlmOptCommandLineOptions final : public CommandLineOptions
     Ascii,
     Llvm,
     Mlir,
+    Tree,
```
sjalander, line 57: Adding something into this class will always require changes to the unordered_map in the cpp (line 311 above).
Would it make sense to declare these in the same place and then use them where they are needed?

This resembles the LLVM attributes that need to be updated when we change LLVM version, which requires 3 different files to be updated.

sjalander (approved):
Only minor comments

sjalander:
@phate Not super fond of the '-' for indentation, but don't know if it would be any easier to read with a simple space ' '.

jlm/tooling/Command.cpp
phate: I had it inlined before, but then moved it out because of readability. Subjective preference. I will leave it as it is.

jlm/tooling/CommandLine.hpp
phate: Yeah, you are right about it. The problem originates from us requiring a string version of the output format value, but c++ does not offer a "ToString()" method on enum types by default (which is insane in 2024). This means we need to implement this ourselves (with this unordered_map).

We have a unit test that ensures we are also adding a line in the map if we add a new output format. The unit test fails if you only add a new output format, but forget the line in the map. The best we can do in the absence of language/compiler support.

phate:
@sjalander I had a simple space before, but found it not very readable with the indentation. I could not clearly see which node is a child of another node. We can make this configurable in another PR if your prefer a simple space.

PR #553, opened 2024-07-19
Remove is_gamma_argument() function

phate:
This PR does the following:
1. Introduces the GammaArgument class, which represents a region argument in a gamma subregion.
2. Removes the is_gamma_argument() function.

### DIFF
--- a/jlm/rvsdg/gamma.hpp
+++ b/jlm/rvsdg/gamma.hpp
@@ -465,6 +465,33 @@ inline gamma_node::gamma_node(jlm::rvsdg::output * predicate, size_t nalternativ
       new gamma_input(this, predicate, ctltype::Create(nalternatives))));
 }
 
+/**
+ * Represents a region argument in a gamma subregion.
+ */
+class GammaArgument final : public argument
+{
+  friend gamma_node;
+
+public:
+  ~GammaArgument() noexcept override;
+
+private:
+  GammaArgument(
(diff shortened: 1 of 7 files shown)


jlm/rvsdg/gamma.hpp
```diff
@@ -465,6 +465,33 @@ inline gamma_node::gamma_node(jlm::rvsdg::output * predicate, size_t nalternativ
...
+private:
+  GammaArgument(
+      rvsdg::region & region,
+      structural_input & input,
+      std::shared_ptr<const rvsdg::type> type)
+      : argument(&region, &input, std::move(type))
+  {}
+
+  static GammaArgument &
+  Create(rvsdg::region & region, structural_input & input, std::shared_ptr<const rvsdg::type> type)
```
haved, line 487: Does the create method need a type specified, or can it get its type from the `input`?

haved (approved):
Looks good, just one question

jlm/rvsdg/gamma.hpp
phate: Remove the type argument.

phate:
@haved Changes since last review:
1. Removed superfluous type from `Create()` method and `GammaArgument` constructor
2. Handed in a `GammaInput` instead of a `structural_input` to `GammaArgument`'s constructor and `Create()` method

haved (approved):

PR #554, opened 2024-07-20
Remove is_gamma_result() function

phate:
This PR does the following:
1. Introduces the GammaResult class, which represents a region result in a gamma subregion.
2. Removes the is_gamma_result() function.


haved (approved):
Looks good

PR #559, opened 2024-07-23
Remove is_theta_argument() function

phate:
This PR does the following:
1. Introduces the ThetaArgument class, which represents a region argument in
a theta subregion.
2. Removes the is_gamma_argument() function.

### DIFF
--- a/jlm/rvsdg/theta.hpp
+++ b/jlm/rvsdg/theta.hpp
@@ -355,6 +355,9 @@ private:
   jlm::rvsdg::theta_input * input_;
 };
 
+/**
+ * Represents a region argument in a theta subregion.
+ */
 class ThetaArgument final : public argument
 {
   friend theta_node;
--- a/jlm/rvsdg/theta.hpp
+++ b/jlm/rvsdg/theta.hpp
@@ -355,6 +355,27 @@ private:
   jlm::rvsdg::theta_input * input_;
 };
 
+class ThetaArgument final : public argument
+{
+  friend theta_node;
+
+public:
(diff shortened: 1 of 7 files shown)


jlm/rvsdg/theta.hpp
```diff
@@ -355,6 +355,30 @@ class theta_output final : public structural_output
...
+public:
+  ~ThetaArgument() noexcept override;
+
+private:
+  ThetaArgument(rvsdg::region & region, theta_input & input)
+      : argument(&region, &input, input.Type())
+  {}
+
+  static ThetaArgument &
+  Create(rvsdg::region & region, theta_input & input)
```
sjalander, line 374: Is there any way of assuring that the region is a theta region?

sjalander (dismissed):
It would be a bit odd to add a ThetaArgument to any type of region. So, is there a way of checking that the region is part of a Theta operation/node?

jlm/rvsdg/theta.hpp
phate: The constructor and creator method of the ThetaArgument class is private and the theta_node class is marked as a friend class. Thus, only the theta_node class is allowed to create a ThetaArgument class. In the theta_node class, the ThetaArgument class is created exactly at one location, and the theta subregion is handed in at that location.

Just to be sure, I added a `JLM_ASSERT()` in the ThetaArgument constructor to check whether it is a theta subregion.

phate:
@sjalander I addressed your comment and need a new review.

sjalander (approved):

Issue #556, opened 2024-07-20
Missing type check in result class constructors


phate:
It seems that the [result class constructors](https://github.com/phate/jlm/blob/b69c33c12d31e001fca22a8dce76457aa75d6f29/jlm/rvsdg/region.cpp#L90C1-L122C2) never check that the type of the handed in `structural_output` is the same as the type of the result itself. Find out whether this is the case (or whether I am misunderstanding something here) and if it is an issue, then fix the problem.

Issue #555, opened 2024-07-20
Missing type check in argument class constructors


phate:
It seems that the [argument class constructors](https://github.com/phate/jlm/blob/b69c33c12d31e001fca22a8dce76457aa75d6f29/jlm/rvsdg/region.cpp#L26C1-L56C2) never check that the type of the handed in `structural_input` is the same as the type of the argument itself. Find out whether this is the case (or whether I am misunderstanding something here) and if it is an issue, then fix the problem.

PR #565, opened 2024-07-24
Clean up region class tests

phate:
This PR does the following:
1. Rename TestRegion.cpp to RegionTests.cpp
2. Transform single unit test to individual unit tests
3. Split out the argument and result class tests into their own respective files
4. Some minor clean-up of the individual unit tests

This PR cleans up the respective unit test files before handling issues #556 and #555.


haved (approved):
Looks good to me!

PR #566, opened 2024-07-30
Add annotation support to region tree

phate:
This PR adds support for annotations to region trees. Here is an example of how it will look like:

```
RootRegion
-STRUCTURAL_TEST_NODE
--Region[0]
--Region[1]
---STRUCTURAL_TEST_NODE
----Region[0]
----Region[1]
----Region[2] NumNodes:0 NumArguments:0
```

The plan is to use this for introducing a pass that enables the printing of RVSDG properties for debugging as well as metric gathering for pass evaluations.

### DIFF
--- a/jlm/rvsdg/region.cpp
+++ b/jlm/rvsdg/region.cpp
@@ -440,9 +440,9 @@ region::ToString(
     char labelValueSeparator)
 {
   std::stringstream stream;
-  for (auto it = annotations.begin(); it != annotations.end(); it++)
+  for (auto & annotation : annotations)
   {
-    auto annotationString = ToString(*it, labelValueSeparator);
+    auto annotationString = ToString(annotation, labelValueSeparator);
     stream << annotationSeparator << annotationString;
   }
 
--- a/jlm/rvsdg/region.cpp
+++ b/jlm/rvsdg/region.cpp
@@ -9,6 +9,7 @@
 #include <jlm/rvsdg/structural-node.hpp>
 #include <jlm/rvsdg/substitution.hpp>
 #include <jlm/rvsdg/traverser.hpp>
+#include <jlm/util/AnnotationMap.hpp>
 
 namespace jlm::rvsdg
 {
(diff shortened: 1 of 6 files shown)


jlm/rvsdg/region.cpp
```diff
@@ -360,47 +361,120 @@ region::NumRegions(const jlm::rvsdg::region & region) noexcept
...
+  return stream.str();
+}
+
+std::string
+region::ToString(const util::Annotation & annotation, char labelValueSeparator)
+{
+  std::string value;
+  if (annotation.HasValueType<std::string_view>())
+  {
+    value = util::strfmt(annotation.Value<std::string_view>());
```
haved, line 458: Do we need strfmt to convert a `string_view` to a string? I would think it is implicit?

On another note, is it maybe a bit limiting to only have `string_view` as the string value, in case some strings need to be heap allocated. I have always wished for a string type that can both represent owned and non-owned strings.

jlm/rvsdg/region.hpp
```diff
@@ -390,6 +396,17 @@ class region
   [[nodiscard]] static size_t
   NumRegions(const jlm::rvsdg::region & region) noexcept;
 
+  /**
+   * Converts \p region and all of its contained structural nodes with subregions to a tree in
+   * ASCII format.
+   *
+   * @param region The top-level region that is converted
+   * @param annotationMap A map with annotations for region%s or structural_node%s.
```
haved, line 404: Why are there percentage signs here instead of apostrophes?

jlm/rvsdg/region.hpp
```diff
@@ -390,6 +396,17 @@ class region
   [[nodiscard]] static size_t
   NumRegions(const jlm::rvsdg::region & region) noexcept;
 
+  /**
+   * Converts \p region and all of its contained structural nodes with subregions to a tree in
+   * ASCII format.
```
haved, line 401: An example of the output would be nice in this comment

jlm/util/AnnotationMap.hpp
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   * Checks if the type of the annotation value is equivalent to \p TValue.
+   *
+   * @return True if the value type if equivalent to \p TValue, otherwise false.
+   */
+  template<typename TValue>
+  [[nodiscard]] bool
+  HasValueType() const noexcept
+  {
+    return std::get_if<TValue>(&Value_) != nullptr;
```
haved, line 77: `std::holds_alternative` is maybe nicer?

jlm/rvsdg/region.cpp
```diff
@@ -360,47 +361,120 @@ region::NumRegions(const jlm::rvsdg::region & region) noexcept
...
     {
-      auto identationString = std::string(identationDepth, identationChar);
       auto nodeString = structuralNode->operation().debug_string();
-      subTree += util::strfmt(identationString, nodeString, '\n');
+      auto annotationString = GetAnnotationString(
+          structuralNode,
+          annotationMap,
+          annotationSeparator,
+          labelValueSeparator);
+      stream << identationString << nodeString << annotationString << '\n';
```
haved, line 412: Is this line supposed to be indented one level more than the region line? It re-uses the same indentationString, but the indentation variable has been incremented, so I'm not quite sure if this is intended.

jlm/rvsdg/region.cpp
```diff
@@ -360,47 +361,120 @@ region::NumRegions(const jlm::rvsdg::region & region) noexcept
...
+  util::AnnotationMap annotationMap;
+  ToTree(region, annotationMap, 0, stream);
+  return stream.str();
+}
+
+void
+region::ToTree(
+    const rvsdg::region & region,
+    const util::AnnotationMap & annotationMap,
+    size_t identationDepth,
```
haved, line 384: Is it spelled "indentation"? Typo?

haved (approved):
Looks good, nice to be able to quickly dump info. Some comments and questions

jlm/rvsdg/region.cpp
phate: Removed the strfmt. I also changed the string value type from `std::string_view` to `std::string`

jlm/rvsdg/region.hpp
phate: I added an example to the documentation of the two methods.

jlm/util/AnnotationMap.hpp
phate: It is nicer. Changed!

jlm/rvsdg/region.cpp
phate: Yes, typo. Changed.

jlm/rvsdg/region.hpp
phate: The doxygen output will look like this:

Screenshot_2024-07-31_20-41-42

The percentage signs ensure that I can create grammatically correct sentences while still getting the links to the data structures.

The "Links to classes" section in the [documentation](https://www.doxygen.nl/manual/autolink.html) gives some more information.

jlm/rvsdg/region.cpp
phate: Great catch! I seem to have accidentally moved the line for creating the indentations above the depth increase. I checked the output of the tests again to ensure it looks correct.

haved (approved):
Looks good! I will try to read up on doxygen to make use of the syntax

PR #568, opened 2024-08-07
Generalize CIRCT and MLIR jlm dialect build scripts

haved:
I was having an issue where tests of the command parser would fail due to some MLIR option, even when neither `--enable-mlir` nor `--enable-hls` are passed to `configure.sh`. That is a separate issue that should be fixed, but I thought it was about time I got MLIR and CIRCT running on my own machine anyway.

These changes should make the build scripts work no matter where LLVM is installed.

This is as a draft PR for now, to check that these changes do not break CI.

PS: Some `.cpp` files used symbols imported from the standard library indirectly, so I had to add some `#include <>` to make it work. I am on arch (btw), but I am guessing these changes will become necessary on Ubuntu eventually as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ Makefile.custom
 docs/html
 build
 build-*
+usr
--- a/configure.sh
+++ b/configure.sh
@@ -114,15 +114,6 @@ if [ "${ENABLE_HLS}" == "yes" ] ; then
 	CXXFLAGS_NO_COMMENT="-Wno-error=comment"
 	CIRCT_LDFLAGS_ARRAY=(
 		"-L${CIRCT_PATH}/lib"
-		"-lMLIR"
-		"-lMLIRBytecodeReader"
-		"-lMLIRBytecodeWriter"
-		"-lMLIRParser"
-		"-lMLIRSupport"
-		"-lMLIRIR"
-		"-lMLIROptLib"
-		"-lMLIRFuncDialect"
-		"-lMLIRTransforms"
 		"-lCIRCTAnalysisTestPasses"
 		"-lCIRCTDependenceAnalysis"
 		"-lCIRCTExportFIRRTL"
@@ -149,6 +140,7 @@ if [ "${ENABLE_HLS}" == "yes" ] ; then
 		"-lCIRCTExportChiselInterface"
 		"-lCIRCTOM"
 		"-lCIRCTSupport"
+		"-lMLIR"
 	)
 fi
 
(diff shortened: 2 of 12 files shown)


haved:
For future reference, these are the commands used to build LLVM. We need extra options to enable RTTI, and the dylib `libMLIR.so`:
```bash
git clone git@github.com:llvm/llvm-project.git ${LLVM_REPO}
git -C ${LLVM_REPO} checkout release/17.x

cmake ${LLVM_REPO}/llvm \
    -GNinja \
    -DCMAKE_INSTALL_PREFIX=/usr/local/lib/llvm17 \
    -DCMAKE_BUILD_TYPE=Release \
    -DLLVM_ENABLE_PROJECTS="clang;lld;mlir" \
    -DLLVM_ENABLE_RTTI=ON \
    -DLLVM_TARGETS_TO_BUILD=X86 \
    -DLLVM_BUILD_LLVM_DYLIB=ON \
    -DLLVM_PARALLEL_COMPILE_JOBS=5 \
    -DLLVM_PARALLEL_LINK_JOBS=1

cmake --build .
sudo cmake --build . --target install
```

sjalander:
@haved 
The PR looks good, but you state that it is a draft PR. Is there more changes that you envision or should I go ahead and approve it?

Regarding building LLVM.
Can you create a build script similar to CIRCT and MLIR for it?

haved:
@sjalander I first want to make sure everything ends up working on my machine, in case I have to modify anything.

I can create a similar script for LLVM, but I would not recommend we ever build LLVM in CI, as it takes forever, and at least 7 GB of disk space.

sjalander:
@haved
The script would not be for the CI, but for people that would like to build their own LLVM version for jlm.

sjalander:
@haved 
Thanks for cleaning up and improving the configure and build scripts.
You are correct that the thing referred to as MLIR is the MLIR backend of jlm, i.e., our own dialect.

I don't think anything is compiled with both the CIRCT_LDFLAGS and MLIR_LDFLAGS being used, as they are for different backends. So the standard MLIR library should only be listed once during compilation.

We could rename MLIR_LDFLAGS to something more descriptive, e.g., RVSDG_MLIR_LDFLAGS.
We could then also separate out the standard MLIR library with its own defined variable, i.e., MLIR_LDFLAGS, that is included when needed to assure that it is only applied once. But, I'm not sure that we would gain much by doing the latter.

haved:
I agree that no target uses both, so there should be no problems with that. Even if some target did it should not be a problem. It is just something to keep in mind when seeing that `-lMLIR` is included in both. We kind have the same naming collision with `llvm` referring to the library built from the `jlm/llvm` folder.

haved:
@sjalander everything works on my end, should we merge the PR?

sjalander (approved):

PR #569, opened 2024-08-07
Avoid OutputFormatParsing test failing when MLIR is disabled

haved:
A test was failing when compiling without `--enable-mlir`, due to `--output-format mlir` being unknown when `ENABLE_MLIR` is not defined.

This issue could have been fixed in a different location, such as not defining the `Mlir` value in the `OutputFormat` enum at all, but that would have created some additional situations where `#ifdef`s would have to be added.

I think it is nice to use as few `#ifdef`s as possible, to avoid situations like this where tests fail in specific `#ifdef` configurations. I would not be opposed to including the `--output-format mlir` option in all builds, but possibly making it invisible in the help-text.
If `OutputFormat::Mlir` is selected without being enabled in the build, it eventually triggers a `JLM_UNREACHABLE` that informs you that jlm has been compiled without the MLIR dialect. I think this is much more user-friendly than just denying that there ever was an output format called `mlir`.

Thoughts? @phate, @sjalander 

PS: I also had to add `#include <algorithm>` to compile this locally.

### DIFF
--- a/tests/jlm/tooling/TestJlcCommandLineParser.cpp
+++ b/tests/jlm/tooling/TestJlcCommandLineParser.cpp
@@ -13,14 +13,38 @@
 static const jlm::tooling::JlcCommandLineOptions &
 ParseCommandLineArguments(const std::vector<std::string> & commandLineArguments)
 {
-  std::vector<const char *> cStrings;
+  auto cleanUp = [](const std::vector<char *> & array)
+  {
+    for (const auto & ptr : array)
+    {
+      delete[] ptr;
+    }
+  };
+
+  std::vector<char *> array;
   for (const auto & commandLineArgument : commandLineArguments)
-    cStrings.push_back(commandLineArgument.c_str());
+  {
+    array.push_back(new char[commandLineArgument.size() + 1]);
+    strncpy(array.back(), commandLineArgument.data(), commandLineArgument.size());
+    array.back()[commandLineArgument.size()] = '\0';
+  }
 
   static jlm::tooling::JlcCommandLineParser commandLineParser;
-  return commandLineParser.ParseCommandLineArguments(
-      static_cast<int>(cStrings.size()),
-      cStrings.data());
+  const jlm::tooling::JlcCommandLineOptions * commandLineOptions;
+  try
+  {
+    commandLineOptions =
+        &commandLineParser.ParseCommandLineArguments(static_cast<int>(array.size()), &array[0]);
+  }
+  catch (...)
+  {
+    cleanUp(array);
+    throw;
+  }
+
+  cleanUp(array);
+
+  return *commandLineOptions;
 }
 
 static void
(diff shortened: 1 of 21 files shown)


haved:
@sjalander 
It has not failed in CI due to `--enable-mlir` being passed to `configure.sh` in the testing workflow.

I agree that tests of the MLIR code should only run when the MLIR dialect is enabled. That is also what is mostly happening. The test in question is not really about MLIR, but rather the command line parsing, which is modified ever so slightly depending on `ENABLE_MLIR`. The code that is currently in this PR attempts to fix the test to not test the specific Mlir option when it is disabled.

It is mostly a matter of taste: where and how much should the preprocessor be used to make mentions of MLIR disappear in builds without MLIR.

sjalander:
I see now that all make checks has both HLS and MLIR enabled. Wondering if we should create tests that run the check with these also disabled to catch these kinds of errors.

@haved 
I agree with that the test is primarily not about MLIR and there is no elegant way of splitting it into multiple versions.

sjalander (approved):

haved:
@phate @sjalander
Update: LLVM does not enable hiding specific options from the help list of a flag, so I kept the current setup where the `mlir` option is skipped when compiling without `-DENABLE_MLIR`.

This PR fixes the issue by skipping testing `--output-format mlir` when MLIR is disabled.

The rest of the changes made by this PR is removing the `try/catch`, `new char[]` and `strcpy` stuff used in `TestJlmOptCommandLineParser.cpp` and `TestJlcCommandLineParser.cpp`, by using the `c_str()` method instead. This required adding `const` to the parser functions, but I think it makes sense for them to be `const` anyways.

tests/jlm/tooling/TestJlcCommandLineParser.cpp
```diff
@@ -13,38 +13,14 @@
...
-
-  std::vector<char *> array;
+  std::vector<const char *> cStrings;
   for (const auto & commandLineArgument : commandLineArguments)
-  {
-    array.push_back(new char[commandLineArgument.size() + 1]);
-    strncpy(array.back(), commandLineArgument.data(), commandLineArgument.size());
-    array.back()[commandLineArgument.size()] = '\0';
-  }
+    cStrings.push_back(commandLineArgument.c_str());
```
sjalander, line 18: Isn't curly brackets supposed to always be used?

tests/jlm/tooling/TestJlmOptCommandLineParser.cpp
```diff
@@ -164,6 +139,11 @@ OutputFormatParsing()
   for (size_t n = start; n != end; n++)
   {
     auto outputFormat = static_cast<JlmOptCommandLineOptions::OutputFormat>(n);
+#ifndef ENABLE_MLIR
```
sjalander, line 142: The PR title is about this fix, but the majority of the changes are related to the cstrings.
It might be better to split it into two PRs, as they are not related.

haved: Yes, I reverted those changes to put them in a separate PR

haved:
@sjalander I reverted the `const char * const *` changes, have a look now

sjalander (approved):

PR #570, opened 2024-08-09
[AndersenAgnostic] Use new Andersen Configuration setup and change how double checking is performed

haved:
This PR is the first step towards getting all the Andersen techniques into main.

It makes a couple of changes, without adding any new techniques:
 - Configurations have a text representation
 - Configurations are not specified via environment variable, but there is instead a function for getting all possible configurations.
 - Double-checking against a different configuration is cleaner and faster. `PointerObjectSet`s are cloned, solved separately, and compared directly after solving.
 - The template stuff used to invoke the worklist solver uses `std::visit` now, making dispatch much nicer.
 - Some additional statistics have been added
 - Re-orders some of the existing operations in the worklist solver. This is the order they ended up in in the final code when all techniques are added.
 - Added `Erase` to `HashSet`, which is a thin wrapper around `std::unordered_set::erase()`.
 - Register tests in `TestAndersen.cpp` individually.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/Andersen.hpp
+++ b/jlm/llvm/opt/alias-analyses/Andersen.hpp
@@ -176,7 +176,11 @@ public:
     [[nodiscard]] static Configuration
     NaiveSolverConfiguration() noexcept
     {
-      return Configuration{};
+      Configuration config;
+      config.EnableOfflineVariableSubstitution(false);
+      config.EnableOfflineConstraintNormalization(false);
+      config.SetSolver(Solver::Naive);
+      return config;
     }
 
     /**
--- a/jlm/llvm/opt/alias-analyses/Andersen.hpp
+++ b/jlm/llvm/opt/alias-analyses/Andersen.hpp
@@ -32,35 +32,22 @@ class Andersen final : public AliasAnalysis
 
 public:
   /**
-   * Environment variable that will trigger double checking of the analysis,
-   * by running analysis again with the naive solver and no extra processing.
-   * Any differences in the produced PointsToGraph result in an error.
+   * Environment variable that when set, triggers analyzing the program with every single
+   * valid combination of Configuration flags.
    */
-  static inline const char * const ENV_COMPARE_SOLVE_NAIVE = "JLM_ANDERSEN_COMPARE_SOLVE_NAIVE";
+  static inline const char * const ENV_TEST_ALL_CONFIGS = "JLM_ANDERSEN_TEST_ALL_CONFIGS";
+
+  /**
+   * Environment variable that will trigger double checking of the analysis.
+   * If ENV_TEST_ALL_CONFIGS is set, the output is double checked against them all.
+   * Otherwise, the output is double checked only against the default naive solver.
+   */
+  static inline const char * const ENV_DOUBLE_CHECK = "JLM_ANDERSEN_DOUBLE_CHECK";
 
   /**
    * Environment variable that will trigger dumping the subset graph before and after solving.
    */
   static inline const char * const ENV_DUMP_SUBSET_GRAPH = "JLM_ANDERSEN_DUMP_SUBSET_GRAPH";
-
-  /**
-   * Environment variable for overriding the default configuration.
-   * The variable should something look like
-   * "+OVS +Normalize -OnlineCD Solver=Worklist WLPolicy=LRF"
-   */
-  static inline const char * const ENV_CONFIG_OVERRIDE = "JLM_ANDERSEN_CONFIG_OVERRIDE";
-  static inline const char * const CONFIG_OVS_ON = "+OVS";
-  static inline const char * const CONFIG_OVS_OFF = "-OVS";
-  static inline const char * const CONFIG_NORMALIZE_ON = "+Normalize";
-  static inline const char * const CONFIG_NORMALIZE_OFF = "-Normalize";
-  static inline const char * const CONFIG_SOLVER_WL = "Solver=Worklist";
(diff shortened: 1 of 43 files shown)


jlm/llvm/opt/alias-analyses/Andersen.hpp
```diff
@@ -191,21 +176,24 @@ class Andersen final : public AliasAnalysis
     [[nodiscard]] static Configuration
     NaiveSolverConfiguration() noexcept
     {
-      auto config = Configuration();
-      config.EnableOfflineVariableSubstitution(false);
-      config.EnableOfflineConstraintNormalization(false);
-      config.SetSolver(Solver::Naive);
-      config.EnableOnlineCycleDetection(false);
-      return config;
+      return Configuration{};
```
sjalander, line 179: Is this a good practice?
A change of the default constructor now has an effect on what the NaiveSolverConfiguration() means.

haved: The solution I went with is not the cleanest, no. I have a comment about default values of members being the "naive" configuration, but it is probably cleaner to not have a default constructor at all, and force assigning every field.

The reason I didn't do this is that I ended up with constructor invocations that looked like `Configuration(false, false, Naive, LeastRecentlyFired, false, false, false, false, true)`.

Constructors are nicer, through, as we get errors if the signature changes.
Maybe I could add names to the values somehow, to make everything clearer?

sjalander: For the NaiveSolverConfiguration(), it could explicitly set the parameters the way the deleted version did, instead of relying on the default constructor default settings.

phate: I agree with Magnus here. Regarding the names for the constructor, C# had support for this:

c.foobar(
bar: false,
blub: 4,
....);

which was really nice. The closest to this you can get in C++ is comments:

c.foobar(
false, // bar
4,       // blub
...);

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -373,6 +390,44 @@ PointerObjectSet::Clone() const
...
+    if (IsPointingToExternal(i) != other.IsPointingToExternal(i))
+      return false;
+
+    auto & thisPointsToSet = GetPointsToSet(i);
+    auto & otherPointsToSet = other.GetPointsToSet(i);
+
+    for (auto thisPointee : thisPointsToSet.Items())
+    {
+      // Skip doubled-up pointers, as they do not affect the points-to set
+      if (HasEscaped(thisPointee) && IsPointingToExternal(i))
```
phate, line 414: Maybe introduce a method `IsDoubledUp()` that is used instead of this pattern (which is used at multiple places)?

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
```diff
@@ -876,10 +920,10 @@ class PointerObjectConstraintSet final
    * Finds a least solution satisfying all constraints, using the Worklist algorithm.
    * Descriptions of the algorithm can be found in
    *  - Pearce et al. 2003: "Online cycle detection and difference propagation for pointer analysis"
-   *  - Hardekopf et al. 2007: "The Ant and the Grasshopper".
+   *    - Online Cycle Detection
```
phate, line 923: Indentation

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -1006,9 +1093,9 @@ PointerObjectConstraintSet::CreateOvsSubsetGraph()
   std::vector<util::HashSet<PointerObjectIndex>> successors(totalNodeCount);
   std::vector<bool> isDirectNode(totalNodeCount, false);
 
-  // Initialize all registers as direct nodes
+  // Initialize all registers with empty points-to sets as direct nodes
   for (auto [_, index] : Set_.GetRegisterMap())
-    isDirectNode[index] = true;
+    isDirectNode[index] = Set_.GetPointsToSet(index).IsEmpty();
```
phate, line 1098: The `isEmpty()` suggests to me that a bool is returned, but the comment above tells me that it is an empty points-to set. If the `IsEmpty()` method does not return a bool, then maybe just call it `Empty()` or `EmptySet()`?

phate (approved):
A few minor comments.

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
haved: I rewrote it a bit and added a general `IsPointingTo`, that returns true for explicit, implicit, and doubled-up pointer/pointee relations.

jlm/llvm/opt/alias-analyses/PointerObjectSet.hpp
haved: This is a sublist, since Difference Propagation is a technique described in the paper. I changed the comment to make it clearer.

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
haved: This is a bool, but the comment is worded poorly. Fixing

phate (approved):

haved:
@phate the CI jobs checks are all listed as "pending". Do they need to be triggered manually somehow?

PR #572, opened 2024-08-11
HLS: Configurable pointer size

sjalander:
Makes it possible to manually change the size of a pointer by changing GetPointerSizeInBits() in base-hls.cpp.


phate (approved):
LGTM

PR #573, opened 2024-08-12
Run hls::cne and llvm::dne as part of the passes

sjalander:


jlm/hls/backend/rvsdg2rhls/add-forks.cpp
```diff
@@ -37,8 +37,8 @@ add_forks(jlm::rvsdg::region * region)
         add_forks(structnode->subregion(n));
       }
     }
-    // If a node has no inputs it is a constant
-    bool isConstant = node->ninputs() == 0;
+    bool isConstant = rvsdg::is<rvsdg::bitconstant_op>(node->operation())
+                   || rvsdg::is<rvsdg::ctlconstant_op>(node->operation());
```
phate, line 41: What is the reason for changing this?

phate (dismissed):

jlm/hls/backend/rvsdg2rhls/add-forks.cpp
phate: @sjalander

sjalander: @davidmetz 
Can you explain why this change has been made?

This is how the code is for the addr-queue branch, but the earlier version of the code, which checks for no inputs, also works for the benchmarks that are included in the hls-test-suite.

phate: @sjalander @davidmetz

sjalander: @phate 
I discussed with David, and this change was to handle local memory. But further changes has med this unnecessary as the test suite works without it.
So I've reverted the change.

phate (approved):

PR #576, opened 2024-08-14
[AndersenAgnostic] Add hybrid cycle detection to Andersen

haved:
Adds Hybrid Cycle Detection from Hardekopf 2007. The implementation is quite simple, but requires some extra info to be kept from the Offline Variable Substitution.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
+++ b/jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
@@ -1317,9 +1317,10 @@ PointerObjectConstraintSet::PerformOfflineVariableSubstitution(bool storeRefCycl
     const size_t derefNodeOffset = Set_.NumPointerObjects();
     for (PointerObjectIndex i = 0; i < Set_.NumPointerObjects(); i++)
     {
-      const auto optRoot = unificationRootPerSCC[sccIndex[i + derefNodeOffset]];
-      if (optRoot)
+      if (auto optRoot = unificationRootPerSCC[sccIndex[i + derefNodeOffset]])
+      {
         RefNodeUnificationRoot_[i] = *optRoot;
+      }
     }
   }
 
--- a/jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
+++ b/jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
@@ -11,6 +11,7 @@
 
 #include <limits>
 #include <queue>
+#include <variant>
 
 namespace jlm::llvm::aa
 {
@@ -1234,7 +1235,7 @@ AssignOvsEquivalenceSetLabels(
 }
 
 size_t
-PointerObjectConstraintSet::PerformOfflineVariableSubstitution()
+PointerObjectConstraintSet::PerformOfflineVariableSubstitution(bool storeRefCycleUnificationRoot)
 {
   // Performing unification on direct nodes relies on all subset edges being known offline.
   // This is only safe if no more constraints are added to the node in the future.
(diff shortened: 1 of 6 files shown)


jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -1297,6 +1298,31 @@ PointerObjectConstraintSet::PerformOfflineVariableSubstitution()
       unificationRoot[equivalenceSetLabel] = i;
   }
 
+  // If hybrid cycle detection is enabled, it requires some information to be kept from OVS
+  if (storeRefCycleUnificationRoot)
+  {
+    // For each ref node that is in a cycle with regular node, store it for hybrid cycle detection
```
phate, line 1304: "...with **a** regular node..."

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -1297,6 +1298,31 @@ PointerObjectConstraintSet::PerformOfflineVariableSubstitution()
...
+
+    // First find one unification root representing each SCC
+    std::vector<std::optional<PointerObjectIndex>> unificationRootPerSCC(numSccs, std::nullopt);
+    for (PointerObjectIndex i = 0; i < Set_.NumPointerObjects(); i++)
+    {
+      if (!unificationRootPerSCC[sccIndex[i]])
+        unificationRootPerSCC[sccIndex[i]] = Set_.GetUnificationRoot(i);
+    }
+
+    // Assign unification roots to ref nodes that are in CYCLES with regular nodes
```
phate, line 1316: Why is cycles written with capital letters?

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -1297,6 +1298,31 @@ PointerObjectConstraintSet::PerformOfflineVariableSubstitution()
...
+    const size_t derefNodeOffset = Set_.NumPointerObjects();
+    for (PointerObjectIndex i = 0; i < Set_.NumPointerObjects(); i++)
+    {
+      const auto optRoot = unificationRootPerSCC[sccIndex[i + derefNodeOffset]];
+      if (optRoot)
+        RefNodeUnificationRoot_[i] = *optRoot;
```
phate, line 1322:
```
if (auto optRoot = ...)
{
  RefNodeUnificationRoot_[i] = ....
}

File: jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp [hunk 4]
```diff
@@ -1360,7 +1386,7 @@ PointerObjectConstraintSet::NormalizeConstraints()
   return reduction;
 }
 
-template<typename Worklist, bool EnableOnlineCycleDetection>
+template<typename Worklist, bool EnableOnlineCycleDetection, bool EnableHybridCycleDetection>
```
phate, line 1389: As far as I understand it, it should never happen that both, `EnableOnlineCycleDetection` and `EnableHybridCycleDetection`, are enabled at the same time. Maybe we can add a `static_assert<>()` that tries to check for that?

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -1458,6 +1484,17 @@ PointerObjectConstraintSet::RunWorklistSolver(WorklistStatistics & statistics)
     callConstraints[root].UnionWith(callConstraints[nonRoot]);
     callConstraints[nonRoot].Clear();
 
+    if constexpr (EnableHybridCycleDetection)
+    {
+      // When unifying, keep at least one of the ref node unification roots
```
phate, line 1489: The comment indicates that at least one root should ALWAYS be kept. The code below does not enforce this though. Could it still be that `RefNodeUnificationRoot_[root].count(root)` is 0 after this piece of code? If not, maybe it would be possible to put an assert there?

phate (approved):

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
haved: I had a bug where I assigned based on equivalence set label, so when I figured out the bug I was probably a bit eager to change the comment. Will calm it down

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
haved: I changed the comment to make it clearer. It is possible that neither the root or the root's new child have their ref node in a cycle with a regular node

phate (approved):

PR #577, opened 2024-08-14
Remove impport class

phate:
This PR does the following:

1. Introduces the GraphImport class and its subclasses to model external entities to the module. These classes serve as a replacement for the old impport classes.
2. Replaces all usages of impport with GraphImport
3. Removes old impport classes.

This is one of the steps necessary in order to completely remove ports from the code base.


jlm/llvm/ir/RvsdgModule.hpp
```diff
@@ -14,70 +14,51 @@
 namespace jlm::llvm
 {
 
-/* impport class */
-
-class impport final : public jlm::rvsdg::impport
+// FIXME: add documentation
```
phate, line 17: Add documentation

jlm/rvsdg/graph.hpp
```diff
@@ -22,48 +22,21 @@
 namespace jlm::rvsdg
 {
 
-/* impport class */
-
-class impport : public port
+// FIXME: add documentation
```
phate, line 25: Add documentation

tests/test-operation.hpp
```diff
@@ -20,6 +20,23 @@
 namespace jlm::tests
 {
 
+// FIXME: documentation
```
phate, line 23: Add documentation

jlm/hls/backend/rvsdg2rhls/rvsdg2rhls.cpp
```diff
@@ -451,9 +452,10 @@ dump_ref(llvm::RvsdgModule & rhls, std::string & path)
   instrument_ref(*reference);
   for (size_t i = 0; i < reference->Rvsdg().root()->narguments(); ++i)
   {
-    auto arg = reference->Rvsdg().root()->argument(i);
-    auto imp = dynamic_cast<const llvm::impport *>(&arg->port());
```
haved, line 455: This right here is the code that fails in the HLS tests in #578. The changes you make in this PR would also fix that issue (getting rid of ports sounds great). I also appreciate that the new code uses AssertedCast.

jlm/llvm/ir/RvsdgModule.hpp
```diff
@@ -14,70 +14,57 @@
...
   {
     return ValueType_;
   }
 
-  [[nodiscard]] const jlm::rvsdg::valuetype &
-  GetValueType() const noexcept
+  GraphImport &
+  Copy(rvsdg::region & region, rvsdg::structural_input * input) override;
+
+  static GraphImport &
```
caleridas, line 50: This is the only part that I find a bit odd, returning a reference -- I would have found it more natural to return a pointer: catching a returned pointer with "auto x = GraphImport::Create(...)" does the right thing, while it will do totally the wrong thing when returning a reference.

caleridas (approved):
Looks good, just (transitively) deriving from output and attaching all information to this looks logical. See one comment about API inside, have a look if you may want to adapt it.

jlm/llvm/ir/RvsdgModule.hpp
phate: > catching a returned pointer with "auto x = GraphImport::Create(...)" does the right thing, while it will do totally the wrong thing when returning a reference.

What do you mean by "will do totally the wrong thing"?

Instead of `auto x = ...`, you do now `auto & x = ...`. I do not see the problem? The copy constructor and copy assignment operator are deleted from output (and therefore argument). Thus, it is also not possible to accidentally do `auto x = ...` and create a unwanted copy. The benefit of a reference is that the user knows a initialized argument is returned (instead of a potential nullptr).

caleridas: okay relying on deleted copy constructor is safe indeed

PR #578, opened 2024-08-15
Add Copy() method to argument class

phate:
This PR does the following:

1. Introduces a Copy() method to the argument class.
2. The introduced method enables us to fix a bug in the copy() method of the region. Previously, arguments were not copied according to their argument subtypes, but simply a new instance of argument was created.
3. Add unit test.

This PR is a necessary step to get rid off ports.

### DIFF
--- a/jlm/llvm/ir/operators/Phi.cpp
+++ b/jlm/llvm/ir/operators/Phi.cpp
@@ -169,8 +169,9 @@ rvargument::~rvargument()
 {}
 
 rvargument &
-rvargument::Copy(rvsdg::region & region, rvsdg::structural_input *)
+rvargument::Copy(rvsdg::region & region, rvsdg::structural_input * input)
 {
+  JLM_ASSERT(input == nullptr);
   return *rvargument::create(&region, Type());
 }
 
--- a/jlm/llvm/ir/operators/Phi.cpp
+++ b/jlm/llvm/ir/operators/Phi.cpp
@@ -168,10 +168,23 @@ rvoutput::~rvoutput()
 rvargument::~rvargument()
 {}
 
+rvargument &
+rvargument::CopyTo(rvsdg::region & region, rvsdg::structural_input *)
+{
+  return *rvargument::create(&region, Type());
+}
+
 /* phi context variable argument class */
 
 cvargument::~cvargument()
 {}
+
+cvargument &
+cvargument::CopyTo(rvsdg::region & region, rvsdg::structural_input * input)
+{
+  auto phiInput = util::AssertedCast<cvinput>(input);
+  return *cvargument::create(&region, phiInput, Type());
+}
 
 /* phi recursion variable result class */
 
--- a/jlm/llvm/ir/operators/Phi.cpp
+++ b/jlm/llvm/ir/operators/Phi.cpp
@@ -169,7 +169,7 @@ rvargument::~rvargument()
 {}
 
 rvargument &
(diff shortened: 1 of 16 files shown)


jlm/llvm/ir/operators/Phi.cpp
```diff
@@ -168,11 +168,24 @@ rvoutput::~rvoutput()
 rvargument::~rvargument()
 {}
 
+rvargument &
+rvargument::CopyTo(rvsdg::region & region, rvsdg::structural_input *)
```
haved, line 172: I assume it is intentional to ignore the `structural_input`, as recursion variables do not take inputs.
Would it improve clarity if we added an assert to ensure that the input is in fact `nullptr`?

jlm/llvm/ir/operators/lambda.cpp
```diff
@@ -418,10 +418,23 @@ output::~output() = default;
 
 fctargument::~fctargument() = default;
 
+fctargument &
+fctargument::CopyTo(rvsdg::region & region, rvsdg::structural_input * input)
+{
```
haved, line 423: Same here, assert `input` is `nullptr`?

haved (approved):
Looks good. Making `argument` abstract seems like a good plan.

jlm/rvsdg/region.cpp
```diff
@@ -66,6 +66,12 @@ argument::argument(
   }
 }
 
+argument &
+argument::CopyTo(rvsdg::region & region, jlm::rvsdg::structural_input * input)
+{
+  return *argument::create(&region, input, Type());
```
haved, line 72: Replacing `Type()` with `port()` here should fix the HLS test failure.

haved:
@phate We looked at why the HLS test fails.

The `test_global.hls` test fails due to the port of a graph import no longer being of type `llvm::impport`. This happens due to the copy creating a new `rvsdg::port` when the argument is copied, instead of copying the port.

I added a comment to `CopyTo` suggesting a fix (replace `Type()` with `port()`). This fix is of course temporary, as the plan is to remove ports entirely.

jlm/rvsdg/region.cpp
phate: Thanks for looking into this. This saved me a lot of time. Appreciated!

jlm/llvm/ir/operators/Phi.cpp
phate: I have added `JLM_ASSERT()`s. The only concern I have is that this might result in unused variables errors during compilation when we compile in release without asserts. I have not tested.

jlm/llvm/ir/operators/lambda.cpp
phate: Done. See above.

jlm/llvm/ir/operators/Phi.cpp
phate: I also opened issue #579 to not forget that we should do something about this.

phate: It compiles, I just checked. I think what saved us is the `-Wno-unused-parameter` compilation flag. We need it in order to convince the compiler to compile with the LLVM headers.

haved (approved):
Looks good. If I remember correctly, the JLM_ASSERT macro turns into a `((void)(CONDITION))` when asserts are disabled, so that should prevent issues with unused variables. I'm not sure though

PR #581, opened 2024-08-16
[AndersenAgnostic] Add lazy cycle detection as an optional technique for Andersen

haved:
When propagating along the edge a->b, it is because a has something new. If b already has everything a has, that might indicate that there is a path from b to a. Lazy cycle detection looks for such a path using a DFS from b, and unifies everything that is on such a path b->*->a. It tries at most once per edge.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/LazyCycleDetection.hpp
+++ b/jlm/llvm/opt/alias-analyses/LazyCycleDetection.hpp
@@ -47,7 +47,7 @@ public:
   }
 
   bool
-  IsInitialized() const
+  IsInitialized() const noexcept
   {
     return NodeStates_.size() == Set_.NumPointerObjects();
   }
@@ -103,7 +103,7 @@ public:
       else if (NodeStates_[node] == NodeStateVisited)
       {
         DfsStack_.pop();
-        NodeStates_[node] = NodeStateLeft;
+        NodeStates_[node] = NodeStatePopped;
 
         // Check if any successors are unified with the subset. If so, join it!
         for (auto successor : GetSuccessors_(node))
@@ -138,7 +138,7 @@ public:
    * @return the number of DFSs performed to look for cycles
    */
   [[nodiscard]] size_t
-  NumCycleDetectionAttempts()
+  NumCycleDetectionAttempts() const noexcept
   {
     return NumCycleDetectAttempts_;
   }
@@ -147,7 +147,7 @@ public:
    * @return the number of cycles detected by Lazy cycle detection
    */
   [[nodiscard]] size_t
-  NumCyclesDetected()
+  NumCyclesDetected() const noexcept
   {
     return NumCyclesDetected_;
   }
@@ -156,7 +156,7 @@ public:
    * @return the number of unifications made while eliminating found cycles
    */
   [[nodiscard]] size_t
-  NumCycleUnifications()
+  NumCycleUnifications() const noexcept
   {
     return NumCycleUnifications_;
   }
@@ -174,7 +174,7 @@ private:
   // Possible states of nodes during the DFS
   static constexpr uint8_t NodeStateNotVisited = 0;
   static constexpr uint8_t NodeStateVisited = 1;
-  static constexpr uint8_t NodeStateLeft = 2;
+  static constexpr uint8_t NodeStatePopped = 2;
   std::vector<uint8_t> NodeStates_;
 
   size_t NumCycleDetectAttempts_ = 0;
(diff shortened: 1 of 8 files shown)


jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -1398,6 +1403,7 @@ PointerObjectConstraintSet::RunWorklistSolver(WorklistStatistics & statistics)
   if constexpr (EnableOnlineCycleDetection)
   {
     static_assert(!EnableHybridCycleDetection, "OnlineCD can not be combined with HybridCD");
+    static_assert(!EnableLazyCycleDetection, "OnlineCD can not be combined with LazyCD");
   }
```
phate, line 1407: What about the combination of Lazy cycle detection and hybrid cycle detection? This is not checked for, is it?

phate: Or can both be enabled at the same time?

phate: It seem to look like it, looking at the rest of the code...

jlm/llvm/opt/alias-analyses/LazyCycleDetection.hpp
```diff
@@ -0,0 +1,187 @@
...
+      return subset;
+    }
+    return std::nullopt;
+  }
+
+  /**
+   * @return the number of DFSs performed to look for cycles
+   */
+  [[nodiscard]] size_t
+  NumCycleDetectionAttempts()
```
phate, line 141: `NumCycleDetectionAttemps() const noexcept`

jlm/llvm/opt/alias-analyses/LazyCycleDetection.hpp
```diff
@@ -0,0 +1,187 @@
...
+  {
+    return NumCycleDetectAttempts_;
+  }
+
+  /**
+   * @return the number of cycles detected by Lazy cycle detection
+   */
+  [[nodiscard]] size_t
+  NumCyclesDetected()
```
phate, line 150: `NumCyclesDetected() const noexcept`

jlm/llvm/opt/alias-analyses/LazyCycleDetection.hpp
```diff
@@ -0,0 +1,187 @@
...
+  {
+    return NumCyclesDetected_;
+  }
+
+  /**
+   * @return the number of unifications made while eliminating found cycles
+   */
+  [[nodiscard]] size_t
+  NumCycleUnifications()
```
phate, line 159: same here

jlm/llvm/opt/alias-analyses/LazyCycleDetection.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+  // A set of all checked simple edges first -> second, to avoid checking again
+  util::HashSet<std::pair<PointerObjectIndex, PointerObjectIndex>> CheckedEdges_;
+
+  // The dfs stack, which may contain the same node multiple times
+  std::stack<PointerObjectIndex> DfsStack_;
+  // Possible states of nodes during the DFS
+  static constexpr uint8_t NodeStateNotVisited = 0;
+  static constexpr uint8_t NodeStateVisited = 1;
+  static constexpr uint8_t NodeStateLeft = 2;
```
phate, line 177: What does the state "NodeStateLeft" mean? Left as in opposite of right, or as in remaining?

tests/jlm/llvm/opt/alias-analyses/TestLazyCycleDetection.cpp
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright 2023, 2024 Håvard Krogstie <krogstie.havard@gmail.com>
+ * See COPYING for terms of redistribution.
+ */
+
+#include "TestRvsdgs.hpp"
```
phate, line 6: `#include <TestRvsdgs.hpp>`

tests/jlm/llvm/opt/alias-analyses/TestLazyCycleDetection.cpp
```diff
@@ -0,0 +1,110 @@
...
+  {
+    (void)set.CreateDummyRegisterPointerObject();
+  }
+
+  // Create a graph that looks like
+  //   --> 1 --> 2 --> 3
+  //  /          |
+  // 0           |
+  //  \          V
+  //   --> 5 --> 4
```
phate, line 35: Thank you!

phate (changes requested):

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
haved: Yes, they can be combined. They are also from the same paper (Hardekopf 2007)

haved:
@phate I fixed your comments. I replaced "left" with "popped", as that is much more descriptive.

phate (approved):

PR #583, opened 2024-08-17
Add copy() method to result class

phate:
This PR does the following:

1. Introduces a Copy() method to the result class.
2. The introduced method enables us to fix a bug in the copy() method of
the region. Previously, results were not copied according to their
result subtypes, but simply a new instance of result was created.
3. Extend unit test.

This PR is a necessary step to get rid off ports.


caleridas:
The semantic of the "copy" method is a bit weird TBH -- I would normally understand that a "copy" operation creates a (free-standing) instance of the original object that can then later be "attached" somewhere. However, in this case it attaches the object directly to its new destination. Maybe "CopyTo"?
I understand the reason for that, namely that the "create" method does not just "create" but "create and attach" (which is actually a little bit convoluted in itself).

caleridas (approved):
Other than comment about API/name (which isn't an essential piece), looks good to me.

PR #585, opened 2024-08-20
Remove expport class

phate:
This PR does the following:

1. Introduces the GraphExport class and its subclasses to model the export
of internal entities from the module. These classes serve as a replacement for the old
expport class.
2. Replaces all usages of expport with GraphExport
3. Removes old expport classes.

This is one of the steps necessary in order to completely remove ports
from the code base.

### DIFF
--- a/jlm/rvsdg/graph.cpp
+++ b/jlm/rvsdg/graph.cpp
@@ -27,26 +27,6 @@ GraphExport::GraphExport(rvsdg::output & origin, std::string name)
       Name_(std::move(name))
 {}
 
-/* expport */
-
-expport::~expport()
-{}
-
-bool
-expport::operator==(const port & other) const noexcept
-{
-  auto p = dynamic_cast<const expport *>(&other);
-  return p && p->type() == type() && p->name() == name();
-}
-
-std::unique_ptr<port>
-expport::copy() const
-{
-  return std::unique_ptr<port>(new expport(*this));
-}
-
-/* graph */
-
 graph::~graph()
 {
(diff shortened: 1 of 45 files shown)


haved (approved):
Looks good, I like the new interface much better!

One question: should the constructor of `GraphExport` make any attempt at asserting that the region it is being added to is a root region, and not, e.g., a theta or lambda region?

jlm/rvsdg/graph.cpp
```diff
@@ -22,26 +22,11 @@ GraphImport::GraphImport(
       Name_(std::move(name))
 {}
 
-/* expport */
-
-expport::~expport()
+GraphExport::GraphExport(rvsdg::output & origin, std::string name)
+    : result(origin.region()->graph()->root(), &origin, nullptr, origin.Type()),
+      Name_(std::move(name))
```
phate, line 27: @haved The constructor explicitly takes the root region and hands it over to the origin. I do not see where you want to put the assert?

haved: Ah, sorry I didn't catch that. So if you try to make a `GraphExport` for an output that is not in the root region, a different assert will complain that the output and result belong to different regions.

It all looks good, go ahead and merge, @phate

phate: @haved Yes, exactly. If the origin's region is not the root region, then an exception should be thrown [here](https://github.com/phate/jlm/blob/d77c368689d5cce5ac8b43e28ded9dca1b51bd21/jlm/rvsdg/node.cpp#L33C1-L34C55).

PR #590, opened 2024-08-22
[AndersenAgnostic] Add the topological worklist solver to Andersen

haved:
Adds the dummy worklist, which only remembers if /anything/ has been pushed.

When given this "worklist", the solver instead performs topological sweeps across all work items.

### DIFF
--- a/jlm/llvm/frontend/ControlFlowRestructuring.cpp
+++ b/jlm/llvm/frontend/ControlFlowRestructuring.cpp
@@ -496,7 +496,7 @@ RestructureControlFlow(llvm::cfg * cfg)
 
   for (const auto & l : tcloops)
     reinsert_tcloop(l);
-  
+
   JLM_ASSERT(is_proper_structured(*cfg));
 }
 
--- a/jlm/llvm/frontend/ControlFlowRestructuring.cpp
+++ b/jlm/llvm/frontend/ControlFlowRestructuring.cpp
@@ -496,8 +496,7 @@ RestructureControlFlow(llvm::cfg * cfg)
 
   for (const auto & l : tcloops)
     reinsert_tcloop(l);
-
-  // TODO: SLOW
+  
   JLM_ASSERT(is_proper_structured(*cfg));
 }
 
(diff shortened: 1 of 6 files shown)


jlm/util/Worklist.hpp
```diff
@@ -294,6 +294,62 @@ class TwoPhaseLrfWorklist final : public Worklist<T>
...
 
+/**
+ * A fake worklist that only holds a single bit of information:
+ *  "Has any item been pushed since the last reset?"
+ * Used to implement the Topological worklist policy, which is not technically a worklist policy
+ * @tparam T the type of the work items.
+ * @see Worklist
+ */
+template<typename T>
+class DummyWorklist final : public Worklist<T>
```
phate, line 305: Could we name this differently? Yes, it is a dummy worklist, but it is used for a specific purpose. Could we get a name that hints at this purpose (checking whether something was pushed since last reset).

jlm/util/Worklist.hpp
```diff
@@ -294,6 +294,62 @@ class TwoPhaseLrfWorklist final : public Worklist<T>
...
+{
+public:
+  ~DummyWorklist() override = default;
+
+  DummyWorklist() = default;
+
+  bool
+  HasMoreWorkItems() override
```
phate, line 313: The name indicates that this should also be `const` maybe even `noexcept`, but the base class does not have it (and there might be a reason why it is not).

phate (approved):

haved:
@phate I struggled a bit to come up with a more descriptive name. I called it `ObserverWorklist`, but am open suggestions

jlm/llvm/frontend/ControlFlowRestructuring.cpp
```diff
@@ -497,6 +497,7 @@ RestructureControlFlow(llvm::cfg * cfg)
   for (const auto & l : tcloops)
     reinsert_tcloop(l);
 
+  // TODO: SLOW
```
phate, line 500: Noise (and yes, you are right, it is very slow)

phate (changes requested):
Introduced some noise with the last push.

jlm/llvm/frontend/ControlFlowRestructuring.cpp
haved: Oups, yes that was not intentional

phate (approved):

PR #594, opened 2024-08-23
Add EntryArgument for HLS loop

phate:
This PR adds an argument subclass for the entry arguments of the HLS loop. This is necessary in order to make the `rvsdg::argument` class abstract.


haved (approved):
Looks good to me

sjalander (approved):

PR #596, opened 2024-08-23
Adds MLIR backend information to README

sjalander:


phate (approved):

README.md
```diff
@@ -20,7 +24,7 @@ Regionalized Value State Dependence Graph (RVSDG) as intermediate representation
 make all
 ```
 
-This presumes that llvm-config-17 can be found in $PATH. If that is not the case,
+This presumes that llvm-config-18 can be found in $PATH. If that is not the case,
```
sjalander, line 27: We had missed that 17 was used also for the llvm-config

README.md
```diff
@@ -76,9 +80,22 @@ and the build has to be configured accordingly. A change of build configuration
 stale intermediate files first, i.e., run 'make clean'.
 CIRCT and the HLS backend can be setup with the following commands:
 ```
-./scripts/build-circt.sh --build-path <CIRCT-build-path> --install-path <path-to-CIRCT>
+./scripts/build-circt.sh
+
+./configure --enable-hls
```
sjalander, line 85: Simplified the help text to rely on the default.

README.md
phate: Maybe you can reformulate it such that we do not mention the version there. Suggestion: "This presumes that the right version of llvm-config can be found ..."

sjalander: Changed to suggestion

phate (approved):

PR #598, opened 2024-08-25
Skip predecessor basic blocks that are unreachable when converting LLVM phi instructions

### LINKED ISSUE #597: LlvmModuleConversion does not handle basic blocks with no predecessors
I am not sure why, but the clang frontend sometimes generates basic blocks with no predecessors. An example of this can be seen when compiling `557.xz`'s, `liblzma_lzma_lzma_decoder.c`.

This it not an issue in and of itself, and seems to be "normal". The produced LLVM IR contains a helpful `No predecessors!`.

It becomes an issue to `jlm-opt` only after I pass the LLVM IR through `opt --passes=mem2reg`, as that introduces a `phi` instruction where one of the predecessor BBs is "dead" (has no predecessors).

A basic block with no predecessors does not get visited by the `::llvm::ReversePostOrderTraversal`, so it never gets added to the `basic_block_map`.

This causes `jlm-opt` to assert false later in `patch_phi_operands` in `LlvmModuleConversion.cpp`, as it expects all the phi-instruction's predecessors to have been visited.

Here is a minimal example causing this assert to trip:
```llvm
; ModuleID = 'source.ll'
source_filename = "source.c"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-f80:128-n8:16:32:64-S128"
target triple = "x86_64-pc-linux-gnu"

; Function Attrs: noinline nounwind sspstrong uwtable
define dso_local i32 @test(i32 noundef %0) #0 {
  br label %2

2:                                                ; preds = %1
  %3 = sdiv i32 %0, 2
  br label %5

4:                                                ; No predecessors!
  br label %5

5:                                                ; preds = %2, %4
  %.0 = phi i32 [ %3, %2 ], [ poison, %4 ]
  ret i32 %.0
}

attributes #0 = { noinline nounwind sspstrong uwtable "frame-pointer"="all" "min-legal-vector-width"="0" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+cmov,+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "tune-cpu"="generic" }

!llvm.module.flags = !{!0, !1, !2, !3, !4}
!llvm.ident = !{!5}

!0 = !{i32 1, !"wchar_size", i32 4}
!1 = !{i32 8, !"PIC Level", i32 2}
!2 = !{i32 7, !"PIE Level", i32 2}
!3 = !{i32 7, !"uwtable", i32 2}
!4 = !{i32 7, !"frame-pointer", i32 2}
!5 = !{!"clang version 18.1.8"}
```

haved:
As detailed in #597, the `::llvm::ReversePostOrderTraversal` does not visit unreachable LLVM basic blocks.

As these BBs are never reached at runtime, and do not exist in our tac, this PR makes phi nodes ignore all operands from these basic blocks. If a phi node is left with only none operand, it is not converted to a `phi_op`, but instead uses the value directly.

Fixes #597, (the minimal example given there now compiles), but there are still issues with the SPEC benchmark `577.xz`.


phate (commented):
A few thoughts:

1. It would be nice to get at least one unit test on this.
2. What about phi nodes where the first basic block is alive, and the second one is dead? You would create a phi node in jlm with only a single operand (which makes no sense).
3. What about phi nodes where you have more than 2 operands/basic blocks (where some can be alive and others dead)?

Example:

Entry -> A
A -> D
B -> D
C -> D
D -> Exit

I think there is two ways out of this:

1. We eliminate unreachable basic blocks in LLVM before we start translating to jlm.
2. We "fix" the function before we start translating and then just translate it (with unreachable basic blocks). Unreachable basic blocks violate one of the assumptions of the control flow graph restructuring algorithm: all basic blocks should be dominated by the entry node. One way to go about this:

a. Find all basic blocks that are not dominated by other basic blocks and the function entry (note that this is different from an unreachable basic block).
b. Insert a dummy basic block at the beginning that points to the entry block and all the basic blocks found above.
c. Translate to jlm

I favor the second approach.

haved:
We risk getting a phi node with a single input, yes, at least in the minimal example I created for #597. I agree we should replace it fully.

This PR currently causes `has_valid_phis` to fail in some cases, so I have to figure out why.

I think it would make sense to include the minimal example from #597 as an "end-to-end-test", like we do with the C tests. Is there any scaffolding in place for testing LLVM IR files, or only for C files? It is still a bit of a mystery to me why `clang` ever generates LLVM IR with trivially unreachable basic blocks, so it is not so easy to create a C test for this.

phate:
@haved You typed a reply while I updated my comment. Sorry, I was a little bit trigger happy with pressing comment.

phate:
> I think it would make sense to include the minimal example from https://github.com/phate/jlm/issues/597 as an "end-to-end-test", like we do with the C tests. Is there any scaffolding in place for testing LLVM IR files, or only for C files? It is still a bit of a mystery to me why clang ever generates LLVM IR with trivially unreachable basic blocks, so it is not so easy to create a C test for this.

We can create such a test manually for the translation. This should not be a huge problem. An alternative would be to use LLVM lit, but we would need to invest in the infrastructure for this first.

haved:
@phate I agree that these CFGs break an important assumption, but I don't think we necessarily need to make them reachable from a dummy node to handle them cleanly.

By definition, the only thing that references an unreachable BB is a phi node, or another unreachable BB. The conversion algorithm can pretend they do not exist, as long as the phi nodes handle them during conversion. If the phi has more than one reachable BB, it remains a phi node. Otherwise it can just become a reference to the single possible value it may have at runtime.

(Due to loops it might be easier to not fully remove the phi node, but instead replace it by a dummy register renaming operation, if those exist. The worry is that two phi nodes can reference each other's results, which makes eliding them one at a time a bit iffy)

One nice property is that if one phi node can be elided like this, all phi nodes in the BB can be elided, so there should be no issues with non-phi instructions being ordered above phis.

haved:
@phate I implemented my suggested fix to this issue, which as you can see did not require that much code. Most of the added lines are comments (and the test of course).

Two notes:
 - Previously, phi instructions converted to tacs would be appended to the top of the BB, while all other tacs were appended to the bottom. Since LLVM phi nodes are always at the top anyway, there should be no reason to do this. I now append all tacs to the bottom, and everything is still green.
 - I fixed a small and not relevant mistake in `test-select.cpp` as I used it for inspiration and noticed the mistake. This change is "noise", but it is also so small that giving it a separate PR felt silly.

haved:
There is still something preventing 557.xz from being compiled though `jlm-opt`, but I am not looking into that until I have done the solve-without-flags we talked about

jlm/llvm/frontend/LlvmInstructionConversion.cpp
```diff
@@ -677,11 +677,49 @@ convert_store_instruction(::llvm::Instruction * i, tacsvector_t & tacs, context
   return nullptr;
 }
 
+/**
+ * Given an LLVM phi instruction, checks if the instruction has only one predecessor basic block
+ * that is reachable (i.e. there exists a path from the entry point to the basic block).
```
sjalander, line 682: Missing a comma after (i.e.

tests/jlm/llvm/frontend/llvm/TestLlvmPhiConversion.cpp
```diff
@@ -0,0 +1,256 @@
...
+br label %bb2
+
+bb5:
+; Here x was even, no need to calculate anything
+br label %bb2
+
+bb6:
+ret i64 %4
+}
+)"""";
```
phate, line 94: This is going to be a nightmare to maintain if the syntax changes. Why not using the LLVM API to setup the test?

tests/jlm/llvm/frontend/llvm/TestLlvmPhiConversion.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+bb6: ; No predecessors
+  br label %bb7
+
+bb7:
+  %7 = phi i64 [%6, %bb5], [poison, %bb6]
+  %8 = mul i64 %7, 10
+  ret i64 %8
+}
+)"""";
```
phate, line 201: Same as above.

phate (commented):
I am a little bit concerned about the robustness of this approach as pointed out in the comments, and some maintainability issues. Maybe we can have a little chat about in the next meeting before we merge it?

phate:
Please do not forget to update the PR message to reflect the new state.

tests/jlm/llvm/frontend/llvm/TestLlvmPhiConversion.cpp
haved: @phate It could be done with the API as well, but I am not sure it is any more likely that the LLVM IR syntax changes, than that the Builder API changes. I think they are around equally maintainable, but my first instinct was that using Builder is a lot more verbose. That might not actually be true, so I can change it to use the builder.

jlm/llvm/Makefile.sub
```diff
@@ -148,6 +148,7 @@ libllvm_TESTS += \
 	tests/jlm/llvm/frontend/llvm/test-export \
 	tests/jlm/llvm/frontend/llvm/TestFNeg \
 	tests/jlm/llvm/frontend/llvm/test-function-call \
+	tests/jlm/llvm/frontend/llvm/TestLlvmPhiConversion \
```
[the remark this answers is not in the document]
sjalander: @phate
Adding test to each file seems redundant as they are already in the test directory.
I suggest to start removing 'test' from the filenamens.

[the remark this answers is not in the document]
phate: No, please do not remove the test suffix. At some point, we want to move the tests into the library folders and get rid off the test subfolder. We need the suffix then.

haved:
@phate I implemented your comments, and changed the tests to use the `IRBuilder` API, which honestly was kind of nice to use, so I agree it should be used instead of the `parseIRModule` method.

phate (approved):

PR #600, opened 2024-08-27
Make argument class abstract

phate:
1. Replaces all usages of `argument::create()` with `TestGraphArgument::Create()` in the unit tests.
2. Makes the argument class abstract
3. Minor clean ups in the argument class
4. Add documentation to argument class


haved (approved):
Looks good, great to have it be abstract

PR #601, opened 2024-08-27
Add ThetaPredicateResult class

phate:
Add class for representing the predicate result of a theta node.


haved (approved):
Looks good.
I am noticing that when you create a node, the constructor of `rvsdg::node` is responsible for attaching the node to the region.

With results, you have to remember to attach it to the region afterward, despite the ctor taking the region as a parameter. I assume there might be a reason it does not attach to the region immediately?

phate:
@haved The reason is historical and some not-finished refactoring at some point. We should align both of them eventually.

PR #602, opened 2024-08-28
Add ExitResult for HLS loop

phate:
1. Add the ExitResult class for representing result values before the HLS loop terminates
2. Replaces respective `result::create()` with `ExitResult::Create()` usages


haved (approved):
Looks good. My only concern is that the name `ExitResult` does not convey that it belongs to loop nodes, but it is namespaced into `hls`, so it probably won't cause any confusion

phate:
@haved A renaming of the HLS inputs/outputs/arguments/results is necessary anyway.

PR #603, opened 2024-08-29
Add jlm-opt support for RVSDG tree annotations

phate:
This PR does the following:
1. Adds support for annotating the number of RVSDG nodes to the RVSDG tree in the RvsdgTreePrinter pass.
2. Adds support for controlling the annotations from jlm-opt
3. Adds a unit test for the printing of the introduced annotation

Usage: `jlm-opt --RvsdgTreePrinter -s /tmp --annotations=NumRvsdgNodes input.ll`

Example output:
```
RootRegion NumRvsdgNodes:2
-STRUCTURAL_TEST_NODE NumRvsdgNodes:2
--Region[0] NumRvsdgNodes:1
--Region[1] NumRvsdgNodes:1
```

### DIFF
--- a/jlm/llvm/opt/RvsdgTreePrinter.hpp
+++ b/jlm/llvm/opt/RvsdgTreePrinter.hpp
@@ -116,7 +116,7 @@ public:
 
 private:
   /**
-   * Computes a map with annotations based on the required \ref Annotation%s in the \ref
+   * Computes a map with annotations based on the required \ref jlm::util::Annotation%s in the \ref
    * Configuration for the individual regions and structural nodes of the region tree.
    *
    * @param rvsdg The RVSDG for which to compute the annotations.
--- a/jlm/llvm/opt/RvsdgTreePrinter.hpp
+++ b/jlm/llvm/opt/RvsdgTreePrinter.hpp
@@ -7,7 +7,14 @@
 #define JLM_LLVM_OPT_RVSDGTREEPRINTER_HPP
 
 #include <jlm/llvm/opt/optimization.hpp>
+#include <jlm/util/AnnotationMap.hpp>
 #include <jlm/util/file.hpp>
+#include <jlm/util/HashSet.hpp>
+
+namespace jlm::rvsdg
+{
+class graph;
+}
 
 namespace jlm::util
 {
@@ -34,8 +41,29 @@ public:
   class Configuration final
   {
   public:
-    explicit Configuration(const util::filepath & outputDirectory)
-        : OutputDirectory_(std::move(outputDirectory))
+    enum class Annotation
+    {
+      /**
+       * Must always be the first enum value. Used for iteration.
+       */
+      FirstEnumValue,
+
+      /**
+       * Annotate regions and structural nodes with the number of RVSDG nodes.
+       */
+      NumRvsdgNodes,
+
+      /**
+       * Must always be the last enum value. Used for iteration.
+       */
+      LastEnumValue
+    };
+
+    Configuration(
+        const util::filepath & outputDirectory,
+        util::HashSet<Annotation> requiredAnnotations)
+        : OutputDirectory_(std::move(outputDirectory)),
+          RequiredAnnotations_(std::move(requiredAnnotations))
     {
       JLM_ASSERT(outputDirectory.IsDirectory());
       JLM_ASSERT(outputDirectory.Exists());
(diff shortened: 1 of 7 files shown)


phate:
The next annotations on the TODO list:
1. NumMemoryStateArguments and NumMemoryStateResults
2. Number of topological orderings for nodes with state edges

The first annotations are for comparing agnostic vs. region-aware memory encoding. The second annotation is for estimating the number of  permutations we could get when performing control flow recovery.

phate:
The file written from the `PrintRvsdgTree()` test is all in a sudden `PrintTreeTest-rvsdgTree-1` instead of `PrintTreeTest-rvsdgTree-0`.

It seems that the two rvsdgModules (from the two RvsdgTreePrinter tests) are allocated at the same address, and with that they get the same counter in the RvsdgModuleCounterMap_ in the RvsdgTreePrinter. The easiest way would be to just not make this map static and be done with it.

haved:
@phate
Relying on the address of the module is a bit unfortunate, and I guess the allocator happily reuses the memory for the second module.

I didn't quite understand how you intended to fix it. Do you want each module to count printed trees individually, or will you have a global counter? Alternatively the static set could be keyed on the name of the module, as that is included in the filename of the output (?), making them unique

jlm/llvm/opt/RvsdgTreePrinter.hpp
```diff
@@ -34,8 +41,29 @@ class RvsdgTreePrinter final : public optimization
...
+       */
+      NumRvsdgNodes,
+
+      /**
+       * Must always be the last enum value. Used for iteration.
+       */
+      LastEnumValue
+    };
+
+    Configuration(
```
sjalander, line 62: Documentation?

jlm/llvm/opt/RvsdgTreePrinter.hpp
```diff
@@ -77,6 +115,12 @@ class RvsdgTreePrinter final : public optimization
   run(RvsdgModule & rvsdgModule);
 
 private:
+  [[nodiscard]] util::AnnotationMap
+  ComputeAnnotationMap(const rvsdg::graph & rvsdg) const;
```
sjalander, line 119: I get it that private methods doesn't contribute to the public API, but wouldn't it still be helpful with documentation for these?

jlm/llvm/opt/RvsdgTreePrinter.hpp
phate: It is the constructor of the Configuration class. What did you have in mind writing there as documentation? The only thing that pops up in my mind: "Creates an instance of the Configuration class". However, this applies for any constructor of a class and does not add any value. Do you want the parameters/attributes in the class documented?

jlm/llvm/opt/RvsdgTreePrinter.hpp
phate: Added documentation.

phate:
@haved I changed the map key to be the pointer of module + the source file name now. This is still not unambiguous, but should do the job for now. There would be some more work necessary in order to get a really unique key.

haved:
@phate
But does the key necessarily need to be unique? Since the file name of the output only depends on the module's name (+ suffix), it kind of makes sense to only use that as the key, and accept that the counting suffix keeps increasing if the same invocation of `jlm-opt`or `jlc` keeps compiling modules with the same name.

phate:
The idea was that there is a unique counter per compilation unit (i.e. RVSDG module / input file) that starts at 0 and increases with every region tree printed. Thus, for a single input.ll, one would get input-regionTree-[0 ...].

However, for that to work one needs to uniquely identify the input file, but I do not really have this information in the RVSDG module (the source file name property is not enforced, neither by LLVM nor us).

Thus, maybe it would be just simpler to have a global counter that simple just keeps counting upwards with every region tree written out!?

haved:
@phate I would be ok with a global counter. It is not common to be compiling multiple modules per execution. The only situation I can think of where it currently comes up is when a test file contains multiple tests with a module in each. If the test wants to read the file, it helps to know that each file is called `uniqueTest-regionTree-0`. This is why I suggested using the module name as a key per counter. If each test function uses a unique name, we know that they will all get the suffix `-0` regardless of execution order.

phate:
@haved I only use the source filename as key now. I hope this will suffice for a while.

haved (approved):

phate:
@sjalander I was not really sure how to attend to one of your comments. Could you please elaborate such that I get this PR merged?

jlm/llvm/opt/RvsdgTreePrinter.hpp
sjalander: My bad, shouldn't have commented on this one.

sjalander (approved):

PR #604, opened 2024-08-30
Make result class abstract

phate:
1. Replaces all usages of `result::create()` in the unit tests with the appropriate subclasses.
2. Makes the result class abstract
3. Minor clean ups in the result class
4. Add documentation to result class

### DIFF
--- a/jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
@@ -109,12 +109,9 @@ RemoveUnusedStatesFromLambda(llvm::lambda::node & lambdaNode)
 
   JLM_ASSERT(lambdaNode.output()->nusers() == 1);
   lambdaNode.region()->RemoveResult((*lambdaNode.output()->begin())->index());
+  auto oldExport = lambdaNode.ComputeCallSummary()->GetRvsdgExport();
+  jlm::llvm::GraphExport::Create(*newLambdaOutput, oldExport ? oldExport->Name() : "");
   remove(&lambdaNode);
-  jlm::rvsdg::result::create(
-      newLambda->region(),
-      newLambdaOutput,
-      nullptr,
-      newLambdaOutput->Type());
 }
 
 static void
--- a/jlm/rvsdg/region.hpp
+++ b/jlm/rvsdg/region.hpp
@@ -87,24 +87,33 @@ private:
   structural_input * input_;
 };
 
+/**
+ * \brief Represents the result of a region.
+ *
+ * Region results represent the final values of the region's acyclic graph. The result values
+ * can be mapped back to the region arguments or the corresponding structural outputs
+ * throughout the execution, but the concrete semantics of this mapping
+ * depends on the structural node the region is part of. A region result is either linked
+ * with a \ref structural_output or is a standalone result.
+ */
 class result : public input
 {
-  jlm::util::intrusive_list_anchor<jlm::rvsdg::result> structural_output_anchor_;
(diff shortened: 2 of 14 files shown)


jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
```diff
@@ -109,12 +109,9 @@ RemoveUnusedStatesFromLambda(llvm::lambda::node & lambdaNode)
 
   JLM_ASSERT(lambdaNode.output()->nusers() == 1);
   lambdaNode.region()->RemoveResult((*lambdaNode.output()->begin())->index());
+  auto oldExport = lambdaNode.ComputeCallSummary()->GetRvsdgExport();
+  jlm::llvm::GraphExport::Create(*newLambdaOutput, oldExport ? oldExport->Name() : "");
```
sjalander, line 113: The GraphExport moves functionality from rvsdg to llvm, making the hls dependence on llvm more explicit. But I guess there's no benefit of trying to keep it more decoupled, as it anyway depends on the lambda.

This made me think about the naming conventions that are used.

The hls backend, of course, takes rvsdg-llvm as input and has a direct dependence on it. This is not reflected in the rvsdg2rhls naming. It maybe would be more logical to call it llvm2rhls, but that has the disadvantage that it can sound like it's llvm-IR and not rvsdg-llvm.
Potentially more informative naming could be: rvsdg-llvm2rvsdg-hls, r-llvm2r-hls, rllvm2rhls.

Similar naming ambiguity exists in the llvm backend with rvsdg2jlm and jlm2llvm, which could sound like the generic rvsdg 'dialect' (jlm/rvsdg) is taken as input. It also makes me wonder why jlm is the name of the intermediate step, as this is the name of the whole compiler. Wouldn't a different, more descriptive, name be suitable here?

jlm/rvsdg/region.hpp
```diff
@@ -87,24 +87,33 @@ class argument : public output
...
-  virtual ~result() noexcept;
+  ~result() noexcept override;
 
 protected:
   result(
-      jlm::rvsdg::region * region,
-      jlm::rvsdg::output * origin,
-      jlm::rvsdg::structural_output * output,
+      rvsdg::region * region,
+      rvsdg::output * origin,
```
sjalander, line 112: Why is 'rvsdg::' kept for 'region' and 'output', but not for 'structural_output'?
Are they used in other namespaces that cause conflicts?

haved (approved):
The changes look nice.
I agree with @sjalander's comments, but this PR can be merged first, and then we can figure out if we want to rename some of the stages of the compiler. I never quite understood why it is called `rvsdg2jlm`.

jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
phate: Hls is dependent on the LLVM dialect anyway. It would not work without it in the first place. Thus, this change does not introduce any unnecessary dependencies (that are not already there and required).

All these "sub-"namespaces should disappear and they class/methods should express their name properly. However, this is for another PR.

jlm/rvsdg/region.hpp
phate: There is no conflict in terms of namespaces, but just potential confusion between type "output" and variable "output", if the namespace is removed. The problem is that the classes/methods have not been renamed to the CamelCase equivalent. That is why we still have this "problem". It will go away once we perform the renaming.

phate:
@sjalander @haved I was more conservative with the  export name in the HLS part of the code now, which fixed the problem. Could you please have a look again and approve it is acceptable.

sjalander (approved):

PR #605, opened 2024-08-31
Ensure jlm-opt creates only single instances of optimizations

phate:
PR #593 made optimizations dynamically configurable, but also screwed up the invariant that an optimization was only created once. This PR restores this invariant and moves the optimization creation from the JlmOptCommandLineOptions class to JlmOptCommand class.

### DIFF
--- a/jlm/tooling/Command.cpp
+++ b/jlm/tooling/Command.cpp
@@ -9,7 +9,22 @@
 #include <jlm/llvm/frontend/LlvmModuleConversion.hpp>
 #include <jlm/llvm/ir/ipgraph-module.hpp>
 #include <jlm/llvm/ir/RvsdgModule.hpp>
+#include <jlm/llvm/opt/alias-analyses/AgnosticMemoryNodeProvider.hpp>
+#include <jlm/llvm/opt/alias-analyses/Andersen.hpp>
+#include <jlm/llvm/opt/alias-analyses/Optimization.hpp>
+#include <jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.hpp>
+#include <jlm/llvm/opt/alias-analyses/Steensgaard.hpp>
+#include <jlm/llvm/opt/cne.hpp>
+#include <jlm/llvm/opt/DeadNodeElimination.hpp>
+#include <jlm/llvm/opt/inlining.hpp>
+#include <jlm/llvm/opt/InvariantValueRedirection.hpp>
+#include <jlm/llvm/opt/inversion.hpp>
 #include <jlm/llvm/opt/OptimizationSequence.hpp>
+#include <jlm/llvm/opt/pull.hpp>
+#include <jlm/llvm/opt/push.hpp>
+#include <jlm/llvm/opt/reduction.hpp>
+#include <jlm/llvm/opt/RvsdgTreePrinter.hpp>
+#include <jlm/llvm/opt/unroll.hpp>
 #include <jlm/rvsdg/view.hpp>
 #include <jlm/tooling/Command.hpp>
 #include <jlm/tooling/CommandPaths.hpp>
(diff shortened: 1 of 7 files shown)


jlm/tooling/CommandLine.hpp
```diff
@@ -160,7 +160,7 @@ class JlmOptCommandLineOptions final : public CommandLineOptions
   static const char *
   ToCommandLineArgument(OutputFormat outputFormat);
 
-  [[nodiscard]] std::unique_ptr<llvm::optimization>
+  [[nodiscard]] llvm::optimization *
```
haved, line 163: If this method uses a static map internally, I think the method itself should also be static, to avoid accidentally using state

haved (commented):
Nice pattern to lazily initialize singletons, though I think we might be able to make the "staticness" even more explicit?

jlm/tooling/CommandLine.hpp
phate: I do not fully see how to do this (without some major refactoring). This method calls `CreateOptimization()`, which reads internal state. Thus, I cannot just make this method static as the `CreateOptimization()` method needs the state.

haved: @phate
If the created optimizations rely on state, that is an argument against including them in a global map, and not per-class map. The next instance of `JlmCommandLineOptions` risks returning optimizations initialized with the wrong data.

That being said, almost all the optimizations are "stateless" when they are not running, so it is mostly not an issue (unless we start running multiple JlmOpt commands in parallel in the same binary any time soon).

The fact that some optimizations rely on state / have a reference to a per `JlmCommandLineOptions` configuration, makes me prefer a solution where it is possible to create separate instances of optimizations (if only those with state). This would also allow the Tree dumper to have its own counter, for instance. Making the map a field instead of a global variable would achieve that, at the expense of making multiple instances of some singletons in very specific circumstances.

phate: @haved I thought about this a bit more and realized that the JlmCommandLineOption class is the wrong to place to have these optimizations in the first place. I moved the entire optimization creation into the JlmOptCommand class now. There are no static variables used any longer. It is much better now (in my opinion). Thanks for being persistent!

jlm/tooling/Command.cpp
```diff
@@ -270,6 +285,19 @@ LlcCommand::ToString(const RelocationModel & relocationModel)
...
 
+JlmOptCommand::JlmOptCommand(
+    std::string programName,
+    const jlm::tooling::JlmOptCommandLineOptions & commandLineOptions)
+    : ProgramName_(std::move(programName)),
+      CommandLineOptions_(std::move(commandLineOptions))
+{
+  for (auto optimizationId : CommandLineOptions_.GetOptimizationIds())
+  {
+    if (auto it = Optimizations_.find(optimizationId); it == Optimizations_.end())
```
haved, line 296: This initialization could have been done lazily in `GetOptimizations()`, is there a reason it is not?

haved (approved):
Looks good, I just have one question regarding the eager initialization in the ctor, instead of doing a bit more lazy approach

jlm/tooling/Command.cpp
phate: @haved The `Run()` method of the `JlmOptCommand` (or of any `Command` class) is marked as `const`. Thus, it is not permitted to modify the instance in this method. However, this method also invokes the `GetOptimizations()` method, which means that the `GetOptimization()` method also needs to be `const`. Thus, I could not have created and stored the optimizations in the `GetOptimization()` method. In the end, it does not make much of a difference whether the optimizations are initialized lazily or eagerly.

haved: Ah that's why. Resolving @phate

PR #606, opened 2024-09-04
Add fp128

halvorlinder:
Tests involving the long double type would fail on ARM processors, because of a discrepancy in bit-length. A new fpsize is now added for ARM.

### DIFF
--- a/tests/jlm/llvm/frontend/llvm/LlvmTypeConversionTests.cpp
+++ b/tests/jlm/llvm/frontend/llvm/LlvmTypeConversionTests.cpp
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2024 Halvor Linder Henriksen <halvor_lh@hotmail.no>
+ * See COPYING for terms of redistribution.
+ */
+
+#include <test-registry.hpp>
+#include <test-util.hpp>
+
+#include <jlm/llvm/frontend/LlvmModuleConversion.hpp>
+#include <jlm/llvm/ir/operators/Load.hpp>
+#include <jlm/llvm/ir/operators/operators.hpp>
+#include <jlm/llvm/ir/print.hpp>
+
+#include <llvm/IR/BasicBlock.h>
+#include <llvm/IR/IRBuilder.h>
+#include <llvm/IR/Module.h>
+
+static void
+TestTypeConversion(
+    llvm::LLVMContext & context,
+    llvm::Type * llvm_type,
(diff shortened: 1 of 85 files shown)


tests/jlm/llvm/frontend/llvm/LlvmTypeConversionTests.cpp
```diff
@@ -0,0 +1,110 @@
...
+  llvm::LLVMContext context;
+
+  TestTypeConversion(context, ::llvm::Type::getHalfTy(context), jlm::llvm::fpsize::half);
+  TestTypeConversion(context, ::llvm::Type::getFloatTy(context), jlm::llvm::fpsize::flt);
+  TestTypeConversion(context, ::llvm::Type::getDoubleTy(context), jlm::llvm::fpsize::dbl);
+  TestTypeConversion(context, ::llvm::Type::getX86_FP80Ty(context), jlm::llvm::fpsize::x86fp80);
+  TestTypeConversion(context, ::llvm::Type::getFP128Ty(context), jlm::llvm::fpsize::fp128);
+
+  return 0;
+}
```
phate, line 106: I appreciate that you wrote the test, but this does not only test the type conversion, but a lot of other stuff with it as well. It should really only test the type conversion itself:

```
#include <jlm/llvm/frontend/LlvmTypeConversion.hpp>

auto llvmType = ::llvm::Type::getHalfTy(context);
... // setup other LLVM types

auto jlmType = jlm::llvm::ConvertType(llvmType, ...);
... // convert other LLVM types

auto floatingPointType = dynamic_cast<const fptype*>(jlmType.get());
assert(floatingPointType && floatingPointeType->size() == fpsize::half);
... // assert other LLVM type conversions
```

The same pattern for the backend type conversion.

You can skip the tests for the instruction conversion (this is covered by the LLVM test suite tests etc.).

phate (changes requested):

tests/jlm/llvm/frontend/llvm/LlvmTypeConversionTests.cpp
halvorlinder: I have made the requested changes, but I did not quite understand which conversions I should test, did you only want them for the floating point types?

phate: Looks good. You only touched the floating point types, so it is fine that you only add tests for those. Thanks a lot!

phate (approved):

PR #608, opened 2024-09-05
Rename result class to RegionResult

phate:


haved (dismissed):
Very nice, this makes it much clearer for new readers of the codebase what is going on

sjalander (approved):

PR #616, opened 2024-09-10
Add missing libllvm headers

phate:
1. Add several headers that were missing in the Makefile
2. Sort header files alphabetically


jlm/llvm/Makefile.sub
```diff
@@ -65,70 +65,77 @@ libllvm_SOURCES = \
     jlm/llvm/opt/unroll.cpp \
 
 libllvm_HEADERS = \
```
phate, line 67: Holding this PR back until @haved merged the dot backend as it touches the files in the Makefile. I rather deal with the conflict here.

phate:
@haved @sjalander This PR is ready for review.

sjalander (commented):
I haven't thought this through but could we add a CI test that checks for this.
Ideally, it should check that all listed .hpp files exist and that none is missing.

jlm/llvm/Makefile.sub
```diff
@@ -66,71 +66,78 @@ libllvm_SOURCES = \
...
-	jlm/llvm/ir/operators/lambda.hpp \
-	jlm/llvm/ir/aggregation.hpp \
-	jlm/llvm/backend/dot/DotWriter.hpp \
-	jlm/llvm/backend/rvsdg2jlm/rvsdg2jlm.hpp \
-	jlm/llvm/backend/rvsdg2jlm/context.hpp \
-	jlm/llvm/backend/jlm2llvm/jlm2llvm.hpp \
-	jlm/llvm/backend/jlm2llvm/type.hpp \
-	jlm/llvm/backend/jlm2llvm/instruction.hpp \
-	jlm/llvm/backend/jlm2llvm/context.hpp \
+    jlm/llvm/backend/dot/DotWriter.hpp \
```
sjalander, line 69: Do we need to list them individually or could we simply use
jlm/llvm/backend/dot/*.hpp

phate: I do not know. @caleridas Could you help me out here, please?

phate:
@sjalander I created issue #617 for the CI request.

jlm/llvm/Makefile.sub
phate: Neither I am unsure whether using wildcards here would be a good idea....

caleridas: would prefer to list them explicitly, and certainly a build rule to check for consistency would be in order -- can take a look at that

using wildcards... instinctively I would prefer not to as it also tends to cause problems with spurious dependencies e.g. with files accidentally left in the build tree -- also, when "installing" (=exporting to other projects) files, it is essential that the correct files are selected

phate:
I will close this PR as it was already handled in #619

PR #618, opened 2024-09-12
Add NumMemoryStateInputsOutputs annotation to RvsdgTreePrinter

phate:
1. Extends the structural test node with an interface for adding inputs/outputs/arguments/results
2. Add support for printing of memory state typed inputs/outputs to RvsdgTreePrinter class
3. Adds support for this annotation to jlm-opt

### DIFF
--- a/jlm/llvm/opt/RvsdgTreePrinter.cpp
+++ b/jlm/llvm/opt/RvsdgTreePrinter.cpp
@@ -123,10 +123,10 @@ RvsdgTreePrinter::AnnotateNumMemoryStateInputsOutputs(
     const rvsdg::graph & rvsdg,
     util::AnnotationMap & annotationMap)
 {
-  static std::string_view argumentLabel("NumMemoryStateTypeArguments");
-  static std::string_view resultLabel("NumMemoryStateTypeResults");
-  static std::string_view inputLabel("NumMemoryStateTypeInputs");
-  static std::string_view outputLabel("NumMemoryStateTypeOutputs");
+  std::string_view argumentLabel("NumMemoryStateTypeArguments");
+  std::string_view resultLabel("NumMemoryStateTypeResults");
+  std::string_view inputLabel("NumMemoryStateTypeInputs");
+  std::string_view outputLabel("NumMemoryStateTypeOutputs");
 
   std::function<void(const rvsdg::region &)> annotateRegion = [&](const rvsdg::region & region)
   {
--- a/jlm/llvm/opt/RvsdgTreePrinter.cpp
+++ b/jlm/llvm/opt/RvsdgTreePrinter.cpp
@@ -74,6 +74,9 @@ RvsdgTreePrinter::ComputeAnnotationMap(const rvsdg::graph & rvsdg) const
     case Configuration::Annotation::NumRvsdgNodes:
       AnnotateNumRvsdgNodes(rvsdg, annotationMap);
       break;
+    case Configuration::Annotation::NumMemoryStateInputsOutputs:
+      AnnotateNumMemoryStateInputsOutputs(rvsdg, annotationMap);
+      break;
     default:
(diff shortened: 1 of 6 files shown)


jlm/llvm/opt/RvsdgTreePrinter.cpp
```diff
@@ -115,6 +118,78 @@ RvsdgTreePrinter::AnnotateNumRvsdgNodes(
...
+        annotationMap.AddAnnotation(structuralNode, { outputLabel, numMemoryStateOutputs });
+
+        for (size_t n = 0; n < structuralNode->nsubregions(); n++)
+        {
+          auto subregion = structuralNode->subregion(n);
+          annotateRegion(*subregion);
+        }
+      }
+    }
+  };
```
phate, line 188: There is a lot of repetition in here. However, in order to remove this, it would be necessary to add better support for iterators/iterator ranges to the base classes.

tests/test-operation.hpp
```diff
@@ -205,6 +205,10 @@ class structural_op final : public rvsdg::structural_op
   copy() const override;
 };
```
phate, line 207: The entire extension to the test node could be put in its own PR if desired.

jlm/llvm/opt/RvsdgTreePrinter.cpp
```diff
@@ -115,6 +118,78 @@ RvsdgTreePrinter::AnnotateNumRvsdgNodes(
...
+RvsdgTreePrinter::AnnotateNumMemoryStateInputsOutputs(
+    const rvsdg::graph & rvsdg,
+    util::AnnotationMap & annotationMap)
+{
+  static std::string_view argumentLabel("NumMemoryStateTypeArguments");
+  static std::string_view resultLabel("NumMemoryStateTypeResults");
+  static std::string_view inputLabel("NumMemoryStateTypeInputs");
+  static std::string_view outputLabel("NumMemoryStateTypeOutputs");
+
+  std::function<void(const rvsdg::region &)> annotateRegion = [&](const rvsdg::region & region)
```
haved, line 131: Is there a reason the lambda is stored as a function pointer in `std::function`, and not an `auto` functor?

haved: Ah, I see now it is recursive. Nice trick, I didn't know you could do recursive lambdas like that

jlm/llvm/opt/RvsdgTreePrinter.cpp
```diff
@@ -115,6 +118,78 @@ RvsdgTreePrinter::AnnotateNumRvsdgNodes(
...
 
+void
+RvsdgTreePrinter::AnnotateNumMemoryStateInputsOutputs(
+    const rvsdg::graph & rvsdg,
+    util::AnnotationMap & annotationMap)
+{
+  static std::string_view argumentLabel("NumMemoryStateTypeArguments");
+  static std::string_view resultLabel("NumMemoryStateTypeResults");
+  static std::string_view inputLabel("NumMemoryStateTypeInputs");
+  static std::string_view outputLabel("NumMemoryStateTypeOutputs");
```
haved, line 129: I checked compiler explorer, and it is better to leave these as non-static. It does not do a strlen at runtime either way, but making them static adds an extra level of indirection to access the length and string data.

haved (approved):
Looks good. My only comment is that we can save 16 bytes and some loads in the final binary by not making the `string_view`s static.

jlm/llvm/opt/RvsdgTreePrinter.cpp
phate: You answered it yourself. :)

jlm/llvm/opt/RvsdgTreePrinter.cpp
phate: I removed that `static`s. Could you approve again, please?

haved (approved):

PR #620, opened 2024-09-13
Add header check to CI

### LINKED ISSUE #617: Add CI test to ensure we have all headers listed
It happened already multiple times that we forgot to list headers in the Makefile. The latest incarnation of this problem is PR #616. It is time to have a basic check in the CI to ensure we always have all the headers listed.

phate:
Closes #617


caleridas (approved):
should have done that immediately...

Issue #586, opened 2024-08-20
Miscompilation of trivial loop after NodePushOut and NodePullIn


haved:
Given the following code
```c
int main() {
    for (int y = 0; y < 1; y++);
    return 0;
}
```

there is a miscompilation causing an infinite loop when compiled like so:
```
clang-17 src/main.c -o build/main.ll -S -emit-llvm -Xclang -disable-O0-optnone
opt-17 build/main.ll -o build/main-mem2reg.ll --passes=mem2reg -S
jlm-opt build/main-mem2reg.ll -o build/main-mem2reg-jlmopt.ll --NodePushOut
jlm-opt build/main-mem2reg-jlmopt.ll -o build/main-mem2reg-jlmopt-twice.ll --NodePullIn
clang-17 build/main-mem2reg-jlmopt-twice.ll -o build/jlm-opt-twice
```

The issue persists if breaking the for-loop into a while-loop.

The issue only appears when invoking `NodePushOut` and `NodePullIn` in two separate `jlm-opt`commands.

Removing any one of these passes causes the issue to disappear.

Compiling `main-mem2reg-jlmopt.ll` works, so the issue seams to appear in `NodePullIn`.

The `mem2reg` is also needed for the issue to appear. For reference, this is the content of `main-mem2reg.ll`:
<details>

```llvm
; ModuleID = 'build/main.ll'
source_filename = "src/main.c"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

; Function Attrs: noinline nounwind uwtable
define dso_local i32 @main() #0 {
  br label %1

1:                                                ; preds = %4, %0
  %.0 = phi i32 [ 0, %0 ], [ %5, %4 ]
  %2 = icmp slt i32 %.0, 1
  br i1 %2, label %3, label %6

3:                                                ; preds = %1
  br label %4

4:                                                ; preds = %3
  %5 = add nsw i32 %.0, 1
  br label %1, !llvm.loop !6

6:                                                ; preds = %1
  ret i32 0
}

[...]

PR #627, opened 2024-09-16
Move RVSDG to CFG gamma tests into single file

phate:
1. Moves all the gamma tests of the RVSDG to CFG pass into a single file
2. Cleans them up a bit
3. Fix indentation of test files in Makefile

This is part of the fix for #586.


haved (approved):
Look good.

Was the indentation change in the Makefile intentional? The TESTS and SOURCES now use 4 spaces, while HEADERS and TEST_LIBS use tab.

I see you omit the "Test" prefix from the functions in the test files. I can start doing that too

PR #630, opened 2024-09-18
Add iterator ranges to Region class

phate:


jlm/rvsdg/region.hpp
```diff
@@ -159,13 +160,128 @@ class Region
...
+
+  using NodeIterator = region_nodes_list::iterator;
+  using NodeConstIterator = region_nodes_list::const_iterator;
+  using NodeRange = util::iterator_range<NodeIterator>;
+  using NodeConstRange = util::iterator_range<NodeConstIterator>;
+
+  using BottomNodeIterator = region_bottom_node_list::iterator;
+  using BottomNodeConstIterator = region_bottom_node_list::const_iterator;
+  using BottomNodeRange = util::iterator_range<BottomNodeIterator>;
+  using BottomNodeConstRange = util::iterator_range<BottomNodeConstIterator>;
```
phate, line 186: This is not completely clean as it exposes the underlying data structure to the outside world. It is for example visible when iterating through the arguments/results as I get pointers instead of references by default.

It is an improvement compared to before, but abstracting this away would have meant to implement our own iterator class in the region class, and I was too lazy for that. I will do it if you insist though.

haved: This is absolutely fine. There is a thing in C++ called `std::reference_wrapper`, which sits somewhere between pointers and references, which in theory could make it possible to store references in a vector, but I think the type is a bit scary, so I will gladly iterate over pointers.

jlm/rvsdg/region.hpp
```diff
@@ -159,13 +160,129 @@ class Region
...
+   * @return Returns an iterator range for iterating through the results of the region.
+   */
+  [[nodiscard]] RegionResultConstRange
+  Results() const noexcept
+  {
+    return { results_.begin(), results_.end() };
+  }
+
+  /**
+   * @return Returns an iterator range for iterating through the top nodes of the region.
```
haved, line 232: Could we get an explanation of what top and bottom nodes are? I assume they are nodes with no predecessor or successor nodes, respectively, but I am not 100% sure.

haved (approved):
Looks good, my only suggestion is adding an explanation of what top and bottom nodes are

jlm/rvsdg/region.hpp
phate: I provided the documentation in #638 as it would not really belong in this PR.

PR #631, opened 2024-09-20
Move CFG conversion to ascii to cfg class

phate:
The ascii output for the control flow graph is not easily readable. This is the first PR on improving it. It does the following:

1. Moves the functions for converting a CFG to ascii to the cfg class
2. Adds a unit test for the conversion of three address codes

In order to add a unit test for the CFG conversion, I first need to make the output more deterministic. This will happen in a follow up PR. This PR is part of issue #586

### DIFF
--- a/tests/jlm/llvm/ir/ThreeAddressCodeTests.cpp
+++ b/tests/jlm/llvm/ir/ThreeAddressCodeTests.cpp
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2024 Håvard Krogstie <krogstie.havard@gmail.com>
+ * See COPYING for terms of redistribution.
+ */
+
+#include <test-operation.hpp>
+#include <test-registry.hpp>
+#include <test-types.hpp>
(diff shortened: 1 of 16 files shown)


tests/jlm/llvm/ir/ThreeAddressCodeTests.cpp
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2024 Håvard Krogstie <krogstie.havard@gmail.com>
```
haved, line 2: Yourself?

haved (changes requested):
Looks good, just a wrong name in the Copyright header

tests/jlm/llvm/ir/ThreeAddressCodeTests.cpp
phate: Done.

haved (approved):

PR #633, opened 2024-09-25
rvsdg2dot use argument and result index as label

### LINKED ISSUE #626: Confusing naming of arguments and results in RVSDG dot graph
The arguments and results of a gamma node in an RVSDG dot graph are named confusing:

image

The corresponding arguments in graph10 and graph11 above represent the same variable (a19 and a21 represent the same variable, as well as a20 and a22), but are named differently. I would have expected the same name in graph10 and graph11. 

A similar problem exists for the results. I would have expected that r6 and r7 is named the same.

haved:
In order to make it easier to read the graphs, arguments and results only get their index. The number in parenthesis is the index of the corresponding input/output of the structural node if any.

image

Let me know if you prefer a different label.
Fixes #626

### DIFF
--- a/jlm/llvm/backend/dot/DotWriter.cpp
+++ b/jlm/llvm/backend/dot/DotWriter.cpp
@@ -145,7 +145,6 @@ CreateGraphNodes(util::Graph & graph, rvsdg::Region & region, util::Graph * type
     AttachNodeOutput(node, argument, typeGraph);
 
     // Give the argument a label using its local index, not the global argument index
-    // Use an uppercase A to distinguish
     node.SetLabel(util::strfmt("a", n));
 
     // If this argument corresponds to one of the structural node's inputs, reference it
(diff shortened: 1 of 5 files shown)


haved:
Here is an alternative label format. Let me know which you prefer:

image

jlm/llvm/backend/dot/DotWriter.cpp
```diff
@@ -144,9 +144,17 @@ CreateGraphNodes(util::Graph & graph, rvsdg::Region & region, util::Graph * type
     auto & argument = *region.argument(n);
     AttachNodeOutput(node, argument, typeGraph);
 
+    // Give the argument a label using its local index, not the global argument index
+    // Use an uppercase A to distinguish
+    node.SetLabel(util::strfmt("a", n));
+
```
phate, line 150: It says "uppercase A" in the comment, but then a lowercase a is used.

phate: Also to distinguish it from what?

phate (dismissed):

jlm/llvm/backend/dot/DotWriter.cpp
haved: Good catch, this comment is not correct

haved:
@phate I fixed the old comment, need another review

phate (approved):

PR #634, opened 2024-09-26
Generalize the memory response and request nodes to handle multiple memory ports

sjalander:
Provides support for memories with multiple read and/or write ports, e.g., BRAMs in AMD FPGAs.

### DIFF
--- a/jlm/hls/backend/rhls2firrtl/RhlsToFirrtlConverter.cpp
+++ b/jlm/hls/backend/rhls2firrtl/RhlsToFirrtlConverter.cpp
@@ -664,16 +664,16 @@ RhlsToFirrtlConverter::MlirGenHlsMemResp(const jlm::rvsdg::simple_node * node)
   auto module = nodeToModule(node, false);
   auto body = module.getBodyBlock();
 
-  mlir::BlockArgument memRes = GetInPort(module, 0);
-  auto memResValid = GetSubfield(body, memRes, "valid");
-  auto memResReady = GetSubfield(body, memRes, "ready");
-  auto memResBundle = GetSubfield(body, memRes, "data");
-  auto memResId = GetSubfield(body, memResBundle, "id");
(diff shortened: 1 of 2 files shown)


jlm/hls/backend/rhls2firrtl/RhlsToFirrtlConverter.cpp
```diff
@@ -784,84 +790,114 @@ RhlsToFirrtlConverter::MlirGenHlsMemReq(const jlm::rvsdg::simple_node * node)
...
-    }
-    int log2Bytes = log2(bitWidth / 8);
-    Connect(thenBody, memReqSize, GetConstant(thenBody, 3, log2Bytes));
+  ::llvm::SmallVector<mlir::Value> loadGranted(loadTypes->size(), zeroBitValue);
+  ::llvm::SmallVector<mlir::Value> storeGranted(storeTypes->size(), zeroBitValue);
+  for (size_t j = 0; j < node->noutputs(); ++j)
+  {
+    auto reqType = dynamic_cast<const jlm::hls::bundletype *>(&node->output(j)->type());
+    // TODO: more robust check
+    auto hasWrite = reqType->elements_.size() == 5;
```
phate, line 799: Can it never happen that the type of the output is not a bundletype?

phate: If that is the case (as the code suggest), then maybe perform a `AssertedCast<>` instead. Otherwise, a check is needed.

phate (changes requested):

jlm/hls/backend/rhls2firrtl/RhlsToFirrtlConverter.cpp
sjalander: Changed to AssertedCast

phate (approved):

PR #635, opened 2024-09-28
Avoid hoisting loads out of gammas

### LINKED ISSUE #632: Polybench tests are failing
It seems that the polybench tests started to fail all in a sudden. See PR #630 and #631.

phate:
Closes #632


haved (approved):
So this is the culprit? Great find!

PR #636, opened 2024-09-28
Remove lambda::cvinput

caleridas:
Remove lambad::cvinput and introduce a new API to map inputs to associated region arguments.

The only reason all these subclasses exist is to provide an API for mapping, however the data held
in these inputs is entirely redundant as the node already provides (and needs to provide) the
correct mapping semantic.

### DIFF
--- a/jlm/llvm/ir/operators/lambda.cpp
+++ b/jlm/llvm/ir/operators/lambda.cpp
@@ -41,18 +41,6 @@ operation::copy() const
 
 node::~node() = default;
 
-node::fctargument_range
-node::fctarguments()
-{
-  fctargiterator end(nullptr);
-
-  if (nfctarguments() == 0)
-    return { end, end };
-
-  fctargiterator begin(fctargument(0));
-  return { begin, end };
-}
-
 node::fctargument_constrange
 node::fctarguments() const
 {
@@ -62,42 +50,6 @@ node::fctarguments() const
     return { end, end };
 
   fctargconstiterator begin(fctargument(0));
-  return { begin, end };
-}
-
-node::ctxvar_range
-node::ctxvars()
-{
-  cviterator end(nullptr);
-
-  if (ncvarguments() == 0)
-    return { end, end };
-
-  cviterator begin(input(0));
-  return { begin, end };
-}
-
-node::ctxvar_constrange
-node::ctxvars() const
-{
-  cvconstiterator end(nullptr);
-
-  if (ncvarguments() == 0)
-    return { end, end };
-
-  cvconstiterator begin(input(0));
-  return { begin, end };
-}
-
-node::fctresult_range
-node::fctresults()
-{
-  fctresiterator end(nullptr);
-
-  if (nfctresults() == 0)
-    return { end, end };
-
-  fctresiterator begin(fctresult(0));
   return { begin, end };
 }
 
@@ -113,10 +65,36 @@ node::fctresults() const
   return { begin, end };
 }
 
-cvinput *
-node::input(size_t n) const noexcept
+[[nodiscard]] node::ContextVar
+node::MapInputContextVar(const rvsdg::input & input) const noexcept
 {
-  return util::AssertedCast<cvinput>(structural_node::input(n));
+  JLM_ASSERT(rvsdg::input::GetNode(input) == this);
+  return GetContextVar(input.index());
+}
+
+[[nodiscard]] std::optional<node::ContextVar>
+node::MapBinderContextVar(const rvsdg::output & output) const noexcept
+{
+  auto argument = util::AssertedCast<const lambda::cvargument>(&output);
+  return ContextVar{ argument->input(), const_cast<lambda::cvargument *>(argument) };
+}
+
+[[nodiscard]] node::ContextVar
+node::GetContextVar(size_t index) const noexcept
+{
+  return ContextVar{ input(index),
+                     subregion()->argument(operation().Type()->NumArguments() + index) };
+}
+
+[[nodiscard]] std::vector<node::ContextVar>
+node::GetContextVars() const noexcept
+{
+  std::vector<ContextVar> vars;
+  for (size_t n = 0; n < ninputs(); ++n)
+  {
+    vars.push_back(GetContextVar(n));
+  }
+  return vars;
 }
 
 lambda::output *
@@ -131,23 +109,18 @@ node::fctargument(size_t n) const noexcept
   return util::AssertedCast<lambda::fctargument>(subregion()->argument(n));
 }
 
-lambda::cvargument *
-node::cvargument(size_t n) const noexcept
-{
-  return input(n)->argument();
-}
-
 lambda::result *
 node::fctresult(size_t n) const noexcept
 {
   return util::AssertedCast<lambda::result>(subregion()->result(n));
 }
 
-cvargument *
-node::add_ctxvar(jlm::rvsdg::output * origin)
+node::ContextVar
+node::AddContextVar(jlm::rvsdg::output * origin)
 {
-  auto input = cvinput::create(this, origin);
-  return cvargument::create(subregion(), input);
+  auto input = rvsdg::structural_input::create(this, origin, origin->Type());
+  auto argument = cvargument::create(subregion(), input);
+  return ContextVar{ input, argument };
 }
 
 rvsdg::RegionArgument &
@@ -251,11 +224,10 @@ node::copy(rvsdg::Region * region, rvsdg::SubstitutionMap & smap) const
 
   /* add context variables */
   rvsdg::SubstitutionMap subregionmap;
-  for (auto & cv : ctxvars())
+  for (const auto & cv : GetContextVars())
   {
-    auto origin = smap.lookup(cv.origin());
-    auto newcv = lambda->add_ctxvar(origin);
-    subregionmap.insert(cv.argument(), newcv);
+    auto origin = smap.lookup(cv.input->origin());
+    subregionmap.insert(cv.inner, lambda->AddContextVar(origin).inner);
   }
 
   /* collect function arguments */
@@ -295,10 +267,12 @@ node::ComputeCallSummary() const
     auto input = worklist.front();
     worklist.pop_front();
 
-    if (auto cvinput = dynamic_cast<lambda::cvinput *>(input))
+    auto inputNode = rvsdg::input::GetNode(*input);
+
+    if (auto lambdaNode = dynamic_cast<lambda::node *>(inputNode))
     {
-      auto argument = cvinput->argument();
-      worklist.insert(worklist.end(), argument->begin(), argument->end());
+      auto & argument = *lambdaNode->MapInputContextVar(*input).inner;
+      worklist.insert(worklist.end(), argument.begin(), argument.end());
       continue;
     }
 
@@ -366,7 +340,6 @@ node::ComputeCallSummary() const
       continue;
     }
 
-    auto inputNode = rvsdg::input::GetNode(*input);
     if (is<CallOperation>(inputNode) && input == inputNode->input(0))
     {
       directCalls.emplace_back(util::AssertedCast<CallNode>(inputNode));
@@ -379,8 +352,7 @@ node::ComputeCallSummary() const
       continue;
     }
 
-    auto simpleInput = dynamic_cast<rvsdg::simple_input *>(input);
-    if (simpleInput != nullptr)
+    if (auto simpleInput = dynamic_cast<rvsdg::simple_input *>(input))
     {
       otherUsers.emplace_back(simpleInput);
       continue;
@@ -397,16 +369,6 @@ node::IsExported(const lambda::node & lambdaNode)
 {
   auto callSummary = lambdaNode.ComputeCallSummary();
   return callSummary->IsExported();
-}
-
-/* lambda context variable input class */
-
-cvinput::~cvinput() = default;
-
-cvargument *
-cvinput::argument() const noexcept
-{
-  return util::AssertedCast<cvargument>(arguments.first());
 }
 
 /* lambda output class */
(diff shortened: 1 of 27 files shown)


caleridas:
The same strategy can be applied to all other kinds of inputs, as well as output/result/argument 
classes and thus remove a lot of redundant boilerplate classes. However before investing time into
this, I would like you to have a look at the resulting API -- in my opinion it is just fine (although
right now it obviously does not look very "consistent" because this still needs to be applied to
gamma, theta etc.).

Ultimately this greatly simplifies things and I would really like to get this in before pushing generic
lambda, phi & friends.

jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -146,8 +142,7 @@ class node final : public jlm::rvsdg::structural_node
   using fctargument_range = jlm::util::iterator_range<fctargiterator>;
   using fctargument_constrange = jlm::util::iterator_range<fctargconstiterator>;
 
-  using ctxvar_range = jlm::util::iterator_range<cviterator>;
-  using ctxvar_constrange = jlm::util::iterator_range<cvconstiterator>;
+  using ctxvar_constrange = jlm::util::iterator_range<rvsdg::node_const_input_iterator>;
```
phate, line 145: I want to get away from the the `ctxvar` naming. It is not very intuitive.

`using LambdaInputConstRange = ...`

jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -273,16 +259,27 @@ class node final : public jlm::rvsdg::structural_node
...
+   * @param input Input to the lambda node
+   * @return The region argument corresponding to the input
+   *
+   * The given input must be an input to this lambda node.
+   * All inputs represent by definition bound context variables
+   * and map to corresponding arguments in the lambda defining
+   * region.
+   */
+  [[nodiscard]] rvsdg::RegionArgument *
+  MapInputToRegionArgument(const rvsdg::input * input) const noexcept;
```
phate, line 282: I find the name a mouthful. How about just `RegionArgument` or `GetRegionArgument`? I also think that the signature should be:
```
[[nodiscard]] rvsdg::RegionArgument *
RegionArgument(const rvsdg::input & input) const noexcept
```

phate: We could also make the return type `rvsdg::RegionArgument &` and just leave it to the programmer to ensure that a correct  input is handed in. An assert in the body of the method would then be appropriate though.

jlm/llvm/ir/operators/lambda.cpp
```diff
@@ -295,9 +268,11 @@ node::ComputeCallSummary() const
     auto input = worklist.front();
     worklist.pop_front();
 
-    if (auto cvinput = dynamic_cast<lambda::cvinput *>(input))
+    auto inputNode = rvsdg::input::GetNode(*input);
+
+    if (auto lambda_node = dynamic_cast<lambda::node *>(inputNode))
```
phate, line 273: `if (auto lambdaNode = ....`

jlm/rvsdg/node.hpp
```diff
@@ -846,6 +846,40 @@ class node
   std::vector<std::unique_ptr<node_output>> outputs_;
 };
 
+class node_input_iterator final : public input::iterator<input>
```
phate, line 849: `class NodeInputIterator`

jlm/rvsdg/node.hpp
```diff
@@ -846,6 +846,40 @@ class node
...
+  next() const override
+  {
+    auto index = value()->index() + 1;
+    auto node = input::GetNode(*value());
+
+    return index >= node->ninputs() ? nullptr : node->input(index);
+  }
+};
+
+class node_const_input_iterator final : public input::iterator<const input>
```
phate, line 866: `class NodeInputConstIterator`

jlm/rvsdg/node.hpp
```diff
@@ -846,6 +846,40 @@ class node
...
+      : input::iterator<input>(value)
+  {}
+
+  [[nodiscard]] input *
+  next() const override
+  {
+    auto index = value()->index() + 1;
+    auto node = input::GetNode(*value());
+
+    return index >= node->ninputs() ? nullptr : node->input(index);
```
phate, line 862: Is node guaranteed to be not NULL here?

jlm/rvsdg/node.hpp
```diff
@@ -846,6 +846,40 @@ class node
...
+      : input::iterator<const input>(value)
+  {}
+
+  [[nodiscard]] const input *
+  next() const override
+  {
+    auto index = value()->index() + 1;
+    auto node = input::GetNode(*value());
+
+    return index >= node->ninputs() ? nullptr : node->input(index);
```
phate, line 879: Is node guaranteed to be not NULL here?

phate (changes requested):
I am absolutely fine with this change. The only comments I have is about naming and some style.

I think we should extend the `input::GetNode()` eventually to `input::GetNode<TNodeType>()` such that it also performs the casting. The caller is most of the time certain what kind of node is wanted.

caleridas:
> I am absolutely fine with this change. The only comments I have is about naming and some style.

Okay good, your style comments are also fine. Since this creates an "incoherent API" in the present form (as lambda::input is different from GammaInput, theta::input, ... are all different), I would like to approach it this to change them all in one go, and then prepare that for merge. That should also from review POV should make it easier to ensure that it fits together.

> I think we should extend the `input::GetNode()` eventually to `input::GetNode<TNodeType>()` such that it also performs the casting. The caller is most of the time certain what kind of node is wanted.

That I think will not work this way -- when inputs become "generic", actually caller will _not_ know what node they expect. They need access to the "generic" node and then try specializing it into what they can handle. See the example section that I also changed as part of this PR:

Before:
```
    if (auto cvinput = dynamic_cast<lambda::cvinput *>(input))
```

After:
```
    if (auto lambda_node = dynamic_cast<lambda::node *>(inputNode))
```

and expect that this is the canonical dispatching pattern that we will then see from now onwards.

Additionally, I would also like to change the `GetNode` API at some point: Basically we have two kinds of inputs/outputs: Those attached to a node, and those that are region arguments/results. I think the ideal clean API should go towards a function that returns _either_ the owning node _or_ the owning region. And then user code should proceed dispatching starting from this discrimination.

jlm/llvm/ir/operators/lambda.hpp
caleridas: Can change it, although I am less fond of references compared to pointers TBH. The semantic difference that references convey is that they tell "cannot be null". On the downside, they tend to cause trouble with accidental copies (`auto x = foo.GetSomething()` ), cannot directly go into `std::optional`/`std::variant`/containers  without sprinkling `&`/`*` adornments etc. I think that ideally we would have "safe wrapper types" with value semantics that act as input / output / node handles etc., cannot be null but behave sanely for copies etc. However I would not go there right now.

caleridas: > I find the name a mouthful. How about just `RegionArgument` or `GetRegionArgument`?

Yes the name is horrible. This is difficult to find name that is neither ambiguous nor too long. `RegionArgument` is already the name of the class, I would like to avoid clashing with that. Other ideas that I came up with:
- `InputToInner` or `ToInnerArgument` or `ToInner` as it maps an (outer) input to an (inner) argument (or arguments, for gamma then). `ToInner` would be less great as we need the equivalent for outputs and would then have an overloaded name.
- `ArgumentOf` or `ArgumentOfInput` or `RegionArgumentOf` or similar
- `ProjectInputToMatchingArgumentOfSubregion` would be the most precise

jlm/llvm/ir/operators/lambda.hpp
caleridas: Sounds good, in principle I would say this does not even need a typename alias -- knowing that it returns a list of inputs ought to be enough. Eventually, a generic `NodeInputRange` to represent arbitrary lists of nodes would most likely emerge as the generic thing in all places where currently specific range types are used.

jlm/llvm/ir/operators/lambda.hpp
phate: The inputs should(!) be marked as explicitly deleted, so the accidental copies should never happen.

How about `MapToRegionArgument` or `MapToArgument`? The argument and return type convey what we map to what, respectively. I think this should give us some wiggle room in terms of naming.

phate:
> That I think will not work this way -- when inputs become "generic", actually caller will not know what node they expect. They need access to the "generic" node and then try specializing it into what they can handle.

The caller is always able to get the most generic node if it is not known what node it is, i.e., `GetNode<rvsdg::node>()`, and then can dispatch. In other words, it serves as a simple getter method (the compiler should be able to optimize the unnecessary cast out). However, there are plenty of locations where the exact node is known and the caller should get it directly. You will not see this here with the cvinputs, but this will most likely later come with the outputs.

jlm/llvm/ir/operators/lambda.hpp
caleridas: `MapToArgument` sounds good in principle -- I think we ought to have somewhat coherent API names for the following mapping operations:

- input of structural node to region argument(s) (applies to lambda, gamma, theta, phi, delta)
- output of structural node to region result(s) (applies to gamma, theta, phi, delta)
- and corresponding reverse mappings

So it would then be `MapToArgument` / `MapToArguments`, `MapToResult` / `MapToResults`, `MapToInput`, `MapToOutput` ? BTW this is where I thought the notion of inner / outer would explain the API quite nicely.

pointers vs. references: presently trying hard to have references, but it causes enormous tying to specific data structures -- e.g. for returning multiple arguments, I would be inclined to simply use `std::vector<RegionArgument*>`  for now at least, until the eventual structures become clear. If I have to maintain various custom iterator ranges for every use case that makes changes a bit slower, but I cannot do anything involving references with standard containers.

Also, we may have cases where there really isn't a mapping -- e.g. the theta continuation predicate does not go outwards. If I have a function mapping region results to outputs of the theta node I either have to throw, or have a way to express `std::optional` of the thing (which also would indeed be expressible by nullptr). I know that there is code that cares about semantically mapping region entry/exits to inputs/outputs, but there is also code just trying to generically trace through and would like to just see the continuation paths.

phate: > So it would then be MapToArgument / MapToArguments, MapToResult / MapToResults, MapToInput, MapToOutput ? BTW this is where I thought the notion of inner / outer would explain the API quite nicely.

The names you suggest sound fine to me. What I do not like about the `Inner`/`Outer` is that there is no counterpart. In other words, `ToInnerArgument` sounds off to me as there is no `ToOuterArgument`. Thus, the `Inner` is redundant here. A simple `ToArgument` or `MapToArgument` is less confusing in my opinion (the inner is implied as the method takes an input and returns an argument).

> pointers vs. references: presently trying hard to have references, but it causes enormous tying to specific data structures -- e.g. for returning multiple arguments, I would be inclined to simply use std::vector<RegionArgument*> for now at least, until the eventual structures become clear. If I have to maintain various custom iterator ranges for every use case that makes changes a bit slower, but I cannot do anything involving references with standard containers.

We currently (try to) use references (where applicable (i.e. no chance of copying) and it is always guaranteed to have a valid element) for individual instances (RegionArgument, RegionResult, etc.), but when we give back collections/containers of instances, then we currently use pointers, i.e. `std::vector<RegionArgument*>` is perfectly fine. I have not dig into how to use references in containers. As pointed out by you or @haved in another PR, it is possible, but it is fine without for now.

> Also, we may have cases where there really isn't a mapping -- e.g. the theta continuation predicate does not go outwards. If I have a function mapping region results to outputs of the theta node I either have to throw, or have a way to express std::optional of the thing (which also would indeed be expressible by nullptr). I know that there is code that cares about semantically mapping region entry/exits to inputs/outputs, but there is also code just trying to generically trace through and would like to just see the continuation paths.

If there is no guarantee that it is possible to get a valid element, then using pointers is perfectly fine (I think). The alternative is throwing an exception (not big fan of it) or leaving it to the user of the method, documenting it and putting asserts in the method to double check. That depends a little bit on the context I think.

jlm/rvsdg/node.hpp
caleridas: removed the entire helper class -- I think working with std::vector<...> of things is really a lot easier and less constraining on the internal data structures. I would be happy to revisit later (e.g. expose std::span things).

jlm/rvsdg/node.hpp
caleridas: removed

jlm/rvsdg/node.hpp
caleridas: removed

jlm/llvm/ir/operators/lambda.cpp
caleridas: done

jlm/llvm/ir/operators/lambda.hpp
caleridas: Used references in all places possible. Now that the overall API becomes clearer I have resorted to pointers in places necessary.

Looking at where mapping is done, it is clear that this 95% at construction time, and very little at tracing. So I think the best thing to do is to provide an API that returns a "description" of the compound thing being created. For reference, I also uploaded changes for gamma and theta to illustrate the concept, see https://github.com/phate/jlm/commit/217e8798714cbd628a6bff9c67f4f157881e7576 and https://github.com/phate/jlm/commit/63e38024acd6820540d549056449ab73ba90aaf0 -- particular example: https://github.com/phate/jlm/blob/217e8798714cbd628a6bff9c67f4f157881e7576/jlm/rvsdg/gamma.hpp#L230

So, caller gets what they need to proceed from call to AddEntryVar (or AddLoopVar, or AddContextVar). Also, all "mapping" operations that want to resolve the full EntryVar/LoopVar/ContextVar just supply the piece they "have" and can get the full thing resolved. I think this makes for a consistent and well-usable API. WDYT?

PR #639, opened 2024-09-29
Bumps hls-suite that now checks cycle count

sjalander:


haved (approved):
This change looks good. I would just like to confirm the plan of action if a PR purposefully wants to change the expected number of cycles. Am I correct in assuming the process becomes:

- Make a new commit with the new numbers in the `hls-test-suite` repo and merge it into its master with a  `[AllowCycleCountMismatch]` in the PR title
 - Update the referenced commit hash in the PR in this repo that causes the cycle counts to change.

That should work, without any further changes to this repo.

My only concern is that we still have some non-determinism. If that turns out to cause non-deterministic cycle counts we might have to revert this PR. Hopefully we have little issues, and can make jlm deterministic instead.

PR #645, opened 2024-10-03
Rename statetype class to StateType

phate:


haved (approved):
Looks good, and the use of override instead of virtual was nice

PR #648, opened 2024-10-06
Remove static output_node::node() method

phate:
The method was superseded by `output::GetNode()`.


caleridas:
I think GetNode is going to be temporary stand-in as well -- I have another PR in prep that adds to input/output a method like this:

`std::variant<node*, Region*> GetOwner()`

That should systematically + exhaustively represent the available choices and be the basis for all dispatch.

phate:
@caleridas Sounds good. However, could we still approve this one, so we have only one method that needs to be replaced. Or are there any concerns?

caleridas:
No concerns, just rebase noise. But let's move ahead.

caleridas (approved):

PR #649, opened 2024-10-06
Simplify annotation computations in RvsdgTreePrinter class

phate:
Use `std::count_if()` instead of explicit iteration to compute relevant annotations in `RvsdgTreePrinter` class.


phate (commented):
I can not make use of the CountWhere() method for the structural inputs and outputs in the AnnotateNumMemoryStateInputsOutputs() method yet, as we do not have ranges for these yet.

caleridas:
Would it not make more sense to defer to the std::count_if / std::ranges::count_if stdlib facilities?

phate:
@caleridas I was not aware of them. Changed the PR. Thanks, I learned something!

haved (approved):
Looks good, nice use of the STL

PR #650, opened 2024-10-06
Remove outdated begin() and end() methods from Region class

phate:


haved (approved):
Nice cleanup, no need to have two almost identical ways of doing the same thing

PR #651, opened 2024-10-07
Privatize bottom nodes in Region class

phate:


caleridas (approved):
One thing that would make it more systematic is if the range accessor would support "empty" and "size" like regular std containers:

`r.BottomNodes().size()` instead of `r.NumBottomNodes()` and `r.BottomNodes().empty()` -- would suit to make the API orthogonal.

phate:
@caleridas I can do this for the range container, but I would not like to use it to replace `NumButtonNodes()`. My plan was to eventually replace the bottom node intrusive list with a HashSet. For a HashSet, the complexity of the `size()` method is constant, where as for the intrusive list it is linear in the number of nodes.

Issue #579, opened 2024-08-16
Clean up jlm builds in CI


phate:
We currently have only a limited set of jlm builds that are tested in the CI:

1. Debug build with clang (and running unit tests)
2. Debug build with g++ (and running unit tests)

We should at least have at least two more builds:

1. Release build without asserts with clang (and running unit tests)
2. Release build without asserts with g++ (and running unit tests)
3. Minimal debug build (disabling hls dialect and mlir dialect) with clang and g++ (and running unit tests)
4. Minimal debug build (disabling hls dialect and mlir dialect) with clang and g++ (and running unit tests)

haved:
One thing to note: When we use `clang++` in CI, it is actually clang++ 14, which is the default on Ubuntu 22.04, and installed by default on GitHub Action's `ubuntu::22.04` image.

When we build CIRCT and the MLIR dialect, we use clang 18, however.

If we want to, we can specify that we actually want `clang++-18` for our clang builds of jlm, but mixing compiler versions does not appear to be an issue. The ABI should be stable, and I think they both use the same standard library from the distribution (?).

PR #653, opened 2024-10-09
unbreak release build: no side effects in JLM_ASSERT

caleridas:
Operations that have side effect should not go into JLM_ASSERT.
Otherwise, they will not be executed in release builds.

### DIFF
--- a/jlm/rvsdg/node.cpp
+++ b/jlm/rvsdg/node.cpp
@@ -113,7 +113,8 @@ output::remove_user(jlm::rvsdg::input * user)
   {
     if (!node->has_users())
     {
-      JLM_ASSERT(region()->AddBottomNode(*node));
+      bool wasAdded = region()->AddBottomNode(*node);
+      JLM_ASSERT(wasAdded);
     }
   }
 }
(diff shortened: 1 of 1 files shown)


jlm/rvsdg/node.cpp
```diff
@@ -113,7 +113,7 @@ output::remove_user(jlm::rvsdg::input * user)
   {
     if (!node->has_users())
     {
-      JLM_ASSERT(region()->AddBottomNode(*node));
+      region()->AddBottomNode(*node);
```
phate, line 116: Could we do something like this for all of them instead:

```
auto wasAdded = region()->AddBottomNode(*node);
JLM_ASSERT(wasAdded)
```

phate (changes requested):
Uff that is a big one. It would have been caught if issue #579 would have been done already.

jlm/rvsdg/node.cpp
caleridas: Doesn't this risk unused variable warnings in release builds instead?

haved: I don't think so, as the definition of `JLM_ASSERT` is
```c++
#define JLM_ASSERT(x) \
  do                  \
  {                   \
  } while (0 && (x))
```

Which means the expression in still "used", without being evaluated. Unless the compiler is eliminating dead code before checking, it should not give a warning.

I tested the following in g++ 14.2.1 and clang++ 18.1.8, with `-Wall -Wextra`, and they only complain about `z`.
```c++
int main() {
    int x = 5;
    int y = 7;
    int z = 10;

    do {
    } while(0 & (x));

    return y;
}
```

caleridas: oh right -- changed it accordingly

phate (approved):

PR #654, opened 2024-10-10
Introduce accessor and convenience API for node type matching

caleridas:
Introduce input::GetOwner and output::GetOwner as the canonical and type-generic way to determine what "kind" an output/input is (either output/input to a node or region argument/result).

Introduce TryGetOwnerNode and TryGetRegionParentNode as convenience API to easily match different node kinds.

### DIFF
--- a/jlm/rvsdg/node.cpp
+++ b/jlm/rvsdg/node.cpp
@@ -167,12 +167,24 @@ node_input::node_input(
       node_(node)
 {}
 
+[[nodiscard]] std::variant<node *, Region *>
+node_input::GetOwner() const noexcept
+{
+  return node_;
+}
+
 /* node_output class */
 
 node_output::node_output(jlm::rvsdg::node * node, std::shared_ptr<const rvsdg::Type> type)
     : jlm::rvsdg::output(node->region(), std::move(type)),
       node_(node)
 {}
+
+[[nodiscard]] std::variant<node *, Region *>
+node_output::GetOwner() const noexcept
+{
+  return node_;
+}
 
 /* node class */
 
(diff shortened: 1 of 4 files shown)


caleridas:
This introduces accessor API that I think is convenient for matching node kinds.

Names are up for debate, couple notes:
- I am inclined to say that what is current named "output" should instead be called "Def", and what is currently called "input" should be named "Use" instead (as that is canonical compiler terminology). Then (Node)Input/(Node)Output/(Region)Argument/(Region)Result would be special forms of def and use. (That's quite a sweeping change, but rvsdg::input and rvsdg::output need renaming eventually anyways to comply with naming scheme, so it is actually not really any additional effort, just a decision).
- alternative for "TryGetOwningNode": "TryGetOwnerNode", "TryGetOriginNode/TryGetTargetNode" (so the name is not overloaded for input/output -- in general, I prefer non-overloaded names), or when going with def/use eventually also "TryDefAsOutputOf<node_type> / TryUseAsInputTo<node_type>" and "TryDefAsArgumentOf<node_type>/TryUseAsResultOf<node_type>" etc.
- I am inclined to also add "Assert..." in addition to the "Try...." versions that also check but either return what was asked for or throw when expectation is not matched. I think that this should generally not be used outside tests, but there it makes intent perfectly clear.

jlm/rvsdg/node.cpp
```diff
@@ -337,4 +349,58 @@ normalize(jlm::rvsdg::node * node)
...
+  Likewise, every \p input is one of the following:
+
+  - the input of a node representing an operation
+  - the exit result of a region
+
+  Analysis code can determine which of the two is the case using
+  \ref rvsdg::output::GetOwner and \ref rvsdg::input::GetOwner, respectively,
+  and then branch deeper based on its results. For convenience, code
+  can more directly check immediately for the specific kinds of nodes
+  involved using
```
phate, line 377: `involved using ...` Something is missing here.

jlm/rvsdg/node.cpp
```diff
@@ -337,4 +349,58 @@ normalize(jlm::rvsdg::node * node)
...
+  However is is generally desirable to check for the specific kinds of
```
phate, line 378: However, it(!) is

phate (changes requested):

haved:
@phate

My immediate reaction was a bit sceptical to def/use, as I think node outputs and inputs make a lot of sense. However when I also consider arguments and results, I agree that it sounds a bit weird to call the arguments of a region "outputs", and the results "inputs". Def/use is common SSA terminology, so we can use that.

Regarding point 2, I think the suggested names are fine.

jlm/rvsdg/node.cpp
caleridas: fixed

jlm/rvsdg/node.cpp
caleridas: fixed

phate (approved):

PR #658, opened 2024-10-12
lambda: revise API as well as results, arguments and outputs

caleridas:
Remove all auxiliary input/output/argument/result classes for lambda. Provide new API for mapping pieces to context variables. Change all dispatch to be based on checking for "lambda node" instead of subclassed inputs/outputs.

This leads to:
- making the lambda abstraction a generic rvsdg concept
- removing all needs to implement subclasses of argument, result, input and output
- dispatching purely based on node kind

### DIFF
--- a/jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
@@ -76,7 +76,7 @@ RemoveUnusedStatesFromLambda(llvm::lambda::node & lambdaNode)
     auto oldArgument = ctxvar.inner;
     auto origin = ctxvar.input->origin();
 
-    auto newArgument = newLambda->AddContextVar(origin).inner;
+    auto newArgument = newLambda->AddContextVar(*origin).inner;
     substitutionMap.insert(oldArgument, newArgument);
   }
 
--- a/jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
@@ -14,7 +14,7 @@ namespace jlm::hls
 {
 
 static bool
-IsPassthroughArgument(const rvsdg::RegionArgument & argument)
+IsPassthroughArgument(const rvsdg::output & argument)
 {
   if (argument.nusers() != 1)
   {
@@ -25,7 +25,7 @@ IsPassthroughArgument(const rvsdg::RegionArgument & argument)
 }
 
 static bool
-IsPassthroughResult(const rvsdg::RegionResult & result)
+IsPassthroughResult(const rvsdg::input & result)
 {
   auto argument = dynamic_cast<rvsdg::RegionArgument *>(result.origin());
   return argument != nullptr;
@@ -71,31 +71,30 @@ RemoveUnusedStatesFromLambda(llvm::lambda::node & lambdaNode)
       lambdaNode.attributes());
 
   rvsdg::SubstitutionMap substitutionMap;
-  for (size_t i = 0; i < lambdaNode.ncvarguments(); ++i)
+  for (const auto & ctxvar : lambdaNode.GetContextVars())
   {
-    auto oldArgument = lambdaNode.cvargument(i);
-    auto origin = oldArgument->input()->origin();
+    auto oldArgument = ctxvar.inner;
+    auto origin = ctxvar.input->origin();
 
-    auto newArgument = newLambda->add_ctxvar(origin);
+    auto newArgument = newLambda->AddContextVar(origin).inner;
     substitutionMap.insert(oldArgument, newArgument);
   }
 
   size_t new_i = 0;
-  for (size_t i = 0; i < lambdaNode.nfctarguments(); ++i)
+  auto newArgs = newLambda->GetFunctionArguments();
+  for (auto argument : lambdaNode.GetFunctionArguments())
   {
-    auto argument = lambdaNode.fctargument(i);
     if (!IsPassthroughArgument(*argument))
     {
-      substitutionMap.insert(argument, newLambda->fctargument(new_i));
+      substitutionMap.insert(argument, newArgs[new_i]);
       new_i++;
     }
   }
   lambdaNode.subregion()->copy(newLambda->subregion(), substitutionMap, false, false);
 
   std::vector<jlm::rvsdg::output *> newResults;
-  for (size_t i = 0; i < lambdaNode.nfctresults(); ++i)
+  for (auto result : lambdaNode.GetFunctionResults())
   {
-    auto result = lambdaNode.fctresult(i);
     if (!IsPassthroughResult(*result))
     {
       newResults.push_back(substitutionMap.lookup(result->origin()));
--- a/jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
@@ -235,14 +235,14 @@ find_decouple_response(
     const jlm::llvm::lambda::node * lambda,
     const jlm::rvsdg::bitconstant_op * request_constant)
 {
-  jlm::rvsdg::RegionArgument * response_function = nullptr;
-  for (size_t i = 0; i < lambda->ncvarguments(); ++i)
+  jlm::rvsdg::output * response_function = nullptr;
+  for (const auto & ctxvar : lambda->GetContextVars())
   {
-    auto ip = lambda->cvargument(i)->input();
+    auto ip = ctxvar.input;
     if (dynamic_cast<const jlm::llvm::PointerType *>(ip)
         && get_impport_function_name(ip) == "decouple_response")
     {
-      response_function = lambda->cvargument(i);
+      response_function = ctxvar.inner;
     }
   }
   JLM_ASSERT(response_function == nullptr);
@@ -647,16 +647,20 @@ jlm::hls::MemoryConverter(jlm::llvm::RvsdgModule & rm)
       lambda->attributes());
 
   rvsdg::SubstitutionMap smap;
-  for (size_t i = 0; i < lambda->ncvarguments(); ++i)
+  for (const auto & ctxvar : lambda->GetContextVars())
   {
-    smap.insert(
-        lambda->cvargument(i),
-        newLambda->add_ctxvar(lambda->cvargument(i)->input()->origin()));
+    smap.insert(ctxvar.inner, newLambda->AddContextVar(ctxvar.input->origin()).inner);
   }
 
-  for (size_t i = 0; i < lambda->nfctarguments(); ++i)
+  auto args = lambda->GetFunctionArguments();
+  auto newArgs = newLambda->GetFunctionArguments();
+  // The new function has more arguments than the old function.
+  // Substitution of existing arguments is safe, but note
+  // that this is not an isomorphism.
+  JLM_ASSERT(args.size() <= newArgs.size());
+  for (size_t i = 0; i < args.size(); ++i)
   {
-    smap.insert(lambda->fctargument(i), newLambda->fctargument(i));
+    smap.insert(args[i], newArgs[i]);
   }
   lambda->subregion()->copy(newLambda->subregion(), smap, false, false);
 
@@ -669,7 +673,7 @@ jlm::hls::MemoryConverter(jlm::llvm::RvsdgModule & rm)
   std::vector<jlm::rvsdg::output *> newResults;
   // The new arguments are placed directly after the original arguments so we create an index that
   // points to the first new argument
-  auto newArgumentsIndex = lambda->nfctarguments();
+  auto newArgumentsIndex = args.size();
   for (auto & portNode : portNodes)
   {
     auto loadNodes = std::get<0>(portNode);
@@ -695,9 +699,9 @@ jlm::hls::MemoryConverter(jlm::llvm::RvsdgModule & rm)
   }
 
   std::vector<jlm::rvsdg::output *> originalResults;
-  for (auto & result : lambda->fctresults())
+  for (auto result : lambda->GetFunctionResults())
   {
-    originalResults.push_back(smap.lookup(result.origin()));
+    originalResults.push_back(smap.lookup(result->origin()));
   }
   originalResults.insert(originalResults.end(), newResults.begin(), newResults.end());
   auto newOut = newLambda->finalize(originalResults);
@@ -724,19 +728,20 @@ jlm::hls::MemoryConverter(jlm::llvm::RvsdgModule & rm)
   newLambda = jlm::util::AssertedCast<jlm::llvm::lambda::node>(root->Nodes().begin().ptr());
 
   // Go through in reverse since we are removing things
-  for (int i = newLambda->ncvarguments() - 1; i >= 0; --i)
+  auto ctxvars = newLambda->GetContextVars();
+  for (size_t n = ctxvars.size(); n > 0; --n)
   {
-    auto cvarg = newLambda->cvargument(i);
-    if (dynamic_cast<const jlm::llvm::PointerType *>(&cvarg->type()))
+    const auto & ctxvar = ctxvars[n - 1];
+    if (dynamic_cast<const jlm::llvm::PointerType *>(&ctxvar.input->type()))
     {
       // The only functions at this time is decoupled loads that are encoded as functions by the
       // user
       auto visited = std::unordered_set<jlm::rvsdg::output *>();
-      if (IsDecoupledFunctionPointer(cvarg->input()->origin(), visited))
+      if (IsDecoupledFunctionPointer(ctxvar.input->origin(), visited))
       {
-        JLM_ASSERT(cvarg->nusers() == 0);
-        auto cvip = cvarg->input();
-        newLambda->subregion()->RemoveArgument(cvarg->index());
+        JLM_ASSERT(ctxvar.inner->nusers() == 0);
+        auto cvip = ctxvar.input;
+        newLambda->subregion()->RemoveArgument(ctxvar.inner->index());
         // TODO: work around const
         newLambda->RemoveInput(cvip->index());
         auto graphImport = util::AssertedCast<const llvm::GraphImport>(cvip->origin());
--- a/jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
@@ -649,7 +649,7 @@ jlm::hls::MemoryConverter(jlm::llvm::RvsdgModule & rm)
   rvsdg::SubstitutionMap smap;
   for (const auto & ctxvar : lambda->GetContextVars())
   {
-    smap.insert(ctxvar.inner, newLambda->AddContextVar(ctxvar.input->origin()).inner);
+    smap.insert(ctxvar.inner, newLambda->AddContextVar(*ctxvar.input->origin()).inner);
   }
 
   auto args = lambda->GetFunctionArguments();
--- a/jlm/hls/backend/rvsdg2rhls/remove-unused-state.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/remove-unused-state.cpp
@@ -209,7 +209,7 @@ remove_lambda_passthrough(llvm::lambda::node * ln)
   for (const auto & ctxvar : ln->GetContextVars())
   {
     // copy over context vars
-    smap.insert(ctxvar.inner, new_lambda->AddContextVar(ctxvar.input->origin()).inner);
+    smap.insert(ctxvar.inner, new_lambda->AddContextVar(*ctxvar.input->origin()).inner);
   }
 
   size_t new_i = 0;
--- a/jlm/hls/backend/rvsdg2rhls/remove-unused-state.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/remove-unused-state.cpp
@@ -206,28 +206,30 @@ remove_lambda_passthrough(llvm::lambda::node * ln)
       ln->attributes());
 
   rvsdg::SubstitutionMap smap;
-  for (size_t i = 0; i < ln->ncvarguments(); ++i)
+  for (const auto & ctxvar : ln->GetContextVars())
   {
-    // copy over cvarguments
-    smap.insert(ln->cvargument(i), new_lambda->add_ctxvar(ln->cvargument(i)->input()->origin()));
+    // copy over context vars
+    smap.insert(ctxvar.inner, new_lambda->AddContextVar(ctxvar.input->origin()).inner);
   }
 
   size_t new_i = 0;
-  for (size_t i = 0; i < ln->nfctarguments(); ++i)
+  auto args = ln->GetFunctionArguments();
+  auto new_args = new_lambda->GetFunctionArguments();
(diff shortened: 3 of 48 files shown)


jlm/llvm/ir/operators/lambda.cpp
```diff
@@ -41,127 +41,96 @@ operation::copy() const
...
-  auto input = cvinput::create(this, origin);
-  return cvargument::create(subregion(), input);
-}
-
-rvsdg::RegionArgument &
+rvsdg::output &
 node::GetMemoryStateRegionArgument() const noexcept
 {
-  auto argument = fctargument(nfctarguments() - 1);
+  auto argument = GetFunctionArguments().back();
```
phate, line 125: This is expensive, no? You build an entire vector just to take the last element. It is possible to directly address the element, no?

These methods are wrong in the first place (after the introduction of memory state encoding), but let's leave them for now.

jlm/llvm/ir/operators/lambda.cpp
```diff
@@ -41,127 +41,96 @@ operation::copy() const
...
   JLM_ASSERT(is<MemoryStateType>(argument->type()));
   return *argument;
 }
 
-rvsdg::RegionResult &
+rvsdg::input &
 node::GetMemoryStateRegionResult() const noexcept
 {
-  auto result = fctresult(nfctresults() - 1);
+  auto result = GetFunctionResults().back();
```
phate, line 133: Same here.

jlm/llvm/ir/operators/lambda.cpp
```diff
@@ -251,32 +220,35 @@ node::copy(rvsdg::Region * region, rvsdg::SubstitutionMap & smap) const
...
-    subregionmap.insert(cv.argument(), newcv);
+    auto origin = smap.lookup(cv.input->origin());
+    subregionmap.insert(cv.inner, lambda->AddContextVar(origin).inner);
   }
 
   /* collect function arguments */
-  for (size_t n = 0; n < nfctarguments(); n++)
+  auto args = GetFunctionArguments();
+  auto newArgs = lambda->GetFunctionArguments();
+  for (size_t n = 0; n < args.size(); n++)
```
phate, line 232: Could we get an assert to ensure that `args.size() == newArgs.size()`?

jlm/llvm/ir/operators/lambda.cpp
```diff
@@ -399,51 +371,16 @@ node::IsExported(const lambda::node & lambdaNode)
...
-cvargument::~cvargument() = default;
-
-cvargument &
-cvargument::Copy(rvsdg::Region & region, jlm::rvsdg::structural_input * input)
+[[nodiscard]] const jlm::llvm::attributeset &
+node::GetArgumentAttributes(const rvsdg::output & argument) const noexcept
 {
-  auto lambdaInput = util::AssertedCast<lambda::cvinput>(input);
-  return *cvargument::create(&region, lambdaInput);
+  return argumentAttributes_[argument.index()];
```
phate, line 377: Could we get an assert that `index < argumentAttributes_.size()`

jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -401,302 +423,9 @@ class node final : public jlm::rvsdg::structural_node
...
-
-  [[nodiscard]] const lambda::result *
-  next() const override
-  {
-    auto index = value()->index();
-    auto lambda = jlm::util::AssertedCast<lambda::node>(value()->region()->node());
-
-    return lambda->nfctresults() > index + 1 ? lambda->fctresult(index + 1) : nullptr;
-  }
+  std::vector<jlm::llvm::attributeset> argumentAttributes_;
```
phate, line 428: `ArgumentAttributes_`

jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
```diff
@@ -14,7 +14,7 @@ namespace jlm::hls
 {
 
 static bool
-IsPassthroughArgument(const rvsdg::RegionArgument & argument)
+IsPassthroughArgument(const rvsdg::output & argument)
```
phate, line 17: Why did you make it more general? I understand that this most likely comes from the decision that you made the argument in the ContextVar and rvsdg::output, but why? We should always give the most specific type back, no?

jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
```diff
@@ -25,7 +25,7 @@ IsPassthroughArgument(const rvsdg::RegionArgument & argument)
 }
 
 static bool
-IsPassthroughResult(const rvsdg::RegionResult & result)
+IsPassthroughResult(const rvsdg::input & result)
```
phate, line 28: same here

phate (changes requested):
I stopped reviewing after I saw the consequences of giving back the most general type for the argument instead of a RegionArgument in the ContextVar struct. Why? Should we not always give back the most specific type?

phate:
@Felix-Rm : FYI

jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -133,51 +129,52 @@ class node final : public jlm::rvsdg::structural_node
...
+    rvsdg::input * input;
+
+    /**
+     * \brief Access to bound object in subregion.
+     *
+     * Supplies access to the value bound into the lambda abstraction
+     * from inside the region contained in the lambda node. This
+     * evaluates to the value bound into the lambda.
+     */
+    rvsdg::output * inner;
```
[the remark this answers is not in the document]
caleridas: I originally had that, but actually ran into several inconveniences. I think it is better to consistently return a single type:

1. Callers cannot "do" anything with this information if they are given anything more specific. The primary utility of any def site is that it can be connected to a use site, and for that nothing else is needed. However, if caller wants to do any tracing through the graph, RegionArgument is insufficient and they need to know what exactly it is that is connected and how. So, in practice, RegionArgument does not gain anything.

2. Sites expecting lists of def sites usually expect std::vector<rvsdg::output *>. There are several places in code that need iteration simply to convert from "collection of RegionArgument *" to "collection of rvsdg::output *" as a consequence of being overly specific.

3. Long-term I am not even certain that keeping output / node_output / RegionArgument as class hierarchy makes too much sense. What they are is effectively:

```
struct NodeInput { Node * owner; size_t index; };
struct RegionArgument { Region * owner; size_t index; };
```

One can have `Node` basetype with virtual method `std::variant<Node *, Region *> GetOwner()` (or dynamic_cast or anything like that), but in effect it is mostly wasted overhead in dynamic dispatch, and it is both more memory and time-efficient to just declare it an alternative of the two to begin with. (Besides codifying the fact that this is also the exhaustive list of options).

jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -133,51 +129,52 @@ class node final : public jlm::rvsdg::structural_node
...
+  };
+
+  [[nodiscard]] std::vector<rvsdg::output *>
```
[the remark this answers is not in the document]
caleridas: As above, and particularly here my comment region "collection of RegionArgument *" to "collection of output *" applies.

jlm/llvm/ir/operators/lambda.hpp
caleridas: now I am confused -- I thought functions and classes UpperCamelCase and variables lowerCamelCase, no?

jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -380,7 +402,7 @@ class node final : public jlm::rvsdg::structural_node
    *
    * \return The output of the lambda node.
    */
-  lambda::output *
+  rvsdg::output *
```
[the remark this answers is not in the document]
caleridas: There are 179 call sites, is this really worth it doing now? I would rather figure a way to restructure the construction API (the "builder" pattern as used for phi actually looks very suitable) instead of causing more churn.

jlm/llvm/ir/operators/lambda.hpp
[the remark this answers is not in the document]
phate: All right, let's leave it like that then.

jlm/llvm/ir/operators/lambda.hpp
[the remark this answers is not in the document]
phate: Okay.

jlm/llvm/ir/operators/lambda.hpp
phate: Function and Classes: UpperCamelCase
Local Variables: lowerCamelCase
Class Attributes: UpperCamelCaseWith_

jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
```diff
@@ -639,16 +639,16 @@ jlm::hls::MemoryConverter(jlm::llvm::RvsdgModule & rm)
...
-    smap.insert(
-        lambda->cvargument(i),
-        newLambda->add_ctxvar(lambda->cvargument(i)->input()->origin()));
+    smap.insert(ctxvar.inner, newLambda->AddContextVar(ctxvar.input->origin()).inner);
   }
 
-  for (size_t i = 0; i < lambda->nfctarguments(); ++i)
+  auto args = lambda->GetFunctionArguments();
+  auto newArgs = newLambda->GetFunctionArguments();
+  for (size_t i = 0; i < args.size(); ++i)
```
phate, line 649: Could we get an assert here: `JLM_ASSERT(args.size() == newArgs.size())`

jlm/hls/backend/rvsdg2rhls/remove-unused-state.cpp
```diff
@@ -205,28 +205,29 @@ remove_lambda_passthrough(llvm::lambda::node * ln)
...
-    smap.insert(ln->cvargument(i), new_lambda->add_ctxvar(ln->cvargument(i)->input()->origin()));
+    // copy over context vars
+    smap.insert(ctxvar.inner, new_lambda->AddContextVar(ctxvar.input->origin()).inner);
   }
 
   size_t new_i = 0;
-  for (size_t i = 0; i < ln->nfctarguments(); ++i)
+  auto args = ln->GetFunctionArguments();
+  auto new_args = new_lambda->GetFunctionArguments();
+  for (size_t i = 0; i < args.size(); ++i)
```
phate, line 217: assert for `args.size() == new_args.size()`

jlm/hls/backend/rvsdg2rhls/rvsdg2rhls.cpp
```diff
@@ -287,27 +287,29 @@ change_linkage(llvm::lambda::node * ln, llvm::linkage link)
...
+    auto origin = cv.input->origin();
+    auto newcv = lambda->AddContextVar(origin);
+    subregionmap.insert(cv.inner, newcv.inner);
   }
 
   /* collect function arguments */
-  for (size_t n = 0; n < ln->nfctarguments(); n++)
+  auto args = ln->GetFunctionArguments();
+  auto newArgs = lambda->GetFunctionArguments();
+  for (size_t n = 0; n < args.size(); n++)
```
phate, line 300: assert

phate (changes requested):
I had another go at it with the premise in mind that we leave it `rvsdg::output` instead of `RegionArgument`. I did not find much more any longer.

jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
caleridas: yes the intent is to get rid of RegionResult (it doesn't really contribute anything besides being a redundant type)

jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
caleridas: added assert here as well as in the other places

jlm/llvm/ir/operators/lambda.hpp
caleridas: renamed

jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -133,51 +129,52 @@ class node final : public jlm::rvsdg::structural_node
...
+     * from inside the region contained in the lambda node. This
+     * evaluates to the value bound into the lambda.
+     */
+    rvsdg::output * inner;
+  };
+
+  [[nodiscard]] std::vector<rvsdg::output *>
+  GetFunctionArguments() const;
+
+  [[nodiscard]] std::vector<rvsdg::input *>
```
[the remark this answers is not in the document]
caleridas: see other comment regarding types

caleridas:
> I stopped reviewing after I saw the consequences of giving back the most general type for the argument instead of a RegionArgument in the ContextVar struct. Why? Should we not always give back the most specific type?

The issue with "expose most specific type" is that it tends to also expose implementation details, making it harder to change implementation details. My usual preference is "expose something implementing the promised interface", question is what "promised interface" is. Looking at the code, the best use for all of RegionArgument is to be used as originators of graph edges, which is exactly rvsdg::output. The place where RegionArgument is of use is when analyzing the graph to determine "what produces an output". However, here RegionArgument is wholly insufficient and does not allow to do anything of use, you generally need to know more specifically "argument of theta" or "argument of lambda" -- but even that is not enough as you usually need more specific semantic information of where the value came from on the outside. That is not representable in the type system, so I think it is not of much use being specific here.

jlm/llvm/ir/operators/lambda.cpp
caleridas: Yes it is more expensive than it could be, I am not sure "GetMemoryStateRegionArgument" should be part of the generic lambda API (it is a very specific thing corresponding to a very specific encoding). I would leave the inefficiency for the moment until moving lambda to generic rvsdg and deciding how to proceed with the encoding.

caleridas:
@sjalander I need help figuring out why/how this change affects HLS. In the precommit run I am getting:

Running: dynamatic/test_memory_5
make: *** Waiting for unfinished jobs....
    SUCCESS
    The execution time has changed
        Golden cycle time: 28
        Simulated cycles: 29

However I have no tests in the jlm repository responsible for this and also no context as to what kind of change would cause this. From my POV the change should be semantically no-op, but I am not sure what exactly is happening.

caleridas:
> @sjalander I need help figuring out why/how this change affects HLS. In the precommit run I am getting:
> 
> Running: dynamatic/test_memory_5 make: *** Waiting for unfinished jobs.... SUCCESS The execution time has changed Golden cycle time: 28 Simulated cycles: 29
> 
> However I have no tests in the jlm repository responsible for this and also no context as to what kind of change would cause this. From my POV the change should be semantically no-op, but I am not sure what exactly is happening.

I also have difficulty reproducing this locally. AFAICT the tests already fail for me at master, so I am not certain what is expected.

sjalander:
@caleridas @phate @haved 
I'm not sure why this is happening. I tried the branch, and dynamatic/test_memory_5 is the only test case that is impacted with an increase of a single cycle. The test of simulated cycles was added to avoid missing changes that can have significant impact on the performance of HLS. Since this PR doesn't have a significant impact, I went ahead and created a PR for the hls_test_suite that updates the golden reference.

However, @caleridas comment about it not working locally for master makes me wonder if something more is going on. Since this branch has conflicts with master, it would be preferable to get the branch in sync with master and then rerun the test suite to verify that change in cycles, before merging the PR for the HLS test suite.

@caleridas 
When you try to run it locally on master, do you use ./scripts/run-hls-test.sh to run the tests?
(This requires jlm built with HLS enabled, a build of CIRCT [./scripts/build-circt.sh], and Verilator installed.)

caleridas:
> @caleridas @phate @haved I'm not sure why this is happening. I tried the branch, and dynamatic/test_memory_5 is the only test case that is impacted with an increase of a single cycle. The test of simulated cycles was added to avoid missing changes that can have significant impact on the performance of HLS. Since this PR doesn't have a significant impact, I went ahead and created a PR for the hls_test_suite that updates the golden reference.
> 
> However, @caleridas comment about it not working locally for master makes me wonder if something more is going on. Since this branch has conflicts with master, it would be preferable to get the branch in sync with master and then rerun the test suite to verify that change in cycles, before merging the PR for the HLS test suite.
> 
> @caleridas When you try to run it locally on master, do you use ./scripts/run-hls-test.sh to run the tests? (This requires jlm built with HLS enabled, a build of CIRCT [./scripts/build-circt.sh], and Verilator installed.)

Thank you, if that resolution is fine with you I am very glad -- my (unproven) suspicion is that something may quite subtly depend on iteration order or similar somewhere, but it is both very hard to find and at the same time very hard to preserve.

Regarding local reproduction, I managed to trace it down to something unrelated. I have verilator & everything installed routinely anyways, so I can also run it.

sjalander:
@caleridas 
There's now a new commit in the HLS test suite that you can use to update the scripts/run-hls-test.sh

99d309be2a9aa8d565c2ece493dc33a447ad166d

jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -305,13 +327,13 @@ class node final : public jlm::rvsdg::structural_node
   /**
    * @return The memory state argument of the lambda subregion.
    */
-  [[nodiscard]] rvsdg::RegionArgument &
+  [[nodiscard]] rvsdg::output &
```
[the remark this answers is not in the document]
caleridas: see other comments, normalizing to a single type

jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -305,13 +327,13 @@ class node final : public jlm::rvsdg::structural_node
...
   GetMemoryStateRegionArgument() const noexcept;
 
   /**
    * @return The memory state result of the lambda subregion.
    */
-  [[nodiscard]] rvsdg::RegionResult &
+  [[nodiscard]] rvsdg::input &
```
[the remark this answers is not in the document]
caleridas: normalizing to single type

jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -221,43 +218,80 @@ class node final : public jlm::rvsdg::structural_node
...
+   * \param origin
+   *   The value to be bound into the lambda node.
+   *
+   * \pre
+   *   \p origin must be from the same region as the lambda node.
+   *
+   * \return The context variable argument of the lambda abstraction.
+   */
+  ContextVar
+  AddContextVar(jlm::rvsdg::output * origin);
```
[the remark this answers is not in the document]
caleridas: AFAICT > 100 call sites that are mostly of the form:

auto value = &jlm::tests::GraphImport::Create(rvsdg, valueType, "x"); // this makes a pointer, or we get the pointer out of a structure that can only hold pointers
lambdaNode->AddContextVar(*value); // making a reference out of the pointer again

changing it in a separate commit in this branch

jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -133,51 +129,52 @@ class node final : public jlm::rvsdg::structural_node
...
-  using fctresult_constrange = jlm::util::iterator_range<fctresconstiterator>;
-
-public:
   ~node() override;
 
 private:
-  node(rvsdg::Region * parent, lambda::operation && op)
-      : structural_node(op, parent, 1)
-  {}
+  node(rvsdg::Region * parent, lambda::operation && op);
```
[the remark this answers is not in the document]
caleridas: Also changed in same commit -- however the real "meat" is changing the "create" methods which also eventually need to be renamed "Create". Leaving that out for now, though.

caleridas:
> @caleridas There's now a new commit in the HLS test suite that you can use to update the scripts/run-hls-test.sh
> 
> 99d309be2a9aa8d565c2ece493dc33a447ad166d

Thanks a lot, this seems to work!

jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -218,43 +215,80 @@ class node final : public rvsdg::StructuralNode
...
+   * \pre
+   *   \p output must be an argument to the subregion of this node
+   *
+   * Returns the context variable description corresponding
+   * to this bound variable reference in the lambda node region.
+   * Note that some arguments of the region are formal call arguments
+   * and do not have an associated context variable description.
+   */
+  [[nodiscard]] std::optional<ContextVar>
+  MapBinderContextVar(const rvsdg::output & output) const noexcept;
```
phate, line 271: The same here, a comment would be nice that states that the owner of the output is the lambda subregion.

phate (approved):
I only have minor comments left. I won't block this because of them any longer. I want this to go in. Your choice whether you address them.

jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -218,43 +215,80 @@ class node final : public rvsdg::StructuralNode
...
+   *   \p input must be input to this node.
+   *
+   * Returns the context variable description corresponding
+   * to this input of the lambda node. All inputs to the lambda
+   * node are by definition bound context variables that are
+   * accessible in the subregion through the corresponding
+   * argument.
+   */
+  [[nodiscard]] ContextVar
+  MapInputContextVar(const rvsdg::input & input) const noexcept;
```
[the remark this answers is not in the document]
caleridas: That is already documented, it is the "\pre" section in the doxygen comment above.

jlm/llvm/ir/operators/lambda.hpp
caleridas: Same, there is already a "\pre" doxygen section.

jlm/llvm/ir/operators/call.hpp
```diff
@@ -219,7 +218,7 @@ class CallTypeClassifier final
   }
 
   static std::unique_ptr<CallTypeClassifier>
-  CreateNonRecursiveDirectCallClassifier(lambda::output & output)
+  CreateNonRecursiveDirectCallClassifier(rvsdg::output & output)
   {
     return std::make_unique<CallTypeClassifier>(CallType::NonRecursiveDirectCall, output);
```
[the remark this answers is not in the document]
caleridas: Added assert as well as documentation for pre-condition. I see that the only call site is actually "ClassifyCall", and that does exactly the same checks -- I guess compiler will be able to elide them, but since these convenience helpers are trivial might make sense to dissolve them into the caller.

jlm/llvm/ir/operators/lambda.cpp
```diff
@@ -41,133 +41,97 @@ operation::copy() const
...
+  auto numArguments = GetOperation().Type()->NumArguments();
+  if (output.index() >= numArguments)
+  {
+    return ContextVar{ input(output.index() - GetOperation().Type()->NumArguments()),
+                       const_cast<rvsdg::output *>(&output) };
+  }
+  else
+  {
+    return std::nullopt;
+  }
```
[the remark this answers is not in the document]
caleridas: okay, do you prefer it that way? I usually prefer it the other way (make if/else cascades "exhaustive", i.e. "if one branch returns, then all do" -- in contrast to having the case that the control flow is joined back together in the end).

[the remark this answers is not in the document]
phate: I normally just exhaustively write it when it is more than two options, i.e., if-then-else if-....However, as stated in the comment, it is a minor thing and certainly personal preference.

phate (approved):
Thanks for pushing this through.

PR #659, opened 2024-10-15
Adds documentation and unit tests for MemoryConverter

sjalander:


jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
```diff
@@ -28,9 +28,10 @@ jlm::hls::route_response(rvsdg::Region * target, jlm::rvsdg::output * response)
   else
   {
     auto parent_response = route_response(target->node()->region(), response);
-    auto ln = dynamic_cast<jlm::hls::loop_node *>(target->node());
-    JLM_ASSERT(ln);
-    auto input = jlm::rvsdg::structural_input::create(ln, parent_response, parent_response->Type());
+    auto lambda = dynamic_cast<jlm::hls::loop_node *>(target->node());
+    JLM_ASSERT(lambda);
```
phate, line 32: Why do you call the variable `lambda`, when it is a hls::loop_node?

phate, line 32: You could use the `AssertedCast<>()` here, then you could throw out the `JLM_ASSERT()`

jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
```diff
@@ -45,9 +46,9 @@ jlm::hls::route_request(rvsdg::Region * target, jlm::rvsdg::output * request)
   }
   else
   {
-    auto ln = dynamic_cast<jlm::hls::loop_node *>(request->region()->node());
-    JLM_ASSERT(ln);
-    auto output = jlm::rvsdg::structural_output::create(ln, request->Type());
+    auto lambda = dynamic_cast<jlm::hls::loop_node *>(request->region()->node());
+    JLM_ASSERT(lambda);
```
phate, line 50: The same here: `lambda` and `AssertedCast<>()`

jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
```diff
@@ -464,6 +470,7 @@ IsDecoupledFunctionPointer(
   visited.insert(output);
 
   bool isDecoupled = false;
+  // Iterate through all users of the output
```
phate, line 473: That comment adds nothing to the code line below, no? The code line below says exactly the same.

jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
```diff
@@ -710,6 +717,11 @@ jlm::hls::MemoryConverter(jlm::llvm::RvsdgModule & rm)
   // once.
   //
   remove_unused_state(root);
+
+  // Need to get the lambda from the root since remote_unused_state replaces the lambda
+  JLM_ASSERT(root->nnodes() == 1);
+  newLambda = jlm::util::AssertedCast<jlm::llvm::lambda::node>(root->nodes.first());
```
phate, line 723: Could you use `*root->Nodes().begin()` instead of `root->nodes.first()`. The plan is to privatize the nodes attribute in the future.

tests/jlm/hls/backend/rvsdg2rhls/MemoryConverterTests.cpp
```diff
@@ -0,0 +1,385 @@
...
+{
+  TestTraceArgument();
+  TestLoad();
+  TestLoadStore();
+  TestThetaLoad();
+
+  return 0;
+}
+
+JLM_UNIT_TEST_REGISTER("jlm/hls/backend/rvsdg2rhls/MemoryConverterTests", Tests)
```
phate, line 385: Could you register each test individually with `JLM_UNIT_TEST_REGISTER`, please. You can remove the `Tests()` method and this registration then afterwards.

phate (changes requested):
Looks good. Only minor comments.

jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
sjalander: My C++ skills make using the API frustrating. Changing to what you suggest gives the following error:
./jlm/util/common.hpp:51:1: note: candidate template ignored: could not match 'From *' against 'jlm::rvsdg::node'
   51 | AssertedCast(From * value)

jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
sjalander: This was from an overly aggressive use of the IDE's refactoring. 
'ln' is used in other places for lambda node.

sjalander: Since this code was not supposed to be touched, I've not made the suggested change.

jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
sjalander: Same as above

jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
sjalander: With my skills, it's not obvious that dereferencing an output gives you it's users. So this comment helps a user like me.
I've therefore kept it, but if you want I can remove it.

jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
phate: Try: `root->Nodes().begin().ptr()`

tests/jlm/hls/backend/rvsdg2rhls/MemoryConverterTests.cpp
phate: Maybe move the registration to the respective function so they are together and one does not need to search for it.

jlm/hls/backend/rvsdg2rhls/rvsdg2rhls.cpp
```diff
@@ -417,25 +417,41 @@ rvsdg2rhls(llvm::RvsdgModule & rhls)
...
+  dump_dot(rhls, "12-memstate_conv.dot");
   remove_redundant_buf(rhls);
+  dump_dot(rhls, "13-redundant.dot");
   // enforce 1:1 input output relationship
   add_sinks(rhls);
+  dump_dot(rhls, "14-sinks.dot");
   add_forks(rhls);
+  dump_dot(rhls, "15-forks.dot");
   add_buffers(rhls, true);
+  dump_dot(rhls, "16-buffers.dot");
```
phate, line 454: Where does this all in a sudden come from?

phate (changes requested):

tests/jlm/hls/backend/rvsdg2rhls/MemoryConverterTests.cpp
sjalander: Moved

jlm/hls/backend/rvsdg2rhls/rvsdg2rhls.cpp
sjalander: Some other work sneaked in.  Been removed.

jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
sjalander: Works

phate (approved):

PR #661, opened 2024-10-16
Updated caching for CI CIRCT build

sjalander:
The 'save-always: true' option is deprecated.
This version should have the advantage that even if the CI fails at a later step, the build should have already been saved and not require to be rebuilt when a fix is pushed.

### DIFF
--- a/.github/actions/BuildCirct/action.yml
+++ b/.github/actions/BuildCirct/action.yml
@@ -26,7 +26,7 @@ runs:
         key: ${{ runner.os }}-circt-${{ steps.get-circt-hash.outputs.hash }}
 
     - name: "Build CIRCT if we didn't hit in the cache"
-      if: steps.cache-circt.outputs.cache-hit != 'true'
+      if: steps.restore-cache-circt.outputs.cache-hit != 'true'
       run: |
         ./scripts/build-circt.sh \
           --build-path ${{ github.workspace }}/build-circt \
--- a/.github/actions/BuildCirct/action.yml
+++ b/.github/actions/BuildCirct/action.yml
@@ -4,6 +4,13 @@ description: "Builds CIRCT, which is used for the HLS backend"
 runs:
   using: "composite"
   steps:
+    - name: "Install LLVM, Clang, MLIR, and Ninja"
+      uses: ./.github/actions/InstallPackages
+      with:
+        install-llvm: true
+        install-mlir: true
+        install-ninja: true
+
     - name: "Get the commit used for building CIRCT and use it as the cache key"
       id: get-circt-hash
       run: |
@@ -18,13 +25,6 @@ runs:
           ${{ github.workspace }}/build-circt/circt
         key: ${{ runner.os }}-circt-${{ steps.get-circt-hash.outputs.hash }}
 
-    - name: "Install LLVM, Clang, MLIR, and Ninja"
-      uses: ./.github/actions/InstallPackages
-      with:
-        install-llvm: true
-        install-mlir: true
-        install-ninja: true
-
     - name: "Build CIRCT if we didn't hit in the cache"
       if: steps.cache-circt.outputs.cache-hit != 'true'
       run: |
@@ -35,10 +35,9 @@ runs:
       shell: bash
 
     - name: "Save CIRCT to the cache"
-      if: steps.cache-circt.outputs.cache-hit != 'true'
       id: save-cache-circt
       uses: actions/cache/save@v4
       with:
         path: |
           ${{ github.workspace }}/build-circt/circt
-        key: ${{ runner.os }}-circt-${{ steps.get-circt-hash.outputs.hash }}
+        key: ${{ steps.restore-cache-circt.outputs.cache-primary-key }}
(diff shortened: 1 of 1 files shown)


sjalander:
The Tests/build reported the following, so attempts are at least made to save the build.
This version should have the advantage that even if the CI fails at a later stage, the build should have already been saved and not require to be rebuilt when a fix is pushed.

Failed to save: Unable to reserve cache with key Linux-circt-2dc8240d91a0f993d616b152aa4d7520156862fe, another job may be creating this cache. More details: Cache already exists. Scope: refs/pull/661/merge, Key: Linux-circt-2dc8240d91a0f993d616b152aa4d7520156862fe, Version: 238d06c864bc65ace87e9f10bc2b1cb6c325b420ebcb72aafc883eb2332e7ee7
Warning: Cache save failed.

.github/actions/BuildCirct/action.yml
```diff
@@ -4,17 +4,18 @@ description: "Builds CIRCT, which is used for the HLS backend"
 runs:
   using: "composite"
   steps:
+    - uses: actions/checkout@v4
+
```
phate, line 8: Why is this all in a sudden needed here?

.github/actions/BuildCirct/action.yml
```diff
@@ -34,3 +35,11 @@ runs:
           --install-path ${{ github.workspace }}/build-circt/circt \
           --llvm-lit-path ~/.local/bin/lit
       shell: bash
+
+    - name: "Save CIRCT to the cache"
+      id: save-cache-circt
+      uses: actions/cache/save@v4
```
phate, line 41: This should only happen if we did not have a cache hit in the first place, right? In other words, we do not need to save anything if we found it in the cache.

.github/actions/BuildCirct/action.yml
sjalander: This is a copy paste mistake from the example I used. I'll remove it.

.github/actions/BuildCirct/action.yml
sjalander: The example I followed didn't make the check. It's more or less instantaneous so doesn't really hurt.

phate: I was not concerned about performance, but more about the fact that you are always overwriting the old cache even if it there was a hit in the cache.

.github/actions/BuildCirct/action.yml
sjalander: Removed

.github/actions/BuildCirct/action.yml
sjalander: I simply follow the example given by the cache action

https://github.com/actions/cache/blob/main/README.md#example-cache-workflow

As far as what I understand, if the save finds the key in the cache, then the cache is not updated.

.github/actions/BuildCirct/action.yml
```diff
@@ -4,28 +4,27 @@ description: "Builds CIRCT, which is used for the HLS backend"
 runs:
   using: "composite"
   steps:
+    - name: "Install LLVM, Clang, MLIR, and Ninja"
```
sjalander, line 7: I moved this first as it is independent of the caching. 
This keeps the cache steps together, which should make things clearer.

phate:
```
 NEEDRESTART-KSTA: 1
Run echo "hash=$(./scripts/build-circt.sh --get-commit-hash)" >> $GITHUB_OUTPUT
Run actions/cache/restore@v4
Received 0 of 540379106 (0.0%), 0.0 MBs/sec
Received 125829120 of 540379106 (23.3%), 59.9 MBs/sec
Received 268435456 of 540379106 (49.7%), 85.3 MBs/sec
Received 427819008 of 540379106 (79.2%), 102.0 MBs/sec
Cache Size: ~515 MB (540379106 B)
/usr/bin/tar -xf /home/runner/work/_temp/528636d8-afe7-4f39-bab5-22cd073bf6fa/cache.tzst -P -C /home/runner/work/jlm/jlm --use-compress-program unzstd
Received 540379106 of 540379106 (100.0%), 103.0 MBs/sec
Cache restored successfully
Cache restored from key: Linux-circt-2dc8240d91a0f993d616b152aa4d7520156862fe
Run ./scripts/build-circt.sh \
Cloning into '/home/runner/work/jlm/jlm/build-circt/circt.git'...
Note: switching to '2dc8240d91a0f993d616b152aa4d7520156862fe'.
```

In the logs, it says that it successfully restored the cache, but then it goes ahead and builds CIRCT anyway. Is it not the point that if we have a cache hit, that we can avoid the rebuilding of CIRCT?

.github/actions/BuildCirct/action.yml
phate: My understanding is that the rebuilding of CIRCT should be skipped when there is a cache hit: https://github.com/actions/cache/blob/main/README.md#skipping-steps-based-on-cache-hit

However, I only skimmed the documentation, so I might have misunderstood.

phate (approved):

PR #662, opened 2024-11-12
Roundtrip alloca and load/store non volatile 

halvorlinder:


haved:
Closing as this is replaced by #781

PR #663, opened 2024-11-16
Privatize top nodes in Region class

phate:


haved (approved):
Nice, the added documentation makes it much clearer what top nodes are

PR #673, opened 2024-11-30
Rename node class to Node

phate:


caleridas (dismissed):
Only looked at node.hpp/node.cpp and spot-checking others, since it is just mechanical rename. Nothing much to say, rename is logical.

haved (approved):

PR #674, opened 2024-12-01
gamma: change API for mapping input to region argument

caleridas:
Remove the GammaInput class and the way it is used to map an input to the corresponding argument inside the lambda region. Replace this with an API function attached to the node instead.

### DIFF
--- a/jlm/mlir/frontend/MlirToJlmConverter.cpp
+++ b/jlm/mlir/frontend/MlirToJlmConverter.cpp
@@ -335,7 +335,7 @@ MlirToJlmConverter::ConvertOperation(
     // Add inputs to the gamma node and to all it's subregions
     for (size_t i = 1; i < inputs.size(); i++)
     {
-      rvsdgGammaNode->add_entryvar(inputs[i]);
+      rvsdgGammaNode->AddEntryVar(inputs[i]);
     }
 
     ::llvm::SmallVector<::llvm::SmallVector<jlm::rvsdg::output *>> regionResults;
@@ -355,7 +355,7 @@ MlirToJlmConverter::ConvertOperation(
         JLM_ASSERT(regionResults[regionIndex].size() == regionResults[0].size());
         exitvars.push_back(regionResults[regionIndex][exitvarIndex]);
       }
-      rvsdgGammaNode->add_exitvar(exitvars);
+      rvsdgGammaNode->AddExitVar(exitvars);
     }
 
     return rvsdgGammaNode;
--- a/jlm/rvsdg/gamma.cpp
+++ b/jlm/rvsdg/gamma.cpp
@@ -29,13 +29,13 @@ perform_predicate_reduction(GammaNode * gamma)
   auto alternative = cop->value().alternative();
 
   rvsdg::SubstitutionMap smap;
-  for (auto it = gamma->begin_entryvar(); it != gamma->end_entryvar(); it++)
-    smap.insert(it->argument(alternative), it->origin());
+  for (const auto & ev : gamma->GetEntryVars())
+    smap.insert(ev.branchArgument[alternative], ev.input->origin());
 
   gamma->subregion(alternative)->copy(gamma->region(), smap, false, false);
 
-  for (auto it = gamma->begin_exitvar(); it != gamma->end_exitvar(); it++)
-    it->divert_users(smap.lookup(it->result(alternative)->origin()));
+  for (auto exitvar : gamma->GetExitVars())
+    exitvar.output->divert_users(smap.lookup(exitvar.branchResult[alternative]->origin()));
 
   remove(gamma);
 }
@@ -44,24 +44,25 @@ static bool
 perform_invariant_reduction(GammaNode * gamma)
 {
   bool was_normalized = true;
-  for (auto it = gamma->begin_exitvar(); it != gamma->end_exitvar(); it++)
+  for (auto exitvar : gamma->GetExitVars())
   {
-    auto argument = dynamic_cast<const rvsdg::RegionArgument *>(it->result(0)->origin());
+    auto argument = dynamic_cast<const rvsdg::RegionArgument *>(exitvar.branchResult[0]->origin());
     if (!argument)
       continue;
 
     size_t n;
     auto input = argument->input();
-    for (n = 1; n < it->nresults(); n++)
+    for (n = 1; n < exitvar.branchResult.size(); n++)
     {
-      auto argument = dynamic_cast<const rvsdg::RegionArgument *>(it->result(n)->origin());
+      auto argument =
+          dynamic_cast<const rvsdg::RegionArgument *>(exitvar.branchResult[n]->origin());
       if (!argument && argument->input() != input)
         break;
     }
 
-    if (n == it->nresults())
+    if (n == exitvar.branchResult.size())
     {
-      it->divert_users(argument->input()->origin());
+      exitvar.output->divert_users(argument->input()->origin());
       was_normalized = false;
     }
   }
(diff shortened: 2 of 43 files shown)


jlm/rvsdg/gamma.cpp
```diff
@@ -265,79 +265,130 @@ GammaOperation::operator==(const operation & other) const noexcept
...
-  size_t n;
-  auto origin = argument->input()->origin();
-  for (n = 1; n < nresults(); n++)
+  return ev;
+}
+
+GammaNode::EntryVar
+GammaNode::GetEntryVar(std::size_t index) const
+{
+  EntryVar ev;
```
phate, line 299: Could we get an assert() checking that the index is within bounds?

jlm/rvsdg/gamma.cpp
```diff
@@ -265,79 +265,130 @@ GammaOperation::operator==(const operation & other) const noexcept
...
+  if (values.size() != nsubregions())
+    throw jlm::util::error("Incorrect number of values.");
 
-  auto node = input_->node();
-  auto index = input_->index();
-  if (index == node->ninputs() - 1)
+  const auto & type = values[0]->Type();
+  auto output = node::add_output(std::make_unique<rvsdg::StructuralOutput>(this, type));
+
+  std::vector<rvsdg::input *> branch_results;
```
phate, line 343: camelCase

jlm/rvsdg/gamma.cpp
```diff
@@ -265,79 +265,130 @@ GammaOperation::operator==(const operation & other) const noexcept
...
+  for (size_t n = 0; n < nsubregions(); n++)
   {
-    input_ = nullptr;
-    return *this;
+    branch_results.push_back(&rvsdg::RegionResult::Create(
+        *subregion(n),
+        *values[n],
+        static_cast<rvsdg::StructuralOutput *>(output),
```
phate, line 349: Move the cast up to the creation of output.

jlm/rvsdg/gamma.cpp
```diff
@@ -265,79 +265,130 @@ GammaOperation::operator==(const operation & other) const noexcept
...
-GammaNode::exitvar_iterator::operator++() noexcept
+std::vector<GammaNode::ExitVar>
+GammaNode::GetExitVars() const
 {
-  if (output_ == nullptr)
-    return *this;
+  std::vector<GammaNode::ExitVar> vars;
+  for (size_t n = 0; n < noutputs(); ++n)
+  {
+    std::vector<rvsdg::input *> branch_results;
```
phate, line 362: camelCase

jlm/rvsdg/gamma.cpp
```diff
@@ -265,79 +265,130 @@ GammaOperation::operator==(const operation & other) const noexcept
...
+  return vars;
+}
 
-  auto node = output_->node();
-  auto index = output_->index();
-  if (index == node->nexitvars() - 1)
+GammaNode::ExitVar
+GammaNode::MapOutputExitVar(const rvsdg::output & output) const
+{
+  std::vector<rvsdg::input *> branch_results;
```
phate, line 375: camelCase

phate, line 375: Some assert that output is the output of a gamma node.

jlm/rvsdg/gamma.cpp
```diff
@@ -265,79 +265,130 @@ GammaOperation::operator==(const operation & other) const noexcept
...
   }
+  return ExitVar{ std::move(branch_results), node::output(output.index()) };
+}
 
-  output_ = node->exitvar(++index);
-  return *this;
+GammaNode::ExitVar
+GammaNode::MapBranchResultExitVar(const rvsdg::input & input) const
+{
+  std::vector<rvsdg::input *> branch_results;
```
phate, line 386: Some assert that the input is a result of the region

phate, line 386: camelCase

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -355,7 +355,7 @@ MlirToJlmConverter::ConvertOperation(
         JLM_ASSERT(regionResults[regionIndex].size() == regionResults[0].size());
         exitvars.push_back(regionResults[regionIndex][exitvarIndex]);
       }
-      rvsdgGammaNode->add_exitvar(exitvars);
+      rvsdgGammaNode->AddExitVar(exitvars).output;
```
phate, line 358: The `.output` is dead code, no?

jlm/rvsdg/gamma.hpp
```diff
@@ -117,180 +116,184 @@ class GammaNode : public StructuralNode
...
+  struct EntryVar
+  {
+    /**
+     * \brief Variable at entry point (input to gamma node).
+     */
+    rvsdg::input * input;
+    /**
+     * \brief Variable inside each of the branch regions (argument in per subregion).
+     */
+    std::vector<rvsdg::output *> branches;
```
phate, line 132: I really do not like the name `branches`. I find it very misleading. What about `subregionArgument` or even `branchArgument`?

jlm/rvsdg/gamma.hpp
```diff
@@ -117,180 +116,184 @@ class GammaNode : public StructuralNode
...
-    GammaOutput * output_;
+  /**
+   * \brief A variable routed out of all gamma regions as result.
+   */
+  struct ExitVar
+  {
+    /**
+     * \brief Variable exit points (results per subregion).
+     */
+    std::vector<rvsdg::input *> branches;
```
phate, line 143: Same here.

jlm/rvsdg/gamma.cpp
```diff
@@ -265,79 +265,130 @@ GammaOperation::operator==(const operation & other) const noexcept
   return op && op->nalternatives_ == nalternatives_;
 }
 
-/* gamma input */
-
-GammaInput::~GammaInput() noexcept = default;
+/* gamma node */
```
phate, line 268: Since you are already touching it, it can be removed....it adds no value.

phate (changes requested):

jlm/rvsdg/gamma.hpp
phate: Or any other ideas? If everything fails, we can use `branches`, but I am not in favor or it.

jlm/mlir/frontend/MlirToJlmConverter.cpp
phate: The CI complains about it as well.

jlm/rvsdg/gamma.hpp
caleridas: Some names that I came up with

- branchVar (this is the vesion of the variable in the respective branches)
- branchVariable, same as branchVar just longer
- branchDef (based on eventually renaming output->Def and input->Use)
- branchArg, shorter than branchArgument
- perBranch: actually quite accurately describes what it is, but at the same time sounds awkward

initially I had some dislike for the term "argument", but actually on thinking more about it "branchArg(ument)" (and "branchRes(ult)" for the converse sounds suitable -- generally I prefer not to abbreviate and would use the full names, but not certain. WDYT?

caleridas: changed it branchArgument & branchResult, I think that is appropriate

phate: branchArgument and branchResult sounds great.

phate (dismissed):
There are still two open issues from the last review round. I will not stop this PR on them though.

jlm/rvsdg/gamma.cpp
caleridas: added assertion

caleridas:
> There are still two open issues from the last review round. I will not stop this PR on them though.

I think I just addressed the two open issues while you were posting this -- I think you need to re-approve.

phate (approved):

PR #677, opened 2024-12-05
Add new node normalization interface and convert load reductions to it

phate:
This PR is a series of PRs to remove the old node normalization interface that is intertwined in the RVSDG library. This PR does the following:

1. Adds a new and simple node normalization interface for optimizing the operands of a node. This enables to create individual instances of normalizations and therefore pass them around, and/or configure optimization by simply providing these instances instead of utilizing the flags of the old interface.
2. Converts the load reductions to this new interface

### DIFF
--- a/jlm/llvm/ir/operators/Load.cpp
+++ b/jlm/llvm/ir/operators/Load.cpp
@@ -448,7 +448,7 @@ LoadStoreStateReduction::GetInstance() noexcept
   return loadStoreStateReduction;
 }
 
-LoadDuplicateStateReduction::~LoadDuplicateStateReduction() = default;
+LoadDuplicateStateReduction::~LoadDuplicateStateReduction() noexcept = default;
 
 LoadDuplicateStateReduction::LoadDuplicateStateReduction() noexcept = default;
 
--- a/jlm/llvm/ir/operators/Load.cpp
+++ b/jlm/llvm/ir/operators/Load.cpp
@@ -368,7 +368,7 @@ LoadStoreStateReduction::LoadStoreStateReduction() noexcept = default;
 bool
 LoadStoreStateReduction::IsReducibleState(
     const rvsdg::output * state,
-    const rvsdg::node * loadAlloca)
+    const rvsdg::Node * loadAlloca)
 {
   if (is<StoreNonVolatileOperation>(rvsdg::output::GetNode(*state)))
   {
--- a/jlm/llvm/ir/operators/Load.cpp
+++ b/jlm/llvm/ir/operators/Load.cpp
@@ -162,20 +162,14 @@ LoadVolatileNode::copy(rvsdg::Region * region, const std::vector<rvsdg::output *
   return &CreateNode(*region, GetOperation(), operands);
 }
 
-/* load normal form */
+LoadMuxReduction::~LoadMuxReduction() noexcept = default;
 
-/*
-  sx1 = MemStateMerge si1 ... siM
-  v sl1 = load_op a sx1
-  =>
-  v sl1 ... slM = load_op a si1 ... siM
-  sx1 = MemStateMerge sl1 ... slM
+LoadMuxReduction::LoadMuxReduction() noexcept = default;
 
-  FIXME: The reduction can be generalized: A load node can have multiple operands from different
-  merge nodes.
-*/
-static bool
-is_load_mux_reducible(const std::vector<rvsdg::output *> & operands)
+bool
+LoadMuxReduction::IsApplicable(
+    const LoadNonVolatileOperation &,
+    const std::vector<rvsdg::output *> & operands)
 {
   // Ignore loads that have no state edge.
   // This can happen when the compiler can statically show that the address of a load is NULL.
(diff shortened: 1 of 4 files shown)


phate:
The idea for going forward:

1. Convert all reductions utilized in the NodeReduction optimization to this new interface
2. Convert the NodeReduction optimization to a simple fix-point pass that applies all node normalizations until the RVSDG does not change any longer. 
3. Convert unit tests to utilize new interface instead of old interfaces
4. ...
5. Remove old normalization interface
6. Profit

haved (commented):
These changes make the code much more readable, and I am eagerly awaiting the change turning normalization into a pass.

I have two comments, one of which is a small suggestion, while the other might be a latent bug, if I am not mistaken.

jlm/llvm/ir/operators/Load.cpp
```diff
@@ -448,30 +441,50 @@ perform_load_store_state_reduction(
...
-  std::unordered_set<rvsdg::output *> seen_state;
+  std::vector<rvsdg::output *> newLoadStates;
+  std::unordered_set<rvsdg::output *> seenStates;
   std::vector<rvsdg::output *> results(operands.size(), nullptr);
   for (size_t n = 1; n < operands.size(); n++)
   {
     auto state = operands[n];
-    if (seen_state.find(state) != seen_state.end())
+    if (seenStates.find(state) != seenStates.end())
       results[n] = state;
```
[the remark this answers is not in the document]
phate: Yikes, great catch! I provided a fix in #678. Let's merge the fix first, then I will rebase this PR.

jlm/rvsdg/NodeNormalization.hpp
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * \brief Performs the normalization of the operands.
+   *
+   * @param operation The operation for which to normalize the operands.
+   * @param operands The operands corresponding to the operation.
+   *
+   * @return The normalized operands.
+   */
+  virtual std::vector<output *>
+  ApplyNormalization(const TOperation & operation, const std::vector<output *> & operands) = 0;
```
caleridas, line 68: Why not combine IsApplicable and ApplyNormalization into single method that returns std::optional<std::vector<output *>> ?

Looking at the single implementation that you wrote already it appears they are going to have a lot of code redundancies anyways.

jlm/rvsdg/NodeNormalization.hpp
```diff
@@ -0,0 +1,73 @@
...
+class output;
+
+/**
+ * \brief Normalizes the operands of an operation.
+ *
+ * The usage expectation is that IsApplicable() is invoked to determine whether operands can be
+ * normalized BEFORE ApplyNormalization() is invoked.
+ *
+ * @tparam TOperation The operation for which to normalize the operands.
+ */
```
caleridas, line 23: While generally I do not object to explicit interfaces, but given the suggestion below this is a somewhat verbose way of writing:

template<class TOperation> using NodeNormalization = std::function< std::optional<...>(const TOperation&, const std::vector<...>)>;

jlm/rvsdg/NodeNormalization.hpp
```diff
@@ -0,0 +1,73 @@
...
+{
+  static_assert(
+      std::is_base_of<Operation, TOperation>::value,
+      "Template parameter TOperation must be derived from jlm::rvsdg::operation.");
+
+public:
+  virtual ~NodeNormalization() noexcept = default;
+
+  NodeNormalization() = default;
+
```
caleridas, line 35: given that this is just an interface, why explicit about all of copy & assign? that I think belongs to implementation

caleridas: Can you maybe also explain in what places that interface would be used? I think having a base interface is only useful when you somehow abstract over the actual implementations (i.e. require the interface in a certain place without knowing the underlying implementation). In all places that it is wired up to in Load.cpp actually you never need an interface, free functions would do as well.

jlm/rvsdg/NodeNormalization.hpp
phate: I was considering this interface originally as well, but decided against it for now. It would have meant that I would not just have to decouple the normalizations from the normal form, but also extensively edit them to fit this new interface, potentially leading to bugs etc. As it is right now, it is more or less a copy-paste exercise to copy the code over from the old methods in the new interface. Once, the interface is up and we have thrown the normalization form, I am planning to go towards this interface. It would also be able to more easily go there, as I could convert each normalization individually at the time.

TLDR: It believed it to be simpler and safer to go this route first, before going to this other interface.

jlm/rvsdg/NodeNormalization.hpp
phate: I would like the normalizations to be able to carry state. Going forward, I would like to implement  the following normalizations:

1. A normalization that takes a vector of normalizations and applies them in sequence
2. A normalization that takes a single normalization (which can be a vector normalization) and applies it until a fixpoint is reached.
3. A normalization that takes a normalization and applies it n times. This can maybe combined with point 2, not sure yet.

Either way, I would like to have the option for them to carry state.

jlm/rvsdg/NodeNormalization.hpp
phate: I can push the copy & assign down to the leave classes.

Regarding the interface: See comment above.

jlm/llvm/ir/operators/Load.cpp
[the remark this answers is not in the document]
phate: Merged master into it.

jlm/llvm/ir/operators/Load.cpp
```diff
@@ -191,113 +185,107 @@ is_load_mux_reducible(const std::vector<rvsdg::output *> & operands)
...
+  auto ld = LoadNonVolatileNode::Create(
+      operands[0],
+      loadStates,
+      operation.GetLoadedType(),
+      operation.GetAlignment());
+
+  std::vector<rvsdg::output *> results(1, ld[0]);
+  results.insert(results.end(), std::next(ld.begin()), ld.end());
+  results.insert(results.end(), otherStates.begin(), otherStates.end());
+  return results;
```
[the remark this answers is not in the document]
phate: Fixed.

jlm/rvsdg/NodeNormalization.hpp
phate: I move copy & paste constructors/operators to leave classes

haved (approved):
Your responses to the comments make sense to me!

jlm/rvsdg/NodeNormalization.hpp
caleridas: well if you have a function returning `std::optional<...>` then it can serve both purposes -- call sites that want to check first simply call it twice, so you have an immediate transition path

I don't understand the "convert each normalization individually" bit -- since you did not define a common underlying interface for all of them, they are entirely decoupled anyways. The "base class" you defined is templated, so cannot be a common interface.

jlm/rvsdg/NodeNormalization.hpp
caleridas: but in the interface description you already explicitly rule state that they need to be reentrant -- which means they will not be able to carry state across multiple calls

neither fixpoint iteration nor chain of other normalizers requires state across calls

phate: What do you mean by "state across calls"? State that is instantiated in IsApplicable() and used in ApplyNormalization()? Or state between multiple invocations of IsApplicable()/ApplyNormalization()?

jlm/rvsdg/NodeNormalization.hpp
phate: My main goal is to get the old node normalization interface out. I did not want to refactor the individual normalizations on top of that.

phate:
Closed as the interface was insufficient. The PR that implements the proper interface is #683

PR #678, opened 2024-12-06
Fix duplicate state normalization for load nodes

phate:
This PR does the following:
1. Splits up all load tests into their own unit tests
2. Fixes the duplicate state normalization for load nodes


phate:
@haved

haved (approved):
Looks great, thank you for the ping as well. I apparently still have not gotten my notification settings correct

PR #679, opened 2024-12-08
Proof of Concept: Node Normalization Interface 2nd

phate:
I got a lot of push back in #677 from @caleridas regarding the interface. I tried his suggestions and put up a proof of concept. In retrospect, I have to say that some of the ideas would not have worked as I planned them in #677, and this turned out to be easier (and less code) than expected. It was a bit of a functional programming exercise in C++ though.

@caleridas @haved I would like to get some feedback before I continue and put up a proper PR for review.

### DIFF
--- a/jlm/llvm/ir/operators/Load.cpp
+++ b/jlm/llvm/ir/operators/Load.cpp
@@ -670,65 +670,71 @@ load_normal_form::normalized_create(
   return simple_normal_form::normalized_create(region, op, operands);
 }
 
-LoadNonVolatileNormalization LoadMuxNormalization =
-    [](const LoadNonVolatileOperation & operation,
-       const std::vector<rvsdg::output *> & operands) -> std::optional<std::vector<rvsdg::output *>>
+std::optional<std::vector<rvsdg::output *>>
+NormalizeLoadMux(
+    const LoadNonVolatileOperation & operation,
+    const std::vector<rvsdg::output *> & operands)
 {
   if (is_load_mux_reducible(operands))
     return perform_load_mux_reduction(operation, operands);
 
   return std::nullopt;
-};
+}
 
-LoadNonVolatileNormalization LoadStoreNormalization =
-    [](const LoadNonVolatileOperation & operation,
-       const std::vector<rvsdg::output *> & operands) -> std::optional<std::vector<rvsdg::output *>>
+std::optional<std::vector<rvsdg::output *>>
+NormalizeLoadStore(
+    const LoadNonVolatileOperation & operation,
+    const std::vector<rvsdg::output *> & operands)
 {
   if (is_load_store_reducible(operation, operands))
     return perform_load_store_reduction(operation, operands);
 
   return std::nullopt;
-};
+}
 
-LoadNonVolatileNormalization LoadAllocaNormalization =
-    [](const LoadNonVolatileOperation & operation,
-       const std::vector<rvsdg::output *> & operands) -> std::optional<std::vector<rvsdg::output *>>
+std::optional<std::vector<rvsdg::output *>>
+NormalizeLoadAlloca(
+    const LoadNonVolatileOperation & operation,
+    const std::vector<rvsdg::output *> & operands)
 {
   if (is_load_alloca_reducible(operands))
     return perform_load_alloca_reduction(operation, operands);
 
   return std::nullopt;
-};
+}
 
-LoadNonVolatileNormalization LoadStoreStateNormalization =
-    [](const LoadNonVolatileOperation & operation,
-       const std::vector<rvsdg::output *> & operands) -> std::optional<std::vector<rvsdg::output *>>
+std::optional<std::vector<rvsdg::output *>>
+NormalizeLoadStoreState(
+    const LoadNonVolatileOperation & operation,
+    const std::vector<rvsdg::output *> & operands)
 {
   if (is_load_store_state_reducible(operation, operands))
     return perform_load_store_state_reduction(operation, operands);
 
   return std::nullopt;
-};
+}
 
-LoadNonVolatileNormalization LoadDuplicateStateNormalization =
-    [](const LoadNonVolatileOperation & operation,
-       const std::vector<rvsdg::output *> & operands) -> std::optional<std::vector<rvsdg::output *>>
+std::optional<std::vector<rvsdg::output *>>
+NormalizeLoadDuplicateState(
+    const LoadNonVolatileOperation & operation,
+    const std::vector<rvsdg::output *> & operands)
 {
   if (is_multiple_origin_reducible(operands))
     return perform_multiple_origin_reduction(operation, operands);
 
   return std::nullopt;
-};
+}
 
-LoadNonVolatileNormalization LoadLoadStateNormalization =
-    [](const LoadNonVolatileOperation & operation,
-       const std::vector<rvsdg::output *> & operands) -> std::optional<std::vector<rvsdg::output *>>
+std::optional<std::vector<rvsdg::output *>>
+NormalizeLoadLoadState(
+    const LoadNonVolatileOperation & operation,
+    const std::vector<rvsdg::output *> & operands)
 {
   if (is_load_load_state_reducible(operands))
     return perform_load_load_state_reduction(operation, operands);
 
   return std::nullopt;
-};
+}
 
 }
 
--- a/jlm/llvm/ir/operators/Load.cpp
+++ b/jlm/llvm/ir/operators/Load.cpp
@@ -670,6 +670,66 @@ load_normal_form::normalized_create(
   return simple_normal_form::normalized_create(region, op, operands);
 }
 
+LoadNonVolatileNormalization LoadMuxNormalization =
+    [](const LoadNonVolatileOperation & operation,
+       const std::vector<rvsdg::output *> & operands) -> std::optional<std::vector<rvsdg::output *>>
+{
+  if (is_load_mux_reducible(operands))
+    return perform_load_mux_reduction(operation, operands);
+
+  return std::nullopt;
+};
+
+LoadNonVolatileNormalization LoadStoreNormalization =
+    [](const LoadNonVolatileOperation & operation,
+       const std::vector<rvsdg::output *> & operands) -> std::optional<std::vector<rvsdg::output *>>
+{
+  if (is_load_store_reducible(operation, operands))
+    return perform_load_store_reduction(operation, operands);
+
+  return std::nullopt;
+};
+
+LoadNonVolatileNormalization LoadAllocaNormalization =
+    [](const LoadNonVolatileOperation & operation,
+       const std::vector<rvsdg::output *> & operands) -> std::optional<std::vector<rvsdg::output *>>
+{
+  if (is_load_alloca_reducible(operands))
+    return perform_load_alloca_reduction(operation, operands);
+
+  return std::nullopt;
+};
+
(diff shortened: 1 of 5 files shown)


jlm/rvsdg/NodeNormalization.hpp
```diff
@@ -0,0 +1,77 @@
...
+template<class TOperation>
+using NodeNormalization = std::function<
+    std::optional<std::vector<output *>>(const TOperation &, const std::vector<output *> &)>;
+
+template<class TOperation>
+using NodeNormalizationSequenceType = std::function<NodeNormalization<TOperation>(
+    const std::vector<NodeNormalization<TOperation>> &)>;
+
+template<class TOperation>
+NodeNormalizationSequenceType<TOperation> NodeNormalizationSequence =
```
haved, line 31: Does `NodeNormalizationSequence` need to be a lambda, and not just a regular function? Will it ever be sent anywhere? It makes sense to use `std::function` for the normalizations themselves, but will this function ever be used in a general setting?

jlm/rvsdg/NodeNormalization.hpp
```diff
@@ -0,0 +1,77 @@
...
+    [](const std::vector<NodeNormalization<TOperation>> & nodeNormalizations)
+    -> NodeNormalization<TOperation>
+{
+  return [&](const TOperation & operation,
```
haved, line 35: The user of this function has to be really careful that the vector still exists when the returned lambda is called.

jlm/llvm/opt/reduction.cpp
```diff
@@ -111,11 +113,39 @@ enable_binary_reductions(rvsdg::Graph & graph)
...
   nf->set_reducible(true);
 }
+#endif
+
+static bool
+NormalizeLoadNode(LoadNonVolatileNode & node)
+{
+  static auto LoadNormalizationSequence =
+      rvsdg::NodeNormalizationSequence<LoadNonVolatileOperation>(
+          { LoadMuxNormalization, LoadAllocaNormalization });
```
haved, line 123: Does this vector live long enough to still be valid when the normalization is actually called?

I think you would have to place this vector as a static on the line above, and pass in a reference?

jlm/rvsdg/NodeNormalization.hpp
```diff
@@ -0,0 +1,77 @@
...
+using NodeReduction = std::function<bool(Node &)>;
+
+template<class TOperation>
+using NodeNormalizationReductionType = std::function<NodeReduction(NodeNormalization<TOperation>)>;
+
+template<class TOperation>
+NodeNormalizationReductionType<TOperation> NodeNormalizationReduction =
+    [](NodeNormalization<TOperation> nodeNormalization) -> NodeReduction
+{
+  return [&](Node & node) -> bool
```
haved, line 59: Here you capture `nodeNormalization` by reference, but that is a function parameter that will no longer be alive by the time this lambda is returned

haved (commented):
I think there are some issues with lambda captures by reference outliving the things they reference. As long as those issues are mitigated the design should work nicely, but using lambdas heavily can definitely make it a bit harder to keep track of lifetimes

jlm/llvm/opt/reduction.cpp
phate: You are right. The vector would need to be placed as state on the line above. Minor issue though.

jlm/rvsdg/NodeNormalization.hpp
phate: The idea for the normalization sequence is to be a normalization itself again. That is why the type takes a sequence of normalizations, which are applied in order until the first one that triggers, and returns a normalization again. This means I can push the normalization sequence in any place where a normal normalization is expected, e.g., NodeReduction.

jlm/rvsdg/NodeNormalization.hpp
phate: It might be better to use type `std::vector<NodeNormalization<TOperation>>` instead of `const std::vector<NodeNormalization<TOperation>>&` here. Then we would create a copy of the normalizations and everything should be fine.

phate: This would also help with your comment above, where the vector should be static at the line above.

jlm/rvsdg/NodeNormalization.hpp
phate: Hmmm yeah, I guess the lambda should be [nodeNormalization]`

jlm/rvsdg/NodeNormalization.hpp
phate: Alternatively, the lambda can also be [nodeNormalizations]`

jlm/rvsdg/NodeNormalization.hpp
haved: I agree that a sequence of normalizations can itself be seen as a normalization. The signature of the function makes this clear as well. The function I am talking about is the "outer" function, which takes a list of normalizations and produces a single normalization. This outer function is not a normalization itself, so it does not have to go via std::function or lambdas, unless there will be other functions with the signature `(vector<Normalization>) -> Normalization` that we went to use in generic settings.

jlm/rvsdg/NodeNormalization.hpp
haved: I would go with the last suggestion. Taking the vector by copy but still capturing it by reference in the lambda would not work.

caleridas (changes requested):
I think this simplifies, but I do not understand why defining top-level things as std::function objects instead of plain functions helpes anything, or why it is in any way helpful to do "indirection by vector" when statically chaining things.

For example, if you have three normalizing functions:

```
std::optional<std::vector<rvsdg::output*>> normal1(const Operation&, const std::vector<rvsdg::output*>&);

std::optional<std::vector<rvsdg::output*>> normal2(const Operation&, const std::vector<rvsdg::output*>&);

std::optional<std::vector<rvsdg::output*>> normal3(const Operation&, const std::vector<rvsdg::output*>&);
```

And now you want to chain these, then there is really no need to create a vector or similar, you can just do:

```
std::optional<std::vector<rvsdg::output*>> chain(const Operation& op, const std::vector<rvsdg::output*>& vals)
{
  std::vector<rvsdg::output *> current = vals;
  std::optional<std::vector<rvsdg::output *>> result;
  for (normal : {normal1, normal2, normal3}) {
    result = normal(op, vals);
    vals = result.value_or(std::move(vals));
  }
  return result;
}
```

where the above can certainly also be templated / generalized and without building an intermediate vector or anything.

Maybe normalizer interface could also be: `std::vector<rvsdg::output *> normalize(std::vector<rvsdg::output *> in)` -- notice the lack of `const &` for the parameter, so you are forced to "move original arguments in" and "get (possibly transformed) arguments out" -- so each normalizer must simply pass through the original arguments if it does not wish to modify them (and using move they are also not copied around). That way chaining simply becomes `normal3(normal2(normal1(outputs)))` which is kind of the most natural form anyways.

I am in general not too certain of the benefits of "hiding control flow" by wrapping things into structures _unless_ there is a need to make things runtime dynamic.

jlm/rvsdg/NodeNormalization.hpp
haved: True. I wish there was a nice way of declaring that a function has a signature that conforms to an interface in C++. It is not a big issue, however, as you would get compilation errors as soon as you try converting your function to an `std::function` through the typedef.

phate: Converted to plain functions

jlm/llvm/opt/reduction.cpp
phate: Should not be an issue any longer.

jlm/rvsdg/NodeNormalization.hpp
phate: Outdated.

jlm/llvm/ir/operators/Load.cpp
```diff
@@ -670,6 +670,66 @@ load_normal_form::normalized_create(
   return simple_normal_form::normalized_create(region, op, operands);
 }
 
+LoadNonVolatileNormalization LoadMuxNormalization =
```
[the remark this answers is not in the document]
phate: I used a plain function.

jlm/rvsdg/NodeNormalization.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+template<class TOperation>
+using NodeNormalizationSequenceType = std::function<NodeNormalization<TOperation>(
+    const std::vector<NodeNormalization<TOperation>> &)>;
+
+template<class TOperation>
+NodeNormalizationSequenceType<TOperation> NodeNormalizationSequence =
+    [](const std::vector<NodeNormalization<TOperation>> & nodeNormalizations)
+    -> NodeNormalization<TOperation>
+{
```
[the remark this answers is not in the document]
phate: Changed it to a simple function.

phate:
@haved @caleridas I updated the interface. Please have a look again.

> Maybe normalizer interface could also be: std::vector<rvsdg::output *> normalize(std::vector<rvsdg::output *> in) -- notice the lack of const & for the parameter, so you are forced to "move original arguments in" and "get (possibly transformed) arguments out" -- so each normalizer must simply pass through the original arguments if it does not wish to modify them (and using move they are also not copied around). That way chaining simply becomes normal3(normal2(normal1(outputs))) which is kind of the most natural form anyways.

The normalization interface currently does not take node arguments in and returns node arguments. It currently takes node arguments in and returns the results of a node. That is also why the normalization sequence runs through the individual normalizations until the first one can be applied, and then stops. I see the appeal of your approach (an was thinking like that before as well), but realized that some of the current normalization do not fit this pattern, or do not fit it nicely.

@caleridas @haved Could you close conversations that you think are resolved, so we can keep an overview of what still needs to be done.

jlm/llvm/opt/reduction.cpp
```diff
@@ -111,11 +113,43 @@ enable_binary_reductions(rvsdg::Graph & graph)
...
+NormalizeLoadNode(LoadNonVolatileNode & node)
+{
+  static std::vector<rvsdg::NodeNormalization<LoadNonVolatileOperation>> loadNormalizations(
+      { NormalizeLoadMux, NormalizeLoadAlloca });
+
+  auto normalizeLoadNode = std::bind(
+      rvsdg::NormalizeSequence<LoadNonVolatileOperation>,
+      loadNormalizations,
+      std::placeholders::_1,
+      std::placeholders::_2);
```
haved, line 128: I did not know that `std::placeholders::*` was a thing, but they are kind of cool.

One thing to note is that this call to `bind` takes `loadNormalizations` by value, which means it is copied when constructing `normalizeLoadNode`.

In order to make the bind object not copy our vector, we can use `std::cref(loadNormalizations)`.

haved: Reading the documentation of `std::bind` was difficult, so I tested in godbolt to confirm: https://godbolt.org/z/rv4K37v3x

haved: Another alternative is of course just making a lambda. `std::bind` is a bit more concise, but less well known

haved (approved):
This design is very readable, while also making it nice to opt-in to dynamic dispatch. I have one comment about `std::bind`.

phate:
@caleridas Any opinion?

haved:
@phate What is the state of this PR now? Should it be closed now that #683 is closed?

PR #680, opened 2024-12-11
Make small fixes to scripts + update commit hashes of CIRCT and hls-test-suite

haved:
The updated hashes give us a smaller CIRCT build, and cleaner output from the hls-test-suite, plus disables the problem test `test_memory_5`.

The other fixes include:
 - avoid crashing if `lit` is not found immediately (it can be supplied using the option instead)
 - avoid looping indefinitely when seeing unknown options
 - make the scripts a little more consistent

### DIFF
--- a/scripts/build-mlir.sh
+++ b/scripts/build-mlir.sh
@@ -52,7 +52,7 @@ while [[ "$#" -ge 1 ]] ; do
 			commit >&1
 			exit 0
 			;;
-		--help)
+		--help|*)
 			usage >&2
 			exit 1
 			;;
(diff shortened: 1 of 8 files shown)


scripts/build-mlir.sh
```diff
@@ -52,7 +52,7 @@ while [[ "$#" -ge 1 ]] ; do
 			commit >&1
 			exit 0
 			;;
-		--help)
+		--help|*)
```
phate, line 55: Out of curiosity: What does this do?

phate (approved):

scripts/build-mlir.sh
phate: Is this the "avoid looping indefinitely"?

haved: @phate yes, this switch statement is inside a while loop that keeps running until all arguments have been shifted. If nothing matches, nothing is shifted, and it loops forever

PR #682, opened 2024-12-13
Adds additional cases for remove_redundant_buf optimization

sjalander:


jlm/hls/backend/rvsdg2rhls/remove-redundant-buf.cpp
```diff
@@ -20,6 +20,11 @@ eliminate_buf(jlm::rvsdg::output * o)
     {
       return eliminate_buf(node->input(1)->origin());
     }
+    else if (dynamic_cast<const jlm::hls::fork_op *>(&node->GetOperation()))
```
phate, line 23: `is<>()`

jlm/hls/backend/rvsdg2rhls/remove-redundant-buf.cpp
```diff
@@ -28,7 +33,16 @@ eliminate_buf(jlm::rvsdg::output * o)
     {
       return true;
     }
+    else if (dynamic_cast<const jlm::hls::load_op *>(&node->GetOperation()))
```
phate, line 36: `is<>()`

jlm/hls/backend/rvsdg2rhls/remove-redundant-buf.cpp
```diff
@@ -28,7 +33,16 @@ eliminate_buf(jlm::rvsdg::output * o)
...
+    {
+      return true;
+    }
+    else if (dynamic_cast<const jlm::hls::store_op *>(&node->GetOperation()))
```
phate, line 40: `is<>()`

phate (changes requested):
Again, could we get some unit tests on this?

jlm/hls/backend/rvsdg2rhls/remove-redundant-buf.cpp
sjalander: Changed

jlm/hls/backend/rvsdg2rhls/remove-redundant-buf.cpp
sjalander: Changed

phate (approved):
The unit tests were merged in another PR, right?

sjalander:
A unit test has not been created for this PR. The reason is that when I started looking into the functionality of this optimization, I'm not able to determine its purpose. The name seems to be somewhat misleading. From what I can tell, it replaces pass through buffers with non-pass through buffers in some specific cases. However, it doesn't seem to have an effect on the hls_test_suite as when I comment out the optimization, then I get no change in the number of cycles.

@davidmetz Do you have a use case where this optimization makes a difference?

PR #683, opened 2024-12-15
Add new normalization interface and convert load reductions to it

phate:
This PR is the first in a series of PRs to remove the old node normalization interface that is intertwined in the RVSDG library. This PR does the following:

1. Adds a new and simple node normalization interface for optimizing the operands of a node. This enables to create individual instances of normalizations and therefore pass them around, and/or configure optimization by simply providing these instances instead of utilizing the flags of the old interface.
2. Converts the load reductions to this new interface
3. Utilizes these new reductions in load tests

### DIFF
--- a/jlm/llvm/ir/operators/Load.hpp
+++ b/jlm/llvm/ir/operators/Load.hpp
@@ -12,6 +12,8 @@
 #include <jlm/rvsdg/simple-node.hpp>
 #include <jlm/rvsdg/simple-normal-form.hpp>
 
+#include <optional>
+
 namespace jlm::llvm
 {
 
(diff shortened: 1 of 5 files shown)


jlm/llvm/ir/operators/Load.hpp
```diff
@@ -575,6 +577,134 @@ class LoadNonVolatileNode final : public LoadNode
...
+ * ... = load_op a1 ss1
+ *
+ * @param operation The load operation on which the transformation is performed.
+ * @param operands The operands of the load node.
+ *
+ * @return If the normalization could be applied, then the results of the load operation after
+ * the transformation. Otherwise, std::nullopt.
+ */
+std::optional<std::vector<rvsdg::output *>>
+NormalizeLoadStoreState(
```
haved, line 664: Is the result of this normalization always applicable for `NormalizeLoadStore`?
If so, I guess we could go all the way to doing load elision in this normalization, but there might be cases im not seeing where that is not possible?

jlm/llvm/ir/operators/Load.hpp
```diff
@@ -575,6 +577,134 @@ class LoadNonVolatileNode final : public LoadNode
...
+ * _ so3 = load_op _ si1
+ *
+ * @param operation The load operation on which the transformation is performed.
+ * @param operands The operands of the load node.
+ *
+ * @return If the normalization could be applied, then the results of the load operation after
+ * the transformation. Otherwise, std::nullopt.
+ */
+std::optional<std::vector<rvsdg::output *>>
+NormalizeLoadLoadState(
```
haved, line 704: Does this normalization create a memory state merge node? A later store that depends on `so3` needs to be serialized after all three loads, so it can no longer just depend on `so3` in the new graph.

haved (approved):
Very nice, I just have one question about the documentation string for one of the normalizations.

Once memory state encoding has been done, the two normalizations that operate on addresses that come directly from alloca nodes  will likely have less to do, but they might still be useful as quick cleanups when the graph changes

jlm/llvm/ir/operators/Load.hpp
phate: I have not seen this, but I guess you are right. However, I would not like to also refactor the normalizations now. I want to get the old normalization interface out. That is my highest priority for now. Looking at the individual normalization comes afterwards. It is a very good point that you raise though.

jlm/llvm/ir/operators/Load.hpp
phate: Yes, it creates one: https://github.com/phate/jlm/blob/5c82eb176c6892c9a0528cb0593decd1e8e8be5d/jlm/llvm/ir/operators/Load.cpp#L554C1-L564C4

haved (approved):
All my comments are resolved, @phate

PR #685, opened 2024-12-16
Add clang-tidy support to CI

### LINKED ISSUE #313: Incorporate and enforce clang-tidy to improve code quality
We catch ourselves commenting on code smells that could be automatically detected with clang-tidy. We should enable clang-tidy and enforce it in the CI to not waste our time in reviews and improve code quality. This could be an incremental approach, where we start out with some simple rules and then gradually add new rules (while fixing the problems in code) in order to avoid a huge PR that touches everything and leads to conflicts for everyone.

phate:
This PR adds support for clang-tidy to the CI. It does not enforce any checks yet, but only adds the necessary infrastructure and fixes the problems in the code such that clang-tidy finishes without errors. Relevant checks will be added in future PRs.

Close #313

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,8 @@
+---
+Checks: '-*,
+  modernize-deprecated-headers,
+  '
+
+WarningsAsErrors: '
+  modernize-deprecated-headers,
+  '
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -4,5 +4,5 @@ Checks: '-*,
   '
 
 WarningsAsErrors: '
-  modernize-deprecated-headers,
+#  modernize-deprecated-headers,
   '
--- a/.github/actions/InstallPackages/action.yml
+++ b/.github/actions/InstallPackages/action.yml
@@ -27,6 +27,11 @@ inputs:
     default: "false"
     required: false
 
+  install-clang-tidy:
+    description: "Install clang-tidy package. Default is 'false'."
+    default: "false"
+    required: false
+
   install-ninja:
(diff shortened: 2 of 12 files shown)


.clang-tidy
```diff
@@ -0,0 +1,8 @@
+---
+Checks: '-*,
+  modernize-deprecated-headers,
+  '
+
+WarningsAsErrors: '
+#  modernize-deprecated-headers,
```
phate, line 7: I needed at least one check for clang-tidy to work. The problems found by this check are currently considered only warnings and not errors. The # leaves it commented out in the 'WarningAsErrors' section.

haved:
@phate looks very nice, though I do not quite understand the `.clang-tidy` file itself. Are we only using it to do one check?

Since the WarningAsError only contains one check, that is also commented out, what will actually cause the GitHub Action to fail?

phate:
> @phate looks very nice, though I do not quite understand the .clang-tidy file itself. Are we only using it to do one check?

Since the WarningAsError only contains one check, that is also commented out, what will actually cause the GitHub Action to fail?

EDIT: Sorry, I read the PR description, this PR is not supposed to add checks

@haved The single check in .clang-tidy is required such that clang-tidy does not complain and actually goes to work. The single check produces some warnings, but I did not want them to be fixed in this PR to aovid to clutter this one too much. This PR is only about the infrastructure and fixing the initial errors such that clang-tidy comes back to us error free. We can enable checks in other PRs.

haved (approved):

PR #686, opened 2024-12-18
Fix duplicate state normalization for store nodes

phate:
This PR does the following:
1. Splits up all store tests into their own unit tests
2. Fixes the duplicate state normalization for store nodes


haved (approved):
Very nice, I had not considered that stores also have the same multi-state reduction

PR #687, opened 2024-12-18
Add modernize-deprecated-headers check to clang-tidy

phate:


haved (approved):
Very nice, I also agree with the way it orders the includes, assuming it did those changes as well

PR #690, opened 2024-12-19
Covert mux normalizations to new normalization interface

phate:


phate:
@haved You are absolutely right. I should have checked before I invested the time in the PR. My bad! I opened a PR that removes the mux_op.

PR #691, opened 2024-12-19
Remove unused state multiplexer operation

phate:


haved (dismissed):
Very nice!
I hope the hls-test-suite will work on its second try, though

haved:
@phate looks like this PR caused one of the hls tests to generate worse RTL. I re-ran the test, but it seems to be consistent

phate:
@haved This does not make any sense. The mux operation was nowhere used. It must be an ordering problem.

@sjalander Could we be a little bit more lenient about this until we fix the non-deterministic behavior in the traversal? I do not fully understand why the cycle count changes just because we have a different node ordering.

haved:
@phate the traverser places pointers to heap allocations in sets and iterates over them. This means any change that affects the number or size of heap allocations potentially affects traversal order

phate:
@haved I fully understand how a different heap order results in a different node traversal. What I fail to understand is how a different node order results in a different number of cycles.

sjalander:
My guess would be that buffers are inserted in different places depending on the order nodes are traversed. These don't necessarily affect the logical function but are inserted to decouple different dependency paths.

I've made a PR for the HLS test suite that disables the error if the cycles doesn't match.

haved (approved):

PR #692, opened 2024-12-20
Add unit test for HLS memory queue

sjalander:


haved (approved):
Nice use of `AssertedCast`. Regarding the tests, I guess the final assert sections could be a bit more strict than only requiring that certain operations existing.

It would be very nice to add documentation to `mem_queue` itself to make it easier to understand what the reader should expect it to do.

jlm/hls/backend/rvsdg2rhls/mem-queue.cpp
```diff
@@ -287,10 +284,9 @@ separate_load_edge(
         else
         {
           // end of loop
-          auto load_user_input = dynamic_cast<jlm::rvsdg::simple_input *>(addr_edge_user);
-          JLM_ASSERT(load_user_input);
-          JLM_ASSERT(
-              dynamic_cast<const jlm::hls::branch_op *>(&load_user_input->node()->GetOperation()));
+          auto load_user_input = jlm::util::AssertedCast<jlm::rvsdg::simple_input>(addr_edge_user);
+          jlm::util::AssertedCast<const jlm::hls::branch_op>(
```
phate, line 288: This here should have stayed a JLM_ASSERT() as the result of the AssertedCast is not used.

jlm/hls/backend/rvsdg2rhls/mem-queue.cpp
```diff
@@ -460,8 +456,7 @@ process_loops(jlm::rvsdg::output * state_edge)
     }
     else if (auto sti = dynamic_cast<jlm::rvsdg::StructuralInput *>(user))
     {
-      auto ln = dynamic_cast<jlm::hls::loop_node *>(sti->node());
-      JLM_ASSERT(ln);
+      jlm::util::AssertedCast<jlm::hls::loop_node>(sti->node());
```
phate, line 459: same here

jlm/hls/backend/rvsdg2rhls/mem-queue.cpp
```diff
@@ -287,10 +284,9 @@ separate_load_edge(
         else
         {
           // end of loop
-          auto load_user_input = dynamic_cast<jlm::rvsdg::simple_input *>(addr_edge_user);
-          JLM_ASSERT(load_user_input);
-          JLM_ASSERT(
-              dynamic_cast<const jlm::hls::branch_op *>(&load_user_input->node()->GetOperation()));
+          auto load_user_input = jlm::util::AssertedCast<jlm::rvsdg::simple_input>(addr_edge_user);
```
phate, line 287: load_user_input is not used either. They should all be JLM_ASSERTS. This is all just sanity checking.

jlm/hls/backend/rvsdg2rhls/mem-queue.cpp
```diff
@@ -272,10 +270,9 @@ separate_load_edge(
           }
           // create mux
           JLM_ASSERT(mem_edge->nusers() == 1);
-          auto mux_user = dynamic_cast<jlm::rvsdg::simple_input *>(*mem_edge->begin());
-          JLM_ASSERT(mux_user);
-          auto mux_op = dynamic_cast<const jlm::hls::mux_op *>(&mux_user->node()->GetOperation());
-          JLM_ASSERT(mux_op);
+          auto mux_user = jlm::util::AssertedCast<jlm::rvsdg::simple_input>(*mem_edge->begin());
+          auto mux_op =
```
phate, line 274: I do not see mux_op used either.

jlm/hls/backend/rvsdg2rhls/mem-queue.cpp
```diff
@@ -222,7 +220,7 @@ separate_load_edge(
       JLM_ASSERT(sti_arg->nusers() == 1);
       auto user = *sti_arg->begin();
       auto si = dynamic_cast<jlm::rvsdg::simple_input *>(user);
-      JLM_ASSERT(dynamic_cast<const jlm::hls::mux_op *>(&si->node()->GetOperation()));
+      jlm::util::AssertedCast<const jlm::hls::mux_op>(&si->node()->GetOperation());
```
phate, line 223: Same here.

phate (dismissed):
I agree with @haved comments. Some documentation and being a bit more strict in the tests would be nice.

jlm/hls/backend/rvsdg2rhls/mem-queue.cpp
sjalander: Reverted

jlm/hls/backend/rvsdg2rhls/mem-queue.cpp
sjalander: Reverted

sjalander: The dynamic cast is needed for the load_user_input as it is used for the second sanity check. I assume that an AssertedCast becomes the equivalent of a dynamic cast when asserts are disabled.
Ideally, the dynamic cast should also be disabled in a release build.

jlm/hls/backend/rvsdg2rhls/mem-queue.cpp
sjalander: Used on line 279

jlm/hls/backend/rvsdg2rhls/mem-queue.cpp
sjalander: Reverted

jlm/hls/backend/rvsdg2rhls/mem-queue.cpp
phate: The AssertedCast becomes a static cast when asserts are disabled. Otherwise, it is a dynamic cast.

jlm/hls/backend/rvsdg2rhls/mem-queue.cpp
```diff
@@ -460,8 +456,7 @@ process_loops(jlm::rvsdg::output * state_edge)
     }
     else if (auto sti = dynamic_cast<jlm::rvsdg::StructuralInput *>(user))
     {
-      auto ln = dynamic_cast<jlm::hls::loop_node *>(sti->node());
-      JLM_ASSERT(ln);
+      JLM_ASSERT(dynamic_cast<jlm::hls::loop_node *>(sti->node()));
```
phate, line 459: JLM_ASSERT(is<>())

phate (approved):

jlm/hls/backend/rvsdg2rhls/mem-queue.cpp
sjalander: Changed to is<>.

phate (approved):

PR #694, opened 2024-12-21
Add modernize-redundant-void-arg check to clang-tidy

phate:


haved (approved):
Very nice, I had seen some of those around the codebase, but it didn't feel right to "correct" them here and there. Much better to have clang-tidy do it

PR #696, opened 2024-12-23
Hls harness rework

haved:
Creates a new cleaned up harness for calling Verilated hls kernels from C code.

It defines a class for the memory queues, which simplifies the usage code a great deal.

I eventually "cracked the code" regarding writing harnesses for hardware:
Only read model outputs right before posedge, only write inputs right after posedge (and eval again).

Example diff between the old and new `test_load.hls.harness.cpp` can be seen [here](https://www.diffchecker.com/1vnmcZ6H/). The diff is not particularly easy to read, as I have re-arranged functions. I would recommend just reading the new harness, as it includes comments.

The new harness passes the tests, but some of the cycle counts have increased by one, due to the old harness allowing the model to "false start" when some computation does not depend on input. I have created a commit in `hls-test-suite` with the updated numbers.

I also cleaned up some other code to use range-based for loops over `.Arguments()` and `.Results()` and similar, instead of doing C-style for loops.

This PR is created partly to see how it fares in CI. It is definitely possible to write some tests for this code, particularly covering how the lambda's arguments, memory queues, and context variables are handled.


jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
```diff
@@ -1,743 +1,645 @@
...
+}
+
+/**
+ * Takes an HLS kernel and determines the return type of the original C function.
+ * Returns a tuple where the first element is a boolean signifying if anything is returned.
+ * The second element is a string containing the kernel's return type when called from C.
+ * @param hls_lambda the lambda node representing the kernel
+ * @return a tuple (has return value, return type)
+ */
+std::pair<bool, std::string>
```
phate, line 44: Could this not be changed to a `std::optional<>`?

jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
```diff
@@ -1,743 +1,645 @@
...
+  for (const auto & ctx : hls_lambda.GetContextVars())
   {
-    cpp << "    return 0;\n";
+    // Context variables should always be external symbols imported by name
+    if (const auto import = dynamic_cast<rvsdg::GraphImport *>(ctx.input->origin()))
+      cpp << "    top->i_data_" << first_ctx_var << " = (uint64_t) &" << import->Name() << ";"
+          << std::endl;
+    else
+      JLM_UNREACHABLE("Context variable is not a GraphImport");
+    first_ctx_var++;
```
phate, line 383: Minor nitpick and very subjetive: Invert the condtion and early check for the unreachable case, and then just put the main line code after the conditional. It would avoid the else case and the interwoven unreachable + `first_ctx_var++`

phate (approved):
LGTM. Just some absolutely minor comments.

jlm/hls/backend/rhls2firrtl/base-hls.hpp
```diff
@@ -72,63 +74,75 @@ class BaseHLS
...
 
+  /**
+   * Extracts all region results of the given kernel that represent memory requests.
+   * They can take multiple values within a single execution of the region.
+   * @param lambda the lambda node holding the hls kernel
+   * @return the results that represent memory requests
+   */
   std::vector<rvsdg::RegionResult *>
-  get_mem_reqs(const llvm::lambda::node * lambda)
+  get_mem_reqs(const llvm::lambda::node & lambda)
```
sjalander, line 102: GetMemoryRequestResults()

sjalander: If we use the term kernel for the HLS lambda then it could be named
GetKernelMemoryRequest()

jlm/hls/backend/rhls2firrtl/base-hls.hpp
```diff
@@ -72,63 +74,75 @@ class BaseHLS
...
-  std::vector<jlm::rvsdg::RegionArgument *>
-  get_mem_resps(const llvm::lambda::node * lambda)
+  /**
+   * Extracts all region arguments of the given kernel that represent memory responses.
+   * They can provide multiple values within a single execution of the region.
+   * @param lambda the lambda node holding the hls kernel
+   * @return the arguments that represent memory responses
+   */
+  std::vector<rvsdg::RegionArgument *>
+  get_mem_resps(const llvm::lambda::node & lambda)
```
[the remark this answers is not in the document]
sjalander: If we use the term kernel for the HLS lambda then it could be named
GetKernelMemoryResponses()

Could take the naming convention one step further:
GetKernelMemoryResponses(const llvm::lambda::node & kernel)
GetKernelMemoryResponses(const llvm::lambda::node & hlsKernel)

jlm/hls/backend/rhls2firrtl/base-hls.hpp
```diff
@@ -72,63 +74,75 @@ class BaseHLS
...
 
+  /**
+   * Extracts all region results from the given kernel that represent results from execution,
+   * as opposed to results used for making memory requests.
+   * @param lambda the lambda node holding the hls kernel
+   * @return the results of the lambda that represent the kernel outputs
+   */
   std::vector<rvsdg::RegionResult *>
-  get_reg_results(const llvm::lambda::node * lambda)
+  get_reg_results(const llvm::lambda::node & lambda)
```
sjalander, line 139: GetKernelResults()

jlm/hls/backend/rhls2firrtl/base-hls.hpp
```diff
@@ -72,63 +74,75 @@ class BaseHLS
...
-  get_reg_args(const llvm::lambda::node * lambda)
+  /**
+   * Extracts all region arguments of the given kernel that represent kernel inputs,
+   * which may include kernel arguments, state types, and context variables (always in that order).
+   * It will not return any arguments that represent memory responses.
+   * @param lambda the lambda node holding the hls kernel
+   * @return the arguments of the lambda that represent kernel inputs
+   */
+  std::vector<rvsdg::RegionArgument *>
+  get_reg_args(const llvm::lambda::node & lambda)
```
sjalander, line 121: These are harder to name, as it's basically get everything except for memory responses.
get_reg_args() sounds like one can get the arguments for a region, but this is only true for lambda.
Potential alternatives:

GetKernelArguments()  
# We should decide if we want to refer to the HLS function as the kernel, and in that case use it consistently.

GetHLSArguments()

jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
```diff
@@ -1,743 +1,645 @@
...
+
 namespace jlm::hls
 {
 
+// The number of cycles before a load is ready
+static constexpr int MEMORY_RESPONSE_LATENCY = 10;
+
 std::string
-VerilatorHarnessHLS::get_text(llvm::RvsdgModule & rm)
+convert_to_c_type(const rvsdg::Type * type)
```
sjalander, line 19: New functions should follow the new naming convention.
ConvertToCType()

jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
```diff
@@ -1,743 +1,645 @@
...
+
+/**
+ * Takes an HLS kernel and determines the return type of the original C function.
+ * Returns a tuple where the first element is a boolean signifying if anything is returned.
+ * The second element is a string containing the kernel's return type when called from C.
+ * @param hls_lambda the lambda node representing the kernel
+ * @return a tuple (has return value, return type)
+ */
+std::pair<bool, std::string>
+convert_to_c_return_type(const llvm::lambda::node & hls_lambda)
```
sjalander, line 45: Naming convention of function and argument.
hlsLambda or should we simply start calling it the “kernel”?

jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
```diff
@@ -1,743 +1,645 @@
...
+/**
+ * Takes an HLS kernel and determines the parameters of the original C function.
+ * Returns a tuple, the first element of which is the number of parameters.
+ * The second element is a string defining the C parameters, like "int32_t a0, void* a1, void* a2".
+ * The third element is a string for calling the C function, like "a0, a1, a2".
+ * @param hls_lambda the lambda node representing the kernel
+ * @return a tuple (number of parameters, string of parameters, string of call arguments)
+ */
+std::tuple<size_t, std::string, std::string>
+convert_to_c_parameter_list(const llvm::lambda::node & hls_lambda)
```
sjalander, line 69: Same as earlier

jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
```diff
@@ -1,743 +1,645 @@
...
-         "{\n";
-  // imports
-  auto root = rm.Rvsdg().root();
-  for (size_t i = 0; i < root->narguments(); ++i)
+
+  return std::make_tuple(argument_index, parameters.str(), arguments.str());
+}
+
+std::string
+VerilatorHarnessHLS::get_text(llvm::RvsdgModule & rm)
```
sjalander, line 97: GetText(llvm::RvsdgModule rvsdgModule)

jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
```diff
@@ -1,743 +1,645 @@
...
+{
+  std::ostringstream cpp;
+  const auto & hls_lambda = *get_hls_lambda(rm);
+  const auto & function_name = hls_lambda.name();
```
sjalander, line 101: Variable names

jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
```diff
@@ -1,743 +1,645 @@
...
+        return addr == other.addr && write == other.write && width == other.width && data == other.data;
     }
-    else if (dynamic_cast<const bundletype *>(&ln->type().ArgumentType(i)))
-    {
-      continue;
+};
+
+// A log of memory accesses made by the kernel
+std::vector<mem_access> memory_accesses;
+// Acesses to regions in this vector of (start, length) pairs are not traced
```
sjalander, line 178: Accesses

jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
```diff
@@ -1,743 +1,645 @@
...
+    Verilated::debug(0);
+
+    // Randomization reset policy
+    // May be overridden by commandArgs
+    Verilated::randReset(2);
+
+    // Verilator must compute traced signals
+    Verilated::traceEverOn(true);
+
+    // Pass arguments so Verilated code can see them, e.g. $value$plusargs
```
sjalander, line 353: , e.g.,

sjalander (dismissed):
Would it make sense to switch to the new naming convention with camelback (or what the specific one used now is called) instead of separating names with '_' for code that's being updated?

jlm/hls/backend/rhls2firrtl/base-hls.hpp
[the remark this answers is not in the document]
haved: We could absolutely change to something closer to the established naming scheme. I was also thinking about the fact that there is no reason for these functions to be protected methods on BaseHLS. They do not use `this` at all

jlm/hls/backend/rhls2firrtl/base-hls.hpp
haved: I kind of want to split this into separate functions for regular arguments, state arguments and context variables, to avoid code that implicitly relies on the ordering being consistent. It is probably a safe assumption, but it makes code harder to read. I was thinking a second PR could fix that

haved:
@sjalander I suggest a second PR that renames those helper functions for getting arguments and results, and also splits them up and changes them from being protected methods to being regular functions

sjalander:
@haved 
That's fine with me.

jlm/hls/backend/rhls2firrtl/base-hls.hpp
```diff
@@ -22,6 +22,8 @@ isForbiddenChar(char c);
 class BaseHLS
 {
 public:
+  virtual ~BaseHLS() = default;
```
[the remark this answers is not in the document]
haved: I did not know this affected the vtables that way. Will fix

jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
haved: Changed

jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
haved: Changed

jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
haved: Changed

jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
haved: Changed

jlm/hls/backend/rhls2firrtl/base-hls.hpp
haved: Left for later PR

jlm/hls/backend/rhls2firrtl/base-hls.hpp
haved: Left for later PR

jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
haved: Changed to AssertedCast instead

jlm/hls/backend/rhls2firrtl/verilator-harness-hls.cpp
haved: Done

phate:
@haved Second PR is fine by me.

phate (approved):

PR #697, opened 2024-12-24
Add misc-unused-parameters check to clang-tidy

phate:


haved (approved):
I guess it takes some getting used to, but it should be a net benefit. For important functions, the header will still contain a declaration where all parameters are named

PR #699, opened 2024-12-25
Update .gitignore

urays:
build-* will tell git to ignore folders and filenames prefixed with build- when committing. However, this would cause the changes to scripts/build-circt.sh and build-mlir.sh to also be ignored.

So, it seems like build-*/ is working as expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,5 @@ Makefile.custom
 .idea
 docs/html
 build
-build-*
+build-*/
 usr


sjalander:
I just now tried making a change to scripts/build-circt.sh and without modifying .gitignore the change is seen when doing `git status` and the change can be staged with `git add scripts/build-circt.sh`. So, I don't understand what this PR changes.

urays:
> I just now tried making a change to scripts/build-circt.sh and without modifying .gitignore the change is seen when doing `git status` and the change can be staged with `git add scripts/build-circt.sh`. So, I don't understand what this PR changes.

Yes, making modifications to the submitted files will cause Git to consider them as changes that need to be committed. 

In fact, the issue I encountered with the original .gitignore was that I created a new folder  (containing new build/test scripts with the build- prefix) on my branch, but when I tried to commit them from scratch, the commit failed. 
All the other files in the new folder were successfully committed, except for the build-* files.

For this, I checked the official Git documentation and found the following sentence on [https://git-scm.com/docs/gitignore](https://git-scm.com/docs/gitignore): 'If there is a separator at the end of the pattern then the pattern will only match directories, otherwise the pattern can match both files and directories.'

So maybe .gitignore could be better?

phate (approved):

PR #700, opened 2024-12-26
Convert gamma reductions to new reduction interface

phate:


phate:
@haved  These were the last normalizations/reductions used by the node reduction pass. After this PR is merged, I will try to refactor the node reduction pass.

haved (approved):
Very nice, great comments

PR #701, opened 2024-12-28
Clean-up node reduction transformation

phate:
This PR cleans up the node reduction transformation. It does the following:

1. Brings the entire transformation to the current naming standards
2. Brings the entire transformation to the current coding standards
3. Utilizes the new normalization interface for the reductions


phate:
@haved This is a preview. I do not have unit tests yet.

jlm/llvm/opt/reduction.cpp
```diff
@@ -4,132 +4,220 @@
...
+    return ReduceStoreNode(simpleNode);
+  }
+  else if (is<rvsdg::unary_op>(&simpleNode))
+  {
+    // FIXME: handle the unary node
+    // See github issue #304
+  }
+  else if (is<rvsdg::binary_op>(&simpleNode))
+  {
+    ReduceBinaryNode(simpleNode);
```
haved, line 157: Missing a `return` here?

haved (commented):
Looks very nice so far,  I think there is a missing `return`, see comment

jlm/llvm/opt/reduction.cpp
phate: Great catch! Fixed.

phate:
@haved It is in my opinion ready for a proper review.

jlm/llvm/opt/reduction.cpp
```diff
@@ -4,132 +4,233 @@
...
+  {
+    // We can not go through the subregions as the structural node might already have been removed.
+    return true;
+  }
+
+  // Reduce all nodes in the subregions
+  for (size_t n = 0; n < structuralNode.nsubregions(); n++)
+  {
+    const auto subregion = structuralNode.subregion(n);
+    ReduceNodesInRegion(*subregion);
```
haved, line 134: Should the result of this function call be logical OR-ed into `reductionPerformed`?

haved (commented):
Looks very nice, but I think a return value might be unintentionally ignored. See comment, @phate.

Also, Happy New Year!

haved:
I guess we could even go as far as marking all those reduce functions as `[[nodiscard]]`, and only let the top level `run` function be `void`. Your call!

jlm/llvm/opt/reduction.cpp
phate: No, it should not. This would be a catastrophe in terms of performance. The reductionPerformed flag only counts for the node itself, but not for any node in the subregions of the structural node. We only want to repeat the iteration throughout the nodes in a region iff a node in THIS region could be reduced, but not in any of the subregions of a structural node from this region. If you would also do it for the subregions, then this flag would bubble up the region tree. In other words, if you reduce a node in the inner most subregion, then all its parents region would need to perform one more iteration as well, which would result in quadratic number of visits of nodes in the graph.

I added some documentation to this method to clarify that it only accounts for the node itself, and not for any node in the subregions.

phate:
> I guess we could even go as far as marking all those reduce functions as [[nodiscard]], and only let the top level run function be void. Your call!

Very good point. I marked them all as nodiscard

haved (approved):
Ah, I see now how the fixed-point iteration works. Your solution is of course much more performent!

PR #702, opened 2024-12-30
Clean up graph class

phate:


haved (approved):
Looks great. 

Just so we are on the same page, the methods that are now called `GetNodeNormalForm` and `MarkDenormalized` are eventually going to be removed entirely, once normalization is done via a pass? @phate

phate:
@haved Yes

PR #703, opened 2025-01-01
Add andersen top down lifetime aware optimization to jlm-opt

Felix-Rm:
This PR adds the `AAAndersenTopDownLifetimeAware` optimization. It builds on the existing `Andersen` alias analysis in combination with the `TopDownMemoryNodeEliminator`.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/Optimization.cpp
+++ b/jlm/llvm/opt/alias-analyses/Optimization.cpp
@@ -4,9 +4,9 @@
  * See COPYING for terms of redistribution.
  */
 
-#include "jlm/llvm/opt/alias-analyses/EliminatedMemoryNodeProvider.hpp"
 #include <jlm/llvm/opt/alias-analyses/AgnosticMemoryNodeProvider.hpp>
 #include <jlm/llvm/opt/alias-analyses/Andersen.hpp>
+#include <jlm/llvm/opt/alias-analyses/EliminatedMemoryNodeProvider.hpp>
 #include <jlm/llvm/opt/alias-analyses/MemoryStateEncoder.hpp>
 #include <jlm/llvm/opt/alias-analyses/Optimization.hpp>
 #include <jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.hpp>
--- a/jlm/llvm/opt/alias-analyses/Optimization.cpp
+++ b/jlm/llvm/opt/alias-analyses/Optimization.cpp
@@ -4,12 +4,14 @@
  * See COPYING for terms of redistribution.
  */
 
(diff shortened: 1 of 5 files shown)


Felix-Rm:
Test suite target in phate/llvm-test-suite#7.

Felix-Rm:
Using the optimization via the test suite currently crashes the compiler (jlm2llvm backend):

Crash found on my system during compilation of  `llvm-test-suite/tools/timeit.c` at [jlm/llvm/backend/jlm2llvm/type.cpp:152](https://github.com/phate/jlm/blob/5241cc79b2122a34a0459e070878d0b6488a4aeb/jlm/llvm/backend/jlm2llvm/type.cpp#L152) during the map lookup because of missing conversion entry for `jlm::llvm::MemoryStateType`.

jlm/llvm/opt/alias-analyses/Optimization.cpp
```diff
@@ -4,12 +4,14 @@
  * See COPYING for terms of redistribution.
  */
 
+#include "jlm/llvm/opt/alias-analyses/EliminatedMemoryNodeProvider.hpp"
```
phate, line 7: Please use `<>` for the includes. Also, the include is not alphabetically ordered.

phate: I suspect this happened by accident in the IDE.

phate (changes requested):

phate:
> Using the optimization via the test suite currently crashes the compiler (jlm2llvm backend):

> Crash found on my system during compilation of llvm-test-suite/tools/timeit.c at [jlm/llvm/backend/jlm2llvm/type.cpp:152](https://github.com/phate/jlm/blob/5241cc79b2122a34a0459e070878d0b6488a4aeb/jlm/llvm/backend/jlm2llvm/type.cpp#L152) during the map lookup because of missing conversion entry for jlm::llvm::MemoryStateType.

The type conversion should never be called in the first place with a MemoryStateType as there is no possible conversion to LLVM. Thus, the problem is already somewhere on the caller side. Let's get the two PRs in such that we can reproduce it and I might be able to take a look at it. This should not be too difficult to fix (TM).

jlm/llvm/opt/alias-analyses/Optimization.cpp
phate: I this include even needed?

Felix-Rm: > I this include even needed?

Needed on line 45 for template instantiation.

phate (approved):
LGTM

haved (approved):
Looks nice, great to start using `EliminatedMemoryNodeProvider`.

phate:
@Felix-Rm Would you like to merge?

PR #704, opened 2025-01-03
SimpleNode: convenience builder functions

caleridas:
Add constructor as well as helper function to create simple and its operator in a single call. Allow to move the created operator in (instead of copying it).

### DIFF
--- a/jlm/rvsdg/simple-node.hpp
+++ b/jlm/rvsdg/simple-node.hpp
@@ -29,6 +29,11 @@ protected:
       const SimpleOperation & op,
       const std::vector<jlm::rvsdg::output *> & operands);
 
+  SimpleNode(
+      rvsdg::Region & region,
+      std::unique_ptr<SimpleOperation> operation,
+      const std::vector<jlm::rvsdg::output *> & operands);
+
 public:
   jlm::rvsdg::simple_input *
   input(size_t index) const noexcept;
@@ -52,6 +57,15 @@ public:
       const std::vector<jlm::rvsdg::output *> & operands)
   {
     return new SimpleNode(region, op, operands);
+  }
+
+  static inline jlm::rvsdg::SimpleNode &
+  Create(
+      rvsdg::Region & region,
+      std::unique_ptr<SimpleOperation> operation,
+      const std::vector<jlm::rvsdg::output *> & operands)
+  {
+    return *new SimpleNode(region, std::move(operation), operands);
   }
 
   static inline std::vector<jlm::rvsdg::output *>
@@ -118,6 +132,91 @@ inline jlm::rvsdg::simple_output *
 SimpleNode::output(size_t index) const noexcept
 {
   return static_cast<simple_output *>(Node::output(index));
+}
+
+/**
+ * \brief Creates a simple node characterized by its operator.
+ *
+ * \tparam OperatorType
+ *   The type of operator wrapped by the node.
+ *
+ * \tparam OperatorArguments
+ *   Argument types of the operator to be constructed (should be
+ *   implied, just specify the OperatorType).
+ *
+ * \param operands
+ *   The operands to the operator (i.e. inputs to the node to be constructed).
+ *
+ * \param operatorArguments
+ *   Constructor arguments for the operator to be constructed.
+ *
+ * \returns
+ *   Reference to the node constructed.
+ *
+ * \pre
+ *   \p operands must be non-empty, must be in the same region, and their
+ *   types must match the operator constructed by this call.
+ *
+ * Constructs a new operator of type \p OperatorType using \p operatorArguments
+ * as constructor arguments. Creates a simple node using the constructed operator
+ * and the given \p operands as operands to the constructed operator.
+ *
+ * Usage example:
+ * \code
+ *   auto element_ptr = CreateOpNode<GetElementPtrOperation>(
+ *     { ptr }, offsetTypes, pointeeTypes).outputs(0);
+ * \endcode
+ */
+template<typename OperatorType, typename... OperatorArguments>
+SimpleNode &
+CreateOpNode(const std::vector<output *> & operands, OperatorArguments... operatorArguments)
+{
+  JLM_ASSERT(!operands.empty());
+  return SimpleNode::Create(
+      *operands[0]->region(),
+      std::make_unique<OperatorType>(std::move(operatorArguments)...),
+      operands);
+}
+
+/**
+ * \brief Creates a simple node characterized by its operator.
+ *
+ * \tparam OperatorType
+ *   The type of operator wrapped by the node.
+ *
+ * \tparam OperatorArguments
+ *   Argument types of the operator to be constructed (should be
+ *   implied, just specify the OperatorType).
+ *
+ * \param region
+ *   The region to create the node in.
+ *
+ * \param operatorArguments
+ *   Constructor arguments for the operator to be constructed.
+ *
+ * \returns
+ *   Reference to the node constructed.
+ *
+ * \pre
+ *   The given operator must not take any operands.
+ *
+ * Constructs a new operator of type \p OperatorType using \p operatorArguments
+ * as constructor arguments. Creates a simple node using the constructed operator
+ * with no operands in the specified region.
+ *
+ * Usage example:
+ * \code
+ *   auto val = CreateOpNode<IntegerConstantOperation>(region, 42).outputs(0);
+ * \endcode
+ */
+template<typename OperatorType, typename... OperatorArguments>
+SimpleNode &
+CreateOpNode(Region & region, OperatorArguments... operatorArguments)
+{
+  return SimpleNode::Create(
+      region,
+      std::make_unique<OperatorType>(std::move(operatorArguments)...),
+      {});
 }
 
 }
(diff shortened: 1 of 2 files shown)


caleridas:
I would have preferred different argumet ordering for the convenience function, i.e.:
CreateOpNode(<operator arguments>, { <operands< })

but variadic template arguments must be last, and there are only hacky ways using lots of template magic to have it otherwise.

jlm/rvsdg/simple-node.hpp
```diff
@@ -29,6 +29,11 @@ class SimpleNode : public Node
       const SimpleOperation & op,
       const std::vector<jlm::rvsdg::output *> & operands);
 
+  SimpleNode(
+      rvsdg::Region * region,
+      std::unique_ptr<SimpleOperation> op,
```
phate, line 34: Minor nitpick: Rename op to operation

jlm/rvsdg/simple-node.hpp
```diff
@@ -54,6 +59,15 @@ class SimpleNode : public Node
     return new SimpleNode(region, op, operands);
   }
 
+  static inline jlm::rvsdg::SimpleNode *
+  Create(
+      rvsdg::Region * region,
+      std::unique_ptr<SimpleOperation> op,
```
phate, line 65: Minor nitpick: naming

jlm/rvsdg/simple-node.hpp
```diff
@@ -54,6 +59,15 @@ class SimpleNode : public Node
     return new SimpleNode(region, op, operands);
   }
 
+  static inline jlm::rvsdg::SimpleNode *
```
phate, line 62: Could we get a reference as return type?

jlm/rvsdg/simple-node.hpp
```diff
@@ -29,6 +29,11 @@ class SimpleNode : public Node
       const SimpleOperation & op,
       const std::vector<jlm::rvsdg::output *> & operands);
 
+  SimpleNode(
+      rvsdg::Region * region,
```
phate, line 33: A reference, please.

jlm/rvsdg/simple-node.hpp
```diff
@@ -54,6 +59,15 @@ class SimpleNode : public Node
     return new SimpleNode(region, op, operands);
   }
 
+  static inline jlm::rvsdg::SimpleNode *
+  Create(
+      rvsdg::Region * region,
```
phate, line 64: a reference, please.

jlm/rvsdg/simple-node.hpp
```diff
@@ -120,6 +134,31 @@ SimpleNode::output(size_t index) const noexcept
...
 
+struct CreateOpNodeResult
+{
+  Node * node;
+  std::vector<output *> outputs;
+};
+
+template<typename OpType, typename... OpParameters>
+CreateOpNodeResult
+CreateOpNode(const std::vector<output *> & arguments, OpParameters... opParameters)
```
phate, line 145: Minor nitpick: Rename arguments to operands

phate, line 145: Could we get some documentation on this function?

jlm/rvsdg/simple-node.hpp
```diff
@@ -120,6 +134,31 @@ SimpleNode::output(size_t index) const noexcept
   return static_cast<simple_output *>(Node::output(index));
 }
 
+struct CreateOpNodeResult
```
phate, line 137: Some documentation, please.

jlm/rvsdg/simple-node.hpp
```diff
@@ -120,6 +134,31 @@ SimpleNode::output(size_t index) const noexcept
...
+{
+  Node * node;
+  std::vector<output *> outputs;
+};
+
+template<typename OpType, typename... OpParameters>
+CreateOpNodeResult
+CreateOpNode(const std::vector<output *> & arguments, OpParameters... opParameters)
+{
+  Region * region = arguments.at(0)->region();
```
phate, line 147: This method assumes at least a single operand. What about nullary nodes? Could we document the constraints as well as insert a `JLM_ASSERT()` at the beginning rather than implicit relying on c++ exceptions via `at()`

jlm/rvsdg/simple-node.hpp
```diff
@@ -120,6 +134,31 @@ SimpleNode::output(size_t index) const noexcept
...
+  for (const auto & argument : arguments)
+  {
+    JLM_ASSERT(argument->region() == region);
+  }
```
phate, line 151: This can be removed as it is checked in the constructor of the input: https://github.com/phate/jlm/blob/6db940a6220bd6be1ba2203edceb6269114ea4b6/jlm/rvsdg/node.cpp#L24C1-L40C2

jlm/rvsdg/simple-node.hpp
```diff
@@ -120,6 +134,31 @@ SimpleNode::output(size_t index) const noexcept
...
+  Node * node =
+      SimpleNode::Create(region, std::make_unique<OpType>(std::move(opParameters)...), arguments);
+  std::vector<output *> outputs;
+  for (std::size_t n = 0; n < node->noutputs(); ++n)
+  {
+    outputs.push_back(node->output(n));
+  }
```
phate, line 158: We have the outptus() function that does exactly this: https://github.com/phate/jlm/blob/6db940a6220bd6be1ba2203edceb6269114ea4b6/jlm/rvsdg/node.hpp#L1052C1-L1059C2

jlm/rvsdg/simple-node.hpp
```diff
@@ -120,6 +134,31 @@ SimpleNode::output(size_t index) const noexcept
...
 }
 
+struct CreateOpNodeResult
+{
+  Node * node;
+  std::vector<output *> outputs;
+};
+
+template<typename OpType, typename... OpParameters>
+CreateOpNodeResult
```
phate, line 144: Why does this function return the node and the results? What is the reason here? Once the caller has the node, it is no problem to also extract the results from it using the `outputs()` function. See below.

phate (changes requested):
A few rough corners that do not make sense to me, as well as some minor nitpicks that I do not expect to be fixed if you do not want to.

jlm/rvsdg/simple-node.hpp
caleridas: Returning node now.

jlm/rvsdg/simple-node.hpp
caleridas: obsolete since returning node now

jlm/rvsdg/simple-node.hpp
caleridas: removed redundant checking

jlm/rvsdg/simple-node.hpp
caleridas: added another method that allows to create nullary nodes and reworked checking

jlm/rvsdg/simple-node.hpp
caleridas: added documentation including usage example

phate (dismissed):
Looks generally fine to me. One final thought: Would it be possible to move these functions into the SimpleNode class instead of having them as free standing functions?

Usage would then be something like this: `SimpleNode::CreateOpNode<>(....)`

I also do not particularly like the name `CreateOpNode`, but I guess a simple `Create()`  or `CreateNode()` won't work as it would clash with the others already being there.

jlm/rvsdg/simple-node.hpp
```diff
@@ -120,6 +134,91 @@ SimpleNode::output(size_t index) const noexcept
...
+ *     { ptr }, offsetTypes, pointeeTypes).outputs(0);
+ * \endcode
+ */
+template<typename OperatorType, typename... OperatorArguments>
+SimpleNode &
+CreateOpNode(const std::vector<output *> & operands, OperatorArguments... operatorArguments)
+{
+  JLM_ASSERT(!operands.empty());
+  return SimpleNode::Create(
+      *operands.at(0)->region(),
```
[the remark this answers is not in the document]
caleridas: Changed array access -- one crucial difference is that JLM_ASSERT is only active in debug builds, checking with at in debug and release.

caleridas:
> Looks generally fine to me. One final thought: Would it be possible to move these functions into the SimpleNode class instead of having them as free standing functions?
> 
> Usage would then be something like this: `SimpleNode::CreateOpNode<>(....)`
> 
> I also do not particularly like the name `CreateOpNode`, but I guess a simple `Create()` or `CreateNode()` won't work as it would clash with the others already being there.

I would actually prefer free functions, I would prefer if everything in the nodes is concerned with their properties and access, but this convenience helper actually crosses "operator" and "node" so it is not exactly clear where it belongs to.

jlm/rvsdg/simple-node.hpp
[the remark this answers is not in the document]
phate: Yeah, I understand, but debug build is good enough. It is also documented that we expect the operand vector not to be empty. Should hopefully be enough to not pay the overhead.

phate (approved):
> I would actually prefer free functions, I would prefer if everything in the nodes is concerned with their properties and access, but this convenience helper actually crosses "operator" and "node" so it is not exactly clear where it belongs to.

Fair point.

PR #705, opened 2025-01-03
Function pointer conversion

caleridas:
Add function / pointer conversion llvm operators. Accept these operators as top-level nodes in the various transformation passes, and handle them in instruction conversion.

There are presently no "users" of these operators or anything that puts them into the graph -- so this is preparatory for making function and pointer distinction in a later commit.

### DIFF
--- a/jlm/llvm/Makefile.sub
+++ b/jlm/llvm/Makefile.sub
@@ -27,6 +27,7 @@ libllvm_SOURCES = \
     jlm/llvm/ir/operators/alloca.cpp \
     jlm/llvm/ir/operators/call.cpp \
     jlm/llvm/ir/operators/delta.cpp \
+    jlm/llvm/ir/operators/FunctionPointer.cpp \
     jlm/llvm/ir/operators/GetElementPtr.cpp \
     jlm/llvm/ir/operators/lambda.cpp \
     jlm/llvm/ir/operators/Load.cpp \
@@ -123,6 +124,7 @@ libllvm_HEADERS = \
 	jlm/llvm/ir/operators/Load.hpp \
 	jlm/llvm/ir/operators/MemCpy.hpp \
 	jlm/llvm/ir/operators/MemoryStateOperations.hpp \
+	jlm/llvm/ir/operators/FunctionPointer.hpp \
 	jlm/llvm/ir/operators/GetElementPtr.hpp \
 	jlm/llvm/ir/operators/delta.hpp \
 	jlm/llvm/ir/operators/Store.hpp \
--- a/jlm/llvm/ir/operators/FunctionPointer.hpp
+++ b/jlm/llvm/ir/operators/FunctionPointer.hpp
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2024 Helge Bahmann <hcb@chaoticmind.net>
+ * See COPYING for terms of redistribution.
+ */
+
+#ifndef JLM_LLVM_IR_OPERATORS_FUNCTIONPOINTER_HPP
+#define JLM_LLVM_IR_OPERATORS_FUNCTIONPOINTER_HPP
+
+#include <jlm/llvm/ir/tac.hpp>
+#include <jlm/llvm/ir/types.hpp>
+#include <jlm/rvsdg/bitstring/type.hpp>
+#include <jlm/rvsdg/unary.hpp>
+
+namespace jlm::llvm
+{
+
+/**
+  \brief Get address of compiled function object.
+  */
+class FunctionToPointerOperation final : public rvsdg::unary_op
+{
+public:
+  ~FunctionToPointerOperation() noexcept override;
+
+  FunctionToPointerOperation(std::shared_ptr<const llvm::FunctionType> fn);
+
+  bool
+  operator==(const Operation & other) const noexcept override;
+
+  [[nodiscard]] std::string
+  debug_string() const override;
+
+  [[nodiscard]] std::unique_ptr<Operation>
+  copy() const override;
+
+  rvsdg::unop_reduction_path_t
+  can_reduce_operand(const jlm::rvsdg::output * arg) const noexcept override;
+
+  jlm::rvsdg::output *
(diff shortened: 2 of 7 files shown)


caleridas:
The operator names are quite verbose -- happy to abbreviate them if desired.

jlm/llvm/Makefile.sub
```diff
@@ -123,6 +124,7 @@ libllvm_HEADERS = \
 	jlm/llvm/ir/operators/Load.hpp \
 	jlm/llvm/ir/operators/MemCpy.hpp \
 	jlm/llvm/ir/operators/MemoryStateOperations.hpp \
+    jlm/llvm/ir/operators/FunctionPointer.hpp \
```
phate, line 127: Indentation

jlm/llvm/ir/operators/FunctionPointer.hpp
```diff
@@ -0,0 +1,103 @@
...
+#include <jlm/rvsdg/bitstring/type.hpp>
+#include <jlm/rvsdg/simple-node.hpp>
+
+namespace jlm::llvm
+{
+
+/**
+  \brief Get address of compiled function object.
+  */
+class FunctionToPointerOperation final : public rvsdg::SimpleOperation
```
phate, line 20: I do not think that the constructor and creators of this class need to take the `PointerType` as argument. It can be explicit I think. The pointer type has no attributes, it is a simply opaque.

jlm/llvm/ir/operators/FunctionPointer.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+private:
+  std::shared_ptr<const llvm::FunctionType> FunctionType_;
+  std::shared_ptr<const llvm::PointerType> PointerType_;
+};
+
+/**
+  \brief Interpret pointer as callable function.
+  */
+class PointerToFunctionOperation final : public rvsdg::SimpleOperation
```
phate, line 61: The same here as above.

jlm/llvm/opt/InvariantValueRedirection.cpp
```diff
@@ -84,6 +85,21 @@ InvariantValueRedirection::RedirectInRootRegion(rvsdg::Graph & rvsdg)
...
+      }
+      else if (dynamic_cast<const PointerToFunctionOperation *>(&simple->GetOperation()))
+      {
+        // Ignore
+      }
+      else
+      {
+        JLM_UNREACHABLE("Unhandled node type.");
+      }
+    }
```
phate, line 102:
```
else if (is<FunctionToPointerOperation>(...) || is<PointerToFunctionOperation>(...))
{
  // Ignore
}
```

There is no need to first cast it to a simple node and then again to the concrete nodes. Also, for readability, we have the is function: https://github.com/phate/jlm/blob/6db940a6220bd6be1ba2203edceb6269114ea4b6/jlm/rvsdg/operation.hpp#L53C1-L60C2

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.cpp
```diff
@@ -874,6 +875,20 @@ RegionAwareMemoryNodeProvider::Propagate(const RvsdgModule & rvsdgModule)
...
+      if (dynamic_cast<const FunctionToPointerOperation *>(&n->GetOperation()))
+      {
+        continue;
+      }
+      if (dynamic_cast<const PointerToFunctionOperation *>(&n->GetOperation()))
+      {
+        continue;
+      }
+      JLM_UNREACHABLE("Unhandled node type!");
+    }
```
phate, line 891: The same issue here as mentioned above.

jlm/llvm/opt/alias-analyses/TopDownMemoryNodeEliminator.cpp
```diff
@@ -492,6 +493,21 @@ TopDownMemoryNodeEliminator::EliminateTopDownRootRegion(rvsdg::Region & region)
...
+      }
+      else if (dynamic_cast<const PointerToFunctionOperation *>(&simpleNode->GetOperation()))
+      {
+        // Nothing needs to be done.
+      }
+      else
+      {
+        JLM_UNREACHABLE("Unhandled node type!");
+      }
+    }
```
phate, line 510: Same here.

phate (changes requested):
I am not sure whether I missed something, but I do not think you need to cast it to a simple node first before you cast it to a concrete node

jlm/llvm/ir/operators/FunctionPointer.hpp
caleridas: you mean "implicit"? i.e. remove the pointerType argument and just infer PointerType::Create()?

phate: Yes, I meant implicit. My bad. Sorry!

caleridas: okay removed

jlm/llvm/opt/InvariantValueRedirection.cpp
caleridas: I still need to apply it to node->GetOperation(), though. I removed the casts to SimpleNode as I can call GetOperation also just on node.

jlm/llvm/Makefile.sub
caleridas: ah there is a mix of tab/spaces in the file :/ fixed this one now, tending to tabs/spaces at a later point

phate (dismissed):
LGTM. I did not think about it the first time, but you might want to inherit it from unary_op class instead of from SimpleOperation.

caleridas:
> LGTM. I did not think about it the first time, but you might want to inherit it from unary_op class instead of from SimpleOperation.

Okay changed it, it forces in the unary reduction methods -- which is the thing you are reworking in another commit, though, so let's see.

phate (dismissed):

phate (approved):

PR #707, opened 2025-01-03
Ignore UndefValueOperation returning MemoryStateType in llvm backend.

Felix-Rm:
This PR fixes the compiler crash mentioned in #703 by skipping over `UndefValueOperation`s with a return type of `MemoryStateType` in the `jlm2llvm` backend. These operations are generated during `MemoryStateEncoder::EncodeCallEntry` after using the lifetime aware optimization added in #703.

### DIFF
--- a/jlm/llvm/backend/jlm2llvm/instruction.cpp
+++ b/jlm/llvm/backend/jlm2llvm/instruction.cpp
@@ -162,7 +162,14 @@ convert_undef(
     context & ctx)
 {
   JLM_ASSERT(is<UndefValueOperation>(op));
-  return ::llvm::UndefValue::get(convert_type(*op.result(0), ctx));
+
+  auto & resultType = *op.result(0);
+
+  // MemoryState has no llvm representation.
+  if (is<MemoryStateType>(resultType))
+    return nullptr;
+
+  return ::llvm::UndefValue::get(convert_type(resultType, ctx));
 }
 
 static ::llvm::Value *
--- a/jlm/llvm/Makefile.sub
+++ b/jlm/llvm/Makefile.sub
@@ -153,6 +153,7 @@ libllvm_TESTS += \
     tests/jlm/llvm/backend/llvm/jlm-llvm/TestAttributeConversion \
     tests/jlm/llvm/backend/llvm/jlm-llvm/test-bitconstant \
     tests/jlm/llvm/backend/llvm/jlm-llvm/test-function-calls \
+    tests/jlm/llvm/backend/llvm/jlm-llvm/test-ignore-memory-state \
     tests/jlm/llvm/backend/llvm/jlm-llvm/test-select-with-state \
     tests/jlm/llvm/backend/llvm/jlm-llvm/test-type-conversion \
     tests/jlm/llvm/frontend/llvm/LlvmTypeConversionTests  \
--- a/tests/jlm/llvm/backend/llvm/jlm-llvm/test-ignore-memory-state.cpp
+++ b/tests/jlm/llvm/backend/llvm/jlm-llvm/test-ignore-memory-state.cpp
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2025 Felix Reißmann <felix.rm.153@gmail.com>
+ * See COPYING for terms of redistribution.
+ */
+
+#include "test-registry.hpp"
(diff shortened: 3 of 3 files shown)


jlm/llvm/backend/jlm2llvm/instruction.cpp
```diff
@@ -161,7 +161,14 @@ convert_undef(
     context & ctx)
 {
   JLM_ASSERT(is<UndefValueOperation>(op));
-  return ::llvm::UndefValue::get(convert_type(*op.result(0), ctx));
+
+  auto & result_type = *op.result(0);
```
phate, line 165: Naming: `auto & resultType = *op.result(0)`

jlm/llvm/backend/jlm2llvm/instruction.cpp
```diff
@@ -161,7 +161,14 @@ convert_undef(
...
+
+  // MemoryStates have no llvm representation.
+  if (is<MemoryStateType>(result_type))
```
phate, line 168: I think we can generalize this to `StateType`.

phate (changes requested):
Could we also get a unit on this to ensure it won't happen again in the future?

jlm/llvm/backend/jlm2llvm/instruction.cpp
Felix-Rm: Generalizing to `StateType` would also skip conversions of `ControlType`. `MemoryStateType`s were not referenced after creation (at least in my tests), making simply ignoring them possible. This seems not to be the case for `ControlType`s. Running `make check` with the proposed generalization results in a test failure of `tests/c-tests/test-irreducible.c`. The assertion at `jlm/llvm/backend/jlm2llvm/context.hpp:107` fails, being called from `jlm/llvm/backend/jlm2llvm/jlm2llvm.cpp:456`, as the referenced node was never converted and registered in the context.

phate: My bad. I did not think about it. Please leave it as it is then.

haved (dismissed):
The change makes sense. I had not realized that `ControlType` was a subclass of `StateType`. I would expect it to be a `ValueType`, since it carries a value, but I guess there is a reason it is not? @phate

phate:
@caleridas See @haved comment above. Do you remember why `ControlType` is a state type instead of a value type? Is this a "mistake" or was there a particular reason for it?

caleridas:
> @caleridas See @haved comment above. Do you remember why `ControlType` is a state type instead of a value type? Is this a "mistake" or was there a particular reason for it?

Yes there is a reason, and it relates to the usage of control in control flow reconstruction where it behaves like a state (cannot fork etc., but is actually even more strict). However However these restrctions really _only_ makes sense when attempting control flow extraction, otherwise it is kind of needlessly constraining. I think it would make sense to relax this such that gamma, theta & friends kind accept bool(ish) value types. So not sure how to proceed precisely if these constraints need to be lifted, but I can see multiple options (e.g. allowing control nodes to accept other input types, relaxing control type, ...).

I don't exactly understand how this relates to this PR, though.

haved:
@caleridas that makes sense, though I would not be against using boolean/integer values instead.

The reason this came up in this PR can be seen in one of the comments @phate made, where he suggested replacing `is<MemoryStateType>()` with just `is<StateType>()`. That turned out not to work due to control types being state types.

caleridas:
> @caleridas that makes sense, though I would not be against using boolean/integer values instead.
> 
> The reason this came up in this PR can be seen in one of the comments @phate made, where he suggested replacing `is<MemoryStateType>()` with just `is<StateType>()`. That turned out not to work due to control types being state types.

Thanks for the explanation. In hindsight I would view "controltype" as too narrow-minded, at the time I put it in thinking too much about control flow generation. I presently have only one ongoing bit of work that affects it, maybe I am going to revisit this bit.

phate:
@Felix-Rm Do you want me to finish up this branch and merge it? Since it is already open for a while....

Felix-Rm:
> @Felix-Rm Do you want me to finish up this branch and merge it? Since it is already open for a while....

I wanted to leave some time for the discussion. If the changes are okay, you can go ahead and merge this PR.

phate (approved):

PR #711, opened 2025-01-03
Move load-mux reduction tests into LoadTests.cpp

phate:
Move load-mux reduction tests into LoadTest.cpp and utilize new interface in the tests.


haved (approved):
LGTM

PR #714, opened 2025-01-03
Port unary operation reduction to new normalization interface

phate:


jlm/rvsdg/unary.cpp
```diff
@@ -95,6 +95,21 @@ unary_normal_form::set_reducible(bool enable)
...
+std::optional<std::vector<rvsdg::output *>>
+NormalizeUnaryOperation(const unary_op & operation, const std::vector<rvsdg::output *> & operands)
+{
+  JLM_ASSERT(operands.size() == 1);
+  auto & operand = *operands[0];
+
+  if (const auto reduction = operation.can_reduce_operand(&operand);
+      reduction != unop_reduction_none)
+  {
+    return std::vector(1, operation.reduce_operand(reduction, &operand));
```
haved, line 107: Would it be possible to use `std::vector{ reduce_operand(...) }` instead of the length + default value constructor?

haved (approved):
Looks very nice, just one small question about the vector constructor used

jlm/rvsdg/unary.cpp
phate: Fixed. It needed double braces for obvious reasons, but I only tried with single braces the first time and did not connect the dots.

haved (approved):

PR #717, opened 2025-01-05
Distinguish between "function" and "pointer" types

caleridas:
Make the lambda operator return a "function" object (instead of pointer) and make the call operator require a "function" object (instead of a pointer). Introduce pointer-to-function and function-to-pointer conversions in places where references to functions are stored in machine data structures or passed as machine-representable values across functions.

In pointer analysis, unify functions and pointers to functions as the same objects. Possibly function objects should be represented as "pointees" instead, but for now simply blurring the distinction between the two delivers the desired results.

### DIFF
--- a/jlm/llvm/frontend/LlvmInstructionConversion.cpp
+++ b/jlm/llvm/frontend/LlvmInstructionConversion.cpp
@@ -16,8 +16,13 @@
 namespace jlm::llvm
 {
 
+// Converts a value into a variable either representing the llvm
+// value or representing a function object.
+// The distinction stems from the fact that llvm treats functions simply
+// as pointers to the function code while we distinguish between the two.
+// This function can return either and caller needs to check / adapt.
 const variable *
-ConvertValue(::llvm::Value * v, tacsvector_t & tacs, context & ctx)
+ConvertValueOrFunction(::llvm::Value * v, tacsvector_t & tacs, context & ctx)
 {
   auto node = ctx.node();
   if (node && ctx.has_value(v))
@@ -36,6 +41,20 @@ ConvertValue(::llvm::Value * v, tacsvector_t & tacs, context & ctx)
     return ConvertConstant(c, tacs, ctx);
 
   JLM_UNREACHABLE("This should not have happened!");
+}
+
+// Converts a value into a variable representing the llvm value.
+const variable *
+ConvertValue(::llvm::Value * v, tacsvector_t & tacs, context & ctx)
+{
+  const variable * var = ConvertValueOrFunction(v, tacs, ctx);
+  if (auto fntype = std::dynamic_pointer_cast<const FunctionType>(var->Type()))
+  {
+    std::unique_ptr<tac> ptr_cast = tac::create(FunctionToPointerOperation(fntype), { var });
+    var = ptr_cast->result(0);
+    tacs.push_back(std::move(ptr_cast));
+  }
+  return var;
 }
 
 /* constant */
@@ -886,6 +905,7 @@ convert_call_instruction(::llvm::Instruction * instruction, tacsvector_t & tacs,
     return convert_memcpy_call(i, tacs, ctx);
 
   auto ftype = i->getFunctionType();
+  auto convertedFType = ConvertFunctionType(ftype, ctx);
 
   auto arguments = create_arguments(i, tacs, ctx);
   if (ftype->isVarArg())
@@ -893,8 +913,42 @@ convert_call_instruction(::llvm::Instruction * instruction, tacsvector_t & tacs,
   arguments.push_back(ctx.iostate());
   arguments.push_back(ctx.memory_state());
 
-  auto fctvar = ConvertValue(i->getCalledOperand(), tacs, ctx);
-  auto call = CallOperation::create(fctvar, ConvertFunctionType(ftype, ctx), arguments);
+  const variable * callee = ConvertValueOrFunction(i->getCalledOperand(), tacs, ctx);
+  // Llvm does not distinguish between "function objects" and
+  // "pointers to functions" while we need to be precise in modelling.
+  // If the called object is a function object, then we can just
+  // feed it to the call operator directly, otherwise we have
+  // to cast it into a function object.
+  if (is<PointerType>(*callee->Type()))
+  {
+    std::unique_ptr<tac> callee_cast =
+        tac::create(PointerToFunctionOperation(convertedFType), { callee });
+    callee = callee_cast->result(0);
+    tacs.push_back(std::move(callee_cast));
+  }
+  else if (auto fntype = std::dynamic_pointer_cast<const FunctionType>(callee->Type()))
+  {
+    // Llvm also allows argument type mismatches if the function
+    // features varargs. The code here could be made more precise by
+    // validating and accepting only vararg-related mismatches.
+    if (*convertedFType != *fntype)
+    {
+      // Since vararg passing is not modelled explicitly, simply hide the
+      // argument mismtach via pointer casts.
+      std::unique_ptr<tac> ptrCast = tac::create(FunctionToPointerOperation(fntype), { callee });
+      std::unique_ptr<tac> fnCast =
+          tac::create(PointerToFunctionOperation(convertedFType), { ptrCast->result(0) });
+      callee = fnCast->result(0);
+      tacs.push_back(std::move(ptrCast));
+      tacs.push_back(std::move(fnCast));
+    }
+  }
+  else
+  {
+    throw std::runtime_error("Unexpected callee type: " + callee->Type()->debug_string());
+  }
+
+  auto call = CallOperation::create(callee, convertedFType, arguments);
 
   auto result = call->result(0);
   auto iostate = call->result(call->nresults() - 2);
--- a/jlm/llvm/opt/alias-analyses/Andersen.cpp
+++ b/jlm/llvm/opt/alias-analyses/Andersen.cpp
@@ -19,7 +19,7 @@ namespace jlm::llvm::aa
 bool
 IsOrContainsPointerType(const rvsdg::Type & type)
 {
-  return IsOrContains<PointerType>(type);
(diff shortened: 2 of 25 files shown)


jlm/llvm/frontend/LlvmInstructionConversion.cpp
```diff
@@ -886,15 +905,37 @@ convert_call_instruction(::llvm::Instruction * instruction, tacsvector_t & tacs,
...
+    tacs.push_back(std::move(callee_cast));
+  }
+  else if (std::dynamic_pointer_cast<const FunctionType>(callee->Type()))
+  {
+    // Nothing to do.
+  }
+  else
+  {
+    throw std::runtime_error("Unexpected callee type: " + callee->Type()->debug_string());
+  }
```
phate, line 936: `JLM_UNREACHABLE()`, no?

jlm/llvm/frontend/LlvmInstructionConversion.cpp
```diff
@@ -886,15 +905,37 @@ convert_call_instruction(::llvm::Instruction * instruction, tacsvector_t & tacs,
...
+  // feed it to the call operator directly, otherwise we have
+  // to cast it into a function object.
+  if (std::dynamic_pointer_cast<const PointerType>(callee->Type()))
+  {
+    std::unique_ptr<tac> callee_cast =
+        tac::create(PointerToFunctionOperation(convertedFType), { callee });
+    callee = callee_cast->result(0);
+    tacs.push_back(std::move(callee_cast));
+  }
+  else if (std::dynamic_pointer_cast<const FunctionType>(callee->Type()))
```
phate, line 929: `is<FunctionType>()`

jlm/llvm/frontend/LlvmInstructionConversion.cpp
```diff
@@ -886,15 +905,37 @@ convert_call_instruction(::llvm::Instruction * instruction, tacsvector_t & tacs,
...
 
-  auto fctvar = ConvertValue(i->getCalledOperand(), tacs, ctx);
-  auto call = CallOperation::create(fctvar, ConvertFunctionType(ftype, ctx), arguments);
+  const variable * callee = ConvertValueOrFunction(i->getCalledOperand(), tacs, ctx);
+  // Llvm does not distinguish between "function objects" and
+  // "pointers to functions" while we need to be precise in modelling.
+  // If the called object is a function object, then we can just
+  // feed it to the call operator directly, otherwise we have
+  // to cast it into a function object.
+  if (std::dynamic_pointer_cast<const PointerType>(callee->Type()))
```
phate, line 922: `is<PointerType>()`

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -19,7 +19,7 @@ namespace jlm::llvm::aa
 bool
 IsOrContainsPointerType(const rvsdg::Type & type)
 {
-  return IsOrContains<PointerType>(type);
+  return IsOrContains<PointerType>(type) || dynamic_cast<const llvm::FunctionType *>(&type);
```
phate, line 22: `is<FunctionType>()`

Also the name of the function fits not what it does any longer.

jlm/llvm/opt/alias-analyses/Steensgaard.cpp
```diff
@@ -24,7 +24,8 @@ namespace jlm::llvm::aa
 static bool
 HasOrContainsPointerType(const rvsdg::output & output)
 {
-  return IsOrContains<PointerType>(output.type());
+  return IsOrContains<PointerType>(output.type())
+      || dynamic_cast<const llvm::FunctionType *>(&output.type());
```
phate, line 28: `is<FunctionType>()`

Also, the name of the function does not really reflect its body any longer

jlm/llvm/opt/alias-analyses/Steensgaard.cpp
```diff
@@ -1060,10 +1061,22 @@ Steensgaard::AnalyzeSimpleNode(const jlm::rvsdg::SimpleNode & node)
...
+    AnalyzeFunctionToPointer(node);
+  }
   else if (is<FreeOperation>(&node) || is<ptrcmp_op>(&node))
   {
     // Nothing needs to be done as FreeOperation and ptrcmp_op do not affect points-to sets
   }
+  else if (is<FunctionToPointerOperation>(&node))
+  {
+    AnalyzeFunctionToPointer(node);
+  }
```
phate, line 1079: This is redundant, you already have it inserted above, no?

phate (changes requested):

haved:
Am I correct in my understanding that functions with identical signatures will have identical types? (Or is this more of a C++ Functor type thing?)

If so, I guess the effect of this change is that function pointers are no longer opaque? Since you mention being able to convert between pointer values and function values, I can only understand that the function type must have a runtime representation that is identical to pointers?

In a situation where a gamma node outputs a different function based on which region it took, a call to the result would likely be indirect. Does this mean the gamma is forced to convert the function values to pointer types, or is the choice of direct vs indirect orthogonal to pointer vs function type?

caleridas:
> Am I correct in my understanding that functions with identical signatures will have identical types? (Or is this more of a C++ Functor type thing?)

Yes that is correct. Distinguishing functors vs functions... in C++ realization of functors it is a function with a hidden implicit parameter, so if you want to model the machine-level representation of C++ functors then it would be a struct with that and not just a lambda. OTOH if you want a high-level abstract representation of functors (like e.g. I will be doing in the Coq representation) then lambda can unify functions/functors, but when desugaring this down towards machine-level representation the context eventually needs to be accounted for.

> If so, I guess the effect of this change is that function pointers are no longer opaque? Since you mention being able to convert between pointer values and function values, I can only understand that the function type must have a runtime representation that is identical to pointers?

There are two things: pointers (entirely opaque) and function types (specific signatures). There is no "runtime representation of a function type" at all. For direct calls there is not even a need to assume a pointer/address (e.g. you might end up inlining the call, or just relative branch or whatever). Pointers OTOH are untyped and have a runtime representation. But an indirect call using a pointer from a high-level POV makes an assumption that the thing behind the pointer is a specific function type. That was already encoded in the call operator having a function type embedded, so doing sort of an "implicit cast". The function/pointer distinction does in principle nothing else but externalizing the assertion behind this cast into a separate kind of node.

> In a situation where a gamma node outputs a different function based on which region it took, a call to the result would likely be indirect. Does this mean the gamma is forced to convert the function values to pointer types, or is the choice of direct vs indirect orthogonal to pointer vs function type?

Representation-wise, there is no problem with gamma returning function objects (assuming they have same types), and then have a call operator applied to this. (This is actually an extremely common pattern in translating dependently-typed/pattern matching constructs, often they are "forced" to return a function operator from a match expression that can be called subsequently).

There is an orthogonal question how to lower this towards machine code (or llvm IR which is just machine code in disguise): You could either "pull the call into the gammas" (which e.g. for dependently-typed translations is only "legal" after e.g. proof expressions and types have become opaque/unified and may therefore only be possible after some transformation steps), or you could decide to reformulate the construct by pushing function-to-pointer casts into the gamma and pointer-to-function casts before the call.

So there is a separation of "what is representable" vs. "what precise representation do you desire for what purpose".

caleridas:
There is a systematic cause for this test failure, evident after adding more debug info to function types etc:

what():  invalid type, expected fun (vararg, iostate, mem) -> (ptr, iostate, mem) received fun (bit32, vararg, iostate, mem) -> (ptr, iostate, mem)

At present, the pointer reinterpretation is also used to "mask" the mismatch resulting from strict argument type matching. I think that varargs should probably be modelled more precisely, but I am going to simply fix this problem for now by inserting an artificial function->pointer->function cast sequence. That makes the test pass and should be enough to unblock this PR for the moment.

@phate

jlm/llvm/frontend/LlvmInstructionConversion.cpp
caleridas: no, I now need the type

jlm/llvm/opt/alias-analyses/Andersen.cpp
caleridas: I am not sure what to do here. Pointer analysis wants to track this as a pointer since it considers functions and pointers to be the same thing (instead of functions being only pointees). So from its point of view we have to "pretend" that it is a pointer indeed and there is justification for keeping the name.

OTOH if consequently go and rename every place that now may refer to either "a pointer or a function" appropriately, this causes to rename pretty much everything in pointer analysis.

jlm/llvm/opt/alias-analyses/Steensgaard.cpp
caleridas: indeed, removed

jlm/llvm/frontend/LlvmInstructionConversion.cpp
caleridas: I would use "unreachable" for logic errors -- i.e. things that indicate an intrinsic problem in data structures or algorithms used, as opposed to "bad user input". But renamed for now, the distinction between these is not sharp in jlm.

caleridas: reverted this -- JLM_UNREACHABLE expects char* not std::string, so I cannot pass information about the type seen, losing information valuable for debugging. Preferring to keep std::runtime_error.

jlm/llvm/frontend/LlvmInstructionConversion.cpp
caleridas: done

jlm/llvm/opt/alias-analyses/Steensgaard.cpp
caleridas: see above

jlm/llvm/opt/alias-analyses/Andersen.cpp
phate: Fair point.

phate (approved):

PR #719, opened 2025-01-06
Split up huge bitslice and bitconcat test into smaller unit tests

phate:


phate:
It is hard to clean up and remove the node normal forms with big convoluted tests.

caleridas (approved):
Makes sense

haved (approved):
The new tests look very nice, though it is a bit spooky to see that the bitstring reductions are being performed automatically without any kind of `finalize()` or `normalize()`. I assume this will be changed in the future, when we make the reductions explicit.

My only comment is a silly one, which I will immediately mark as resolved @phate

PR #720, opened 2025-01-06
Clean up theta dead node elimination

caleridas:
Restructure dead node elimination such that it never requires "unbalanced" theta inputs / outputs -- interleave cleaning up the subregion instead.

Remove the support from theta node to deal with temporary invariant violation by inbalanced input / output. Introduce a method instead that can remove loop-invariant dead variables. It turns out that this is all that dead node elimination needs in practice.

### DIFF
--- a/jlm/rvsdg/theta.cpp
+++ b/jlm/rvsdg/theta.cpp
@@ -47,6 +47,31 @@ ThetaNode::AddLoopVar(rvsdg::output * origin)
   return LoopVar{ input, &thetaArgument, &thetaResult, output };
 }
 
+void
+ThetaNode::RemoveLoopVars(std::vector<LoopVar> loopvars)
+{
+  // Sort them by descending index to avoid renumbering issues.
+  std::sort(
+      loopvars.begin(),
+      loopvars.end(),
+      [](const LoopVar & x, const LoopVar & y)
+      {
+        return x.input->index() > y.input->index();
+      });
+
+  for (const auto & loopvar : loopvars)
+  {
+    JLM_ASSERT(loopvar.output->IsDead());
+    JLM_ASSERT(loopvar.post->origin() == loopvar.pre);
+    JLM_ASSERT(loopvar.pre->nusers() == 1);
+
+    subregion()->RemoveResult(loopvar.post->index());
+    subregion()->RemoveArgument(loopvar.pre->index());
+    RemoveOutput(loopvar.output->index());
+    RemoveInput(loopvar.input->index());
+  }
+}
+
 ThetaNode *
 ThetaNode::copy(rvsdg::Region * region, rvsdg::SubstitutionMap & smap) const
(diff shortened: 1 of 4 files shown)


caleridas:
Throw an error in hls test suite that I cannot reproduce locally:

Building: dynamatic/gemver
Linking in '/tmp/gemver.c_686476290//gemver-IMY1CX-clang.ll'
Writing bitcode...
jlm-hls: ./jlm/rvsdg/node.hpp:686: jlm::rvsdg::node_input *jlm::rvsdg::Node::input(size_t) const: Assertion `index < ninputs()' failed.
Aborted (core dumped)
make: *** [Makefile.sub:36: /home/runner/work/jlm/jlm/usr/hls-test-suite//build/dynamatic/gemver.hls] Error 1
make: *** Waiting for unfinished jobs....

The very same test runs without issues on my machine, hard to figure what is happening without a backtrace.

jlm/rvsdg/theta.cpp
```diff
@@ -113,9 +136,8 @@ ThetaNode::MapPostLoopVar(const rvsdg::input & result) const
     // There is nothing sensible to return here.
     throw std::logic_error("cannot map loop continuation predicate to loop variable");
   }
-  auto peer = MapOutputToInputIndex(result.index() - 1);
-  return LoopVar{ peer ? input(*peer) : nullptr,
-                  peer ? subregion()->argument(*peer) : nullptr,
+  return LoopVar{ input(result.index()),
+                  subregion()->argument(result.index()),
```
haved, line 140: These two lines (139, 140) need `result.index() - 1`. This fixed the crash

haved: @caleridas We tracked the crash to this

caleridas: That is the site of the crash but not the cause: I restructured DNE such that unbalanced input/output nodes never occur in loop treatment -- for jlm/llvm/opt/DeadNodeElimination.cpp, that is, that also has unit test coverage to validate this invariant.

However I suspect that the actual error is in jlm/hls/backend/rvsdg2rhls/DeadNodeElimination.cpp which has some open-coded treatment of input removal that then also breaks invariants. I will keep adding invariant checks until finding this, but since it does not reproduce on my machine I will have a lot of CI roundtrips to see where it is coming from. Maybe if you have a reproducer you could also take a look?

haved: I agree that there are issues in other places, since the local reproduction was non-deterministic.

But do you not agree that the input and argument should have `index - 1`, when `index` is the index of the result? The list of results has the extra predicate input, so all other indices need to be offset by 1? @caleridas

caleridas: oooh sorry my bad -- yes you are spot on, that should do it

PR #721, opened 2025-01-07
Generate comment in PR from CI if cycles differ

sjalander:
If the cycle count for one or more of the hls test suite benchmarks results in a different cycle count than expected, then a comment is generated that appears in the PR.


haved (approved):
Looks good to me

phate (approved):

PR #726, opened 2025-01-08
Simplify HLS loop dead node elimination

caleridas:
After removing theta invariant violation from dead node elimination from llvm/opt, try to make HLS handling similarly regular.

### DIFF
--- a/jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
@@ -14,6 +14,12 @@ namespace jlm::hls
 {
 
 static bool
+IsPassthroughLoopVar(const rvsdg::ThetaNode::LoopVar & loopvar)
+{
+  return loopvar.pre->nusers() == 1 && loopvar.post->origin() == loopvar.pre;
+}
+
+static bool
 IsPassthroughArgument(const rvsdg::output & argument)
 {
   if (argument.nusers() != 1)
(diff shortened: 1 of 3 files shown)


caleridas:
This _looks_ like the right thing to do, but I am not entirely certain of the intent of why hls dead node elimination is divergent.

phate:
The UnusedStateRemoval transformation is currently nowhere used except in its own tests. The original transformation is in remove-unused-state.hpp/*.cpp, but this one had some problems which stepped on my foot while refactoring (it is/was buggy). So I cleaned it up, implemented some tests along with it to highlight the bugs (while preserving the buggy semantics), and replaced the usage of the old one with the new one. After that, the new and refactored version was used. At some point, @sjalander came and needed to merge something back again from David and the newly used one was replaced again with the old messy one. Now, the newly refactored one is lying around, while the old one is still used. In other words, it is a mess, and we went often one step forward with the HLS part and then two steps backward again.

sjalander:
The UnusedStateRemoval/remove-unused-state pass operates on an RVSDG/LLVM graph, i.e., there shouldn't be any HLS specific nodes in the graph. Therefore, existing passes should be able to replace this code, e.g., InvariantValueRedirection might be able to do the same thing. @phate @caleridas @haved 

NTNU's cluster is being updated today and tomorrow, so I currently don't have a setup where I can try to make changes.

phate:
So if I see this correctly, it might be that we do not even need this PR:
1. distribute constant is discussed in #734 whether it should be removed
2. and UnusedStateRemoval might be able to be replaced by InvariantValueRedirection

sjalander:
@caleridas @phate should we merge this PR?

PR https://github.com/phate/jlm/pull/833 took the UnusedStateRemoval code into use, making this PR more relevant. The distribute_constants is still part of the HLS flow as it is about optimizing the final timing of the synthesized accelerator, making it more challenging to assess potential impact if it is removed.

PR #731, opened 2025-01-09
Make source file optional in Statistics class

phate:
It could be that we want to gather statistics about the modules that were not created from a file. Support this in the statistics class. This in turn will allow to push the optional source file to the rvsdg::RvsdgModule.


sjalander (approved):

phate:
Not longer required.

PR #733, opened 2025-01-09
Avoid creating an additional constant in a region where the constant already exists

haved:
When a theta is given a passthrough constant, it can redirect users of the output to the original constant.


phate:
@haved This might be the right time to write some unit tests for this pass?

haved:
@phate Magnus and I looked at what the NodeReduction pass accomplishes, and it seems good at propagating constants, in addition to doing constant folding. Do we need a separate pass in hls that does this, or could we use shared code instead?

phate:
@haved I do not know what the distribute constants pass does, but if it really only performs constant folding and propagation, then this is better done using node reductions.

sjalander:
The distributed constant pass doesn't have any effect on the results. So I've opened a PR that removes it #734. We should try to add NodeReduction as an optimization.

sjalander (approved):

PR #734, opened 2025-01-09
DO NOT MERGE: Remove ineffective distribute-constants optimization

sjalander:
The distributed-constants optimization has no effect on the cycle count for the hls test suite. So, it is being removed.


haved:
Woah, I tried replacing it with NodeReduction and got errors, I had not considered just downright removing it. 

I think the reason there was a dedicated "move constants in" pass was to avoid buffers holding constant values being created, e.g. before thetas. Do we have any quick way of determining if this change has a drastic effect on the size of the HLS?

sjalander:
I don't think we currently have any quick way of doing it. But I guess it wouldn't be difficult to count nodes and different types of nodes in the final graph and compare numbers.

sjalander:
@haved Adding the NodeReduction before mem_sep_argument() causes an assert to be triggered similarly if it is placed before mem_queue(). I've not debugged these to find out what assumptions that break.
It is possible to add the NodeReduction after memstate_conv(), but without adding the statistics mentioned earlier, it's not that easy to say if it has a benefit.

sjalander:
@haved
I also replaced the unused state with llvm::InvariantValueRedirection.
Can you check with your new statistics option to see if this has an impact on the number of nodes.

I was not able to eliminate the code completely, as the remove_unused_state() is also used in the mem_conv() pass, and simply replacing it there caused logical faults.

PR #737, opened 2025-01-11
Fixed: TestMlirToJlmConverter was forced to suspend.

urays:
Under the LLVM-18.1.8 version, test (mlir/frontend/TestMlirToJlmConverter) was forced to suspend.

When I execute the make check command, the following message appears:
```
Running test build/tests/jlm/mlir/backend/TestJlmToMlirConverter ... success
Running test build/tests/jlm/mlir/frontend/TestMlirToJlmConverter ... failed. See build/tests/jlm/mlir/frontend/TestMlirToJlmConverter.log:
jlm/mlir/frontend/TestGammaGen
Creating MLIR context
Creating function arguments
Creating function results
Creating LambdaRefType
Creating function attributes
Creating LambdaNode
Adding arguments to the region
Creating LambdaResult
Creating OmegaResult
Converting MLIR to RVSDG
TestMlirToJlmConverter: /home/urays/devspace/llvm-project/18.1.8/mlir/lib/IR/Block.cpp:239: Operation *mlir::Block::getTerminator(): Assertion `mightHaveTerminator()' failed.
Aborted (core dumped)
make: *** [Makefile.rules:129: run-build/tests/jlm/mlir/frontend/TestMlirToJlmConverter] Error 134
```
---
This seems to be caused by differences in mlir versions. I compared the code of mlir-18.1.8 with that of mlir-18.init.
> mlir-18init:
```c++
  //===--------------------------------------------------------------------===//
  // Terminator management
  //===--------------------------------------------------------------------===//

  /// Get the terminator operation of this block. This function asserts that
  /// the block has a valid terminator operation.
  Operation *getTerminator();
```
```c++
//===----------------------------------------------------------------------===//
// Terminator management
//===----------------------------------------------------------------------===//

/// Get the terminator operation of this block. This function asserts that
/// the block has a valid terminator operation.
Operation *Block::getTerminator() {
  assert(!empty() && back().mightHaveTrait<OpTrait::IsTerminator>());
  return &back();
}
```
> mlir-18.1.8 (and later versions):
```c++
  //===--------------------------------------------------------------------===//
  // Terminator management
  //===--------------------------------------------------------------------===//

  /// Get the terminator operation of this block. This function asserts that
  /// the block might have a valid terminator operation.
  Operation *getTerminator();

  /// Check whether this block might have a terminator.
  bool mightHaveTerminator();
```
```c++
//===----------------------------------------------------------------------===//
// Terminator management
//===----------------------------------------------------------------------===//

/// Get the terminator operation of this block. This function asserts that
/// the block might have a valid terminator operation.
Operation *Block::getTerminator() {
  assert(mightHaveTerminator());
  return &back();
}

/// Check whether this block might have a terminator.
bool Block::mightHaveTerminator() {
  return !empty() && back().mightHaveTrait<OpTrait::IsTerminator>();
}
```
---
So, I try to modify the JLM/mlir/fronted/MlirToJlmConveter. CPP: ConverBlock (...). "
Now, when I recompiled and ran the command “make check”, all the tests worked.

### DIFF
--- a/jlm/mlir/frontend/MlirToJlmConverter.cpp
+++ b/jlm/mlir/frontend/MlirToJlmConverter.cpp
@@ -84,10 +84,10 @@ MlirToJlmConverter::ConvertBlock(::mlir::Block & block, rvsdg::Region & rvsdgReg
   // Create an RVSDG node for each MLIR operation and store each pair in a
   // hash map for easy lookup of corresponding RVSDG nodes
   std::unordered_map<::mlir::Operation *, rvsdg::Node *> operationsMap;
+  ::llvm::SmallVector<jlm::rvsdg::output *> inputs;
   for (auto & mlirOp : block.getOperations())
   {
-    ::llvm::SmallVector<jlm::rvsdg::output *> inputs =
-        GetConvertedInputs(mlirOp, operationsMap, rvsdgRegion);
+    inputs = GetConvertedInputs(mlirOp, operationsMap, rvsdgRegion);
 
     if (auto * node = ConvertOperation(mlirOp, rvsdgRegion, inputs))
     {
@@ -96,9 +96,13 @@ MlirToJlmConverter::ConvertBlock(::mlir::Block & block, rvsdg::Region & rvsdgReg
   }
 
   // The results of the region/block are encoded in the terminator operation
-  ::mlir::Operation * terminator = block.getTerminator();
+  if (block.mightHaveTerminator())
+  {
+    ::mlir::Operation * terminator = block.getTerminator();
+    inputs = GetConvertedInputs(*terminator, operationsMap, rvsdgRegion);
+  }
 
-  return GetConvertedInputs(*terminator, operationsMap, rvsdgRegion);
+  return inputs;
 }
 
 rvsdg::Node *


jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -96,9 +96,13 @@ MlirToJlmConverter::ConvertBlock(::mlir::Block & block, rvsdg::Region & rvsdgReg
   }
 
   // The results of the region/block are encoded in the terminator operation
-  ::mlir::Operation * terminator = block.getTerminator();
```
sjalander, line 99: An MLIR block is the equivalent of a region in RVSDG in this case, as MLIR doesn't allow regions without at least one block. The Terminator is what encodes the results/outputs of a region, so not having a Terminator is a region without outputs, i.e., it should be dead. If we want to allow dead regions/structural-nodes to be generated in MLIR, then a Terminator with no results should be added to the block, e.g., ::mlir::rvsdg::ThetaResult or ::mlir::rvsdg::GammaResult.

I'm not 100% sure what the issue is with the new MLIR version. But if it is that there isn't always a Terminator in the block, then wouldn't it be better to modify the MLIR backend to assure that there is always a block being insterted in each an every block?

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -96,9 +96,13 @@ MlirToJlmConverter::ConvertBlock(::mlir::Block & block, rvsdg::Region & rvsdgReg
...
+  if (block.mightHaveTerminator())
+  {
+    ::mlir::Operation * terminator = block.getTerminator();
+    inputs = GetConvertedInputs(*terminator, operationsMap, rvsdgRegion);
+  }
 
-  return GetConvertedInputs(*terminator, operationsMap, rvsdgRegion);
+  return inputs;
```
sjalander, line 105: The internal nodes in the region are not accessible from outside the structural node. So it should only be the results encoded in the Terminator, i.e., the outputs of the structural node, that should be relevant to return here.

Returning also the internal inputs shouldn't create a logical fault. But the unordered_map would in the end contain the inputs of all the nodes for the whole RVSDG graph, instead of only tracking the inputs relevant within each region.

sjalander (commented):
I might have misunderstood something, but if possible, I think the fix maybe should be done in the backend instead of frontend.

jlm/mlir/frontend/MlirToJlmConverter.cpp
urays: Oh, you're right！I did make a mistake.
Thank you !

PR #741, opened 2025-01-11
Remove binary normal form class

phate:


haved (approved):
Very very nice

caleridas (approved):
Fine to remove, the present form is not suitable. No idea how exactly the logically equivalent normalizations / transformations will eventually work out.

PR #742, opened 2025-01-11
Node operator type safety

caleridas:
Merge two commits:

1.  Make node operations type-safe
    
    Only override Node::GetOperation in the most derived subclasses, and
    let them store properly typed operation classes (instead of downcasting
    them on each access).
    
    This later also allows to more flexibly handle contents of operations in
    a node-specific fashions.

2.  Avoid operator copy on construction of SimpleNode
    
    Remove remaining constructor of SimpleNode that makes a copy of the given
    operator. Leave only the one that allows moving an operator in.

This moves the architecture further towards making the nodes mostly "shells"
that contain graph-connectivity information, while the business logic is associated
with the operators.


phate (approved):
LGTM

PR #743, opened 2025-01-11
Factor out CallSummary from lambda node

caleridas:
This is in preparation of moving lambda to generic rvsdg code. Computing call summary is actually a graph-level analysis pass that does not belong to the lambda node itself actually. Additionally, the call summary uses llvm-specific structures.

While changing this, also remove the unnecessary pointer indirection on call return -- just return the struct as copy. Remove redundante IsExported method from lambda node that just computes CallSummary under the hood. This happens to fix an accidental inefficiency where the call summary was computed twice (once implicitly, once explicitly) instead of being cached.


caleridas:
Notably, there are only two users of this (apart from unit tests): Steensgard alias analysis, and inliner. Both have very specific semantics and requirements that go well beyond any lambda node itself, so I would argue from that POV alone it should be separated. Additionally, there are llvm-specific assumptions in all of these that do not hold for generic graphs.

jlm/llvm/Makefile.sub
```diff
@@ -108,6 +109,7 @@ libllvm_HEADERS = \
 	jlm/llvm/ir/linkage.hpp \
 	jlm/llvm/ir/Annotation.hpp \
 	jlm/llvm/ir/attribute.hpp \
+    jlm/llvm/ir/CallSummary.hpp \
```
phate, line 112: Indentation

phate (changes requested):
I have one additional request. You factored out the call summary into its own file, but its tests are still in the TestLamba.cpp. Could you also create its own test file CallSummaryTests.cpp and move them there?

caleridas:
> I have one additional request. You factored out the call summary into its own file, but its tests are still in the TestLamba.cpp. Could you also create its own test file CallSummaryTests.cpp and move them there?

Hm, I will take a look but I suspect it presently causes a lot of merge conflicts with things I have in my working branch (reshuffling function type, further clean up of lambda methods, moving lambda node ...).

caleridas:
> > I have one additional request. You factored out the call summary into its own file, but its tests are still in the TestLamba.cpp. Could you also create its own test file CallSummaryTests.cpp and move them there?
> 
> Hm, I will take a look but I suspect it presently causes a lot of merge conflicts with things I have in my working branch (reshuffling function type, further clean up of lambda methods, moving lambda node ...).

Moved into separate unit test.

phate (approved):

PR #744, opened 2025-01-11
Add common node elimination normalization

phate:


jlm/rvsdg/simple-node.cpp
```diff
@@ -136,4 +136,39 @@ SimpleNode::copy(rvsdg::Region * region, SubstitutionMap & smap) const
...
+      {
+        return outputs(&node);
+      }
+    }
+  }
+  else
+  {
+    for (const auto & user : *operands[0])
+    {
+      if (const auto node = TryGetOwnerNode<SimpleNode>(*user); isCongruent(*node))
```
caleridas, line 164: this is a very weird use of sequence point in a conditional, maybe just formulate it in a more elementary/simplistic fashion, even if produces more lines?

caleridas (approved):
Looks good, only one stylistic comment (not essential).

jlm/rvsdg/simple-node.cpp
phate: Done.

jlm/rvsdg/NodeNormalization.hpp
```diff
@@ -51,6 +51,14 @@ ReduceNode(const NodeNormalization<TOperation> & nodeNormalization, Node & node)
   if (auto results = nodeNormalization(*operation, operands))
   {
     divert_users(&node, *results);
+    if (node.output(0)->nusers() != 0)
+    {
+      // In case of common node elimination, it can happen that the outputs of the current node are
+      // returned. The diversion of its users above is therefore a no-op. However, this means that
+      // the outputs of node still have users, and we can not just remove the node below.
```
caleridas, line 58: s/outputs of node/outputs of the node/

caleridas (dismissed):

jlm/rvsdg/simple-node.cpp
```diff
@@ -136,4 +136,42 @@ SimpleNode::copy(rvsdg::Region * region, SubstitutionMap & smap) const
...
+    {
+      if (isCongruent(node))
+      {
+        return outputs(&node);
+      }
+    }
+  }
+  else
+  {
+    for (const auto & user : *operands[0])
```
haved, line 162: This for loop does not skip the user that is the operation currently being considered for reduction. Is this on purpose?

Put another way, isn't this for loop guaranteed to find a congruent operation?

jlm/rvsdg/NodeNormalization.hpp
phate: Done.

jlm/rvsdg/simple-node.cpp
phate: > This for loop does not skip the user that is the operation currently being considered for reduction. Is this on purpose?

> Put another way, isn't this for loop guaranteed to find a congruent operation?

Yes, it is deliberate, and yes the loop is guaranteed to find a congruent operation IF it was invoked with an operation and operands from a node in the same region as the node.

> I see that the "driver code" has an extra check to avoid problems caused by returning the same results as the operation already had, but this driver code only checks the first output, which feels a little arbitrary.

> Would it be nicer to instead check that the operation of the other user does not have the same address as the operation being reduced? SimpleNode holds unique_ptrs to SimpleOperations, so that check should not have false positives.

Do you mean that the isCongruent() check should be: `node.GetOperation() == operation && operands == rvsdg::operands(&node) && &node.GetOperation() != &operation`? That could work if I am not mistaken and it would avoid the extra if-statement in the ReduceNode().

haved: Yes that is what I meant. The only issue I have with the current solution is that it feels a bit weird to only check if the first output is still in use, even if the node has multiple outputs. I understand why it works, and the comment correctly points the reader to common node elimination, so it is no big deal, and works well

haved: Actually, what would happen if the 0th output of the node was dead to begin with? Then the check would not realize that the node has not been eliminated

phate: > Actually, what would happen if the 0th output of the node was dead to begin with? Then the check would not realize that the node has not been eliminated?

Very good point. This would not have worked.

@haved I changed it now to your suggestion. However, this approach does have the drawback that it would not work for simple nodes with static operations, i.e., not parametrized. I dot not think that we have any right now though.

FYI @caleridas

haved (approved):

PR #745, opened 2025-01-12
Move FunctionType from llvm to rvsdg

caleridas:
Move the definition of FunctionType from llvm submodule to core rvsdg.

Also, make its debug output a bit more verbose.


jlm/rvsdg/FunctionType.cpp
```diff
@@ -0,0 +1,115 @@
...
+    if (first)
+    {
+      ss << ", ";
+      first = false;
+    }
+    ss << argtype->debug_string();
+  }
+
+  ss << ") -> (";
+  for (const auto & restype : ResultTypes_)
```
phate, line 42: I think you forgot to set first back to true here.

jlm/rvsdg/FunctionType.hpp
```diff
@@ -0,0 +1,92 @@
...
+class FunctionType final : public ValueType
+{
+public:
+  ~FunctionType() noexcept override;
+
+  FunctionType(
+      std::vector<std::shared_ptr<const jlm::rvsdg::Type>> argumentTypes,
+      std::vector<std::shared_ptr<const jlm::rvsdg::Type>> resultTypes);
+
+  inline const std::vector<std::shared_ptr<const jlm::rvsdg::Type>> &
```
phate, line 33: minor nitpick: remove inline

jlm/rvsdg/FunctionType.hpp
```diff
@@ -0,0 +1,92 @@
...
+    return ArgumentTypes_;
+  }
+
+  const std::vector<std::shared_ptr<const jlm::rvsdg::Type>> &
+  Results() const noexcept
+  {
+    return ResultTypes_;
+  }
+
+  size_t
```
phate, line 45: nodiscard

jlm/rvsdg/FunctionType.hpp
```diff
@@ -0,0 +1,92 @@
...
+  NumResults() const noexcept
+  {
+    return ResultTypes_.size();
+  }
+
+  size_t
```
phate, line 51: nodiscard

jlm/rvsdg/FunctionType.hpp
```diff
@@ -0,0 +1,92 @@
...
+  NumArguments() const noexcept
+  {
+    return ArgumentTypes_.size();
+  }
+
+  const jlm::rvsdg::Type &
```
phate, line 57: nodiscard

jlm/rvsdg/FunctionType.hpp
```diff
@@ -0,0 +1,92 @@
...
+  ResultType(size_t index) const noexcept
+  {
+    JLM_ASSERT(index < ResultTypes_.size());
+    return *ResultTypes_[index];
+  }
+
+  const jlm::rvsdg::Type &
```
phate, line 64: nodiscard

jlm/rvsdg/FunctionType.hpp
```diff
@@ -0,0 +1,92 @@
...
+      std::vector<std::shared_ptr<const jlm::rvsdg::Type>> argumentTypes,
+      std::vector<std::shared_ptr<const jlm::rvsdg::Type>> resultTypes);
+
+  inline const std::vector<std::shared_ptr<const jlm::rvsdg::Type>> &
+  Arguments() const noexcept
+  {
+    return ArgumentTypes_;
+  }
+
+  const std::vector<std::shared_ptr<const jlm::rvsdg::Type>> &
```
phate, line 39: nodiscard

jlm/rvsdg/FunctionType.hpp
phate: nodiscard

caleridas: why not writing out inline? I tend to use it to mark things that are really explicitly marked to be inlined (it is not required, similar to override, but documents the intent)

jlm/rvsdg/FunctionType.hpp
caleridas: I am BTW inclined to remove all those convenience accosser members -- just exposing Results() is enough, caller can just apply "size" to that etc.

jlm/rvsdg/FunctionType.cpp
caleridas: Not only that, but part of the logic was backwards. Fixed it now.

phate (approved):

PR #746, opened 2025-01-13
Add general debug output file creation to StatisticsCollector

haved:
Since the `StatisticsCollector` is sent around to the different passes, I thought it made sense to use it as a general helper for creating debug output files. The `RvsdgTreePrinterPass` already used the same folder, but it was manually configured to do so.

The `StatisticsCollectorSettings` now consists of four parts:
 - The demanded statistics.
 - The output directory, which is `/tmp/jlm` by default. If the folder does not exist when a request for an output file is made, the directory is made (max one dir deep).
 - The module name, which becomes the prefix of all files created by the collector. It is typically the basename of the input file, e.g., "test-loops".
 - A unique string, which is 6 random alphanumeric chars by default. Importantly, the same 6 chars are used by all files created by the same `StatisticsCollector`.

Using this, the `RvsdgTreePrinterPass` now gets the paths for its output files from the collector. The collector also does the counting when the pass is called several times.

This change required some added functionality to the `file.hpp` utilities, with unit tests written.

Lastly, a `StatisticsCollector` is created for `jlm-hls`, and sent in to the functions running passes.
Actually making `jlm-hls`, `jhls`, and `jlc` take command line parameters to configure these collectors remains.


jlm/util/file.hpp
```diff
@@ -18,15 +18,15 @@ namespace jlm::util
 class filepath final
 {
 public:
-  inline filepath(const std::string & path)
+  filepath(const std::string & path)
```
phate, line 21: Since you touch it, it should have explicit.

jlm/util/file.hpp
```diff
@@ -174,19 +215,42 @@ class filepath final
     return std::filesystem::is_regular_file(fileStatus);
   }
 
-  inline std::string
+  /**
+   * Creates a directory with the given filepath.
```
phate, line 219: The method does not take a file path any longer

jlm/util/file.hpp
```diff
@@ -174,19 +215,42 @@ class filepath final
...
+   *
+   * @throws jlm::util::error if the filepath aleady exists, if the parent directory does not exist,
+   * or if any other filesystem error occurs
```
phate, line 222: You could also just say: if an error occurs. Is it necessary to explicitly mention that the file path already exist or that the parent directory does not exist.

tests/jlm/tooling/TestJlmOptCommand.cpp
```diff
@@ -51,6 +52,9 @@ TestStatistics()
       "-o outputFile.ll ",
       "inputFile.ll");
 
+  std::cout << "rec: " << receivedCommandLine << std::endl;
+  std::cout << "exp: " << expectedCommandLine << std::endl;
+
```
phate, line 57: This should be removed, no?

phate (changes requested):

jlm/util/file.hpp
haved: Many places use it implicitly, so we agreed to not add explicit

phate (approved):

PR #747, opened 2025-01-13
Clean up API of lambda node before RVSDG move

caleridas:
Remove all non-orthogonal convenience member functions (i.e.
functions reaching from lambda::node into lambda::operation).
Redudant helpers may save typing, but generally cause churn
during decoupling.

Extract memory state modelling from lambda node: The modelling
is quite specific to how llvm-level transformations represent functions
etc. Everything related to that modelling should probably be
consolidated into a location (i.e. the "memory encoding bits")
rather than being spread out to the pieces they apply to.

Move argument attributes to lambda operation: This removes the last
bit of state from the lambda node which now really only contains graph
structural information.
    
Moving the attributes also has the effect of subjecting them to implicit
copy, removing the need to manually copy them in various places.


phate (approved):
Sorry that it took so long.

caleridas:
> Sorry that it took so long.

No worries, next part is just one rebase away :)

PR #760, opened 2025-01-18
Replace optimization interface with Transformation interface

phate:
This PR replaces the optimization interface with the Transformation interface. Specifically, this PR does the following:

1. Removes the optimization interface in the llvm sublibrary
2. Adds the Transformation interface to the rvsdg sublibrary.
3. Replaces all usages of the optimization interface with the Transformation interface

Ultimately, this lifts the concept of an RVSDG transformation to the rvsdg library, enabling other sublibraries, such as the HLS library, to tap into it in the future and set up a coherent transformation pipeline.

Moreover, this will enable to lift common transformations, such as dead node elimination, to the RVSDG library level. Ultimately, this will help to remove code duplication that is currently present in the code base, i.e., DNE clones in the LLVM and HLS sublibraries.


haved (approved):
Very nice

haved (approved):

PR #761, opened 2025-01-19
Changes needed to compile on macOS

sjalander:
Currently, tests/jlm/util/TestBijectiveMap.cpp causes a compilation error on macOS

In file included from tests/jlm/util/TestBijectiveMap.cpp:9:
In file included from ./jlm/util/BijectiveMap.hpp:11:
/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/unordered_map:680:17: fatal error: no viable overloaded '='
        __ref() = __v.__get_value();
        ~~~~~~~ ^ ~~~~~~~~~~~~~~~~~
/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/__hash_table:1779:47: note: in instantiation of member
      function 'std::__1::__hash_value_type<int, const std::__1::basic_string<char> >::operator=' requested here
                __cache->__upcast()->__value_ = *__first;


.github/workflows/macos.yml
```diff
@@ -0,0 +1,28 @@
+name: macos
```
sjalander, line 1: Seems like github charges for mac runners.
So, the question is if this should be included and disabled for now or simply leave it out.

sjalander:
I don't have clang-format-18 on the mac, so currently need to fetch the branch on Linux and perform it there.
Ideally, instead of the CI checking the format, it would be more convenient to have it perform the format and if there's a change then add a commit to the PR. I haven't looked into if this is possible, though.

.github/workflows/macos.yml
phate: I am not really willing to pay for it, as it is not that important enough.

scripts/build-all.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
```
phate, line 1: Why do we need this script? We already have a configure.sh script?

sjalander: It was a help while making all the changes necessary for getting it to compile on Mac. Both to keep changes in one place while not all the other scripts had been updated and to have a standard way of compiling everything.

I would prefer a script like this for the CI that can be run locally to check that everything get set up correctly. Instead of the way it is now, where a fault in the CI is debugged by pushing changes to a PR.

.github/workflows/macos.yml
sjalander: Removed

scripts/run-hls-test.sh
```diff
@@ -14,8 +14,19 @@ JLM_BIN_DIR=${JLM_ROOT_DIR}/build
...
+  PARALLEL_THREADS=`nproc`
+  MAKE_OPT="-O"
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+  PARALLEL_THREADS=`sysctl -n hw.ncpu`
+  MAKE_OPT="-O"
+else
+  echo "warning: Operating system not recognized." >&2
+  PARALLEL_THREADS=1
+  MAKE_OPT=""
+fi
```
phate, line 29: Maybe we should have a hard exit here.

scripts/build-mlir.sh
```diff
@@ -10,7 +10,19 @@ MLIR_BUILD=${JLM_ROOT_DIR}/build-mlir
...
 LLVM_VERSION=18
-LLVM_CONFIG_BIN=llvm-config-${LLVM_VERSION}
+
+# Set operating system specific configurations
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+  LLVM_CONFIG_BIN="llvm-config-"${LLVM_VERSION}
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+  LLVM_CONFIG_BIN="/usr/local/Cellar/llvm@"${LLVM_VERSION}"/"${LLVM_VERSION}"*/bin/llvm-config"
+  # Use the same MACOSX_DEPLOYMENT_TARGET as used for the LLVM and MLIR libraries
+  # to aviod warnings during linking
```
phate, line 20: type: avoid

scripts/build-circt.sh
```diff
@@ -12,7 +12,19 @@ CIRCT_INSTALL=${JLM_ROOT_DIR}/usr
...
 LLVM_VERSION=18
-LLVM_CONFIG_BIN=llvm-config-${LLVM_VERSION}
+
+# Set operating system specific configurations
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+  LLVM_CONFIG_BIN="llvm-config-"${LLVM_VERSION}
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+  LLVM_CONFIG_BIN="/usr/local/Cellar/llvm@"${LLVM_VERSION}"/"${LLVM_VERSION}"*/bin/llvm-config"
+  # Use the same MACOSX_DEPLOYMENT_TARGET as used for the LLVM and MLIR libraries
+  # to aviod warnings during linking
```
phate, line 22: typo: avoid

phate (changes requested):
Could we handle the code changes and the script changes in different PRs? The code changes are no problem to review and merge, but the script changes would need some more time.

scripts/build-mlir.sh
sjalander: Fixed

scripts/build-circt.sh
sjalander: Fixed

sjalander:
The code changes are not useful without the other changes.

scripts/run-hls-test.sh
sjalander: This is not used any longer.

sjalander:
@phate I've tried your suggestion about using configure.sh to set OS specific settings. To do this, I added the generation of a shell.configure, which is then sourced by the different shell scripts.

Since one now need to configure jlm before being able to build MLIR and CIRCT, the CI actions needed a slight modification such that the configure step appears before the build steps.

configure.sh
```diff
@@ -191,3 +192,30 @@ EOF
...
+  echo "export C_INCLUDE_PATH=`xcrun --show-sdk-path`/usr/include/" >> Makefile.config
+  echo "CPPFLAGS += -I`xcrun --show-sdk-path`/usr/include/" >> Makefile.config
+  echo "LDFLAGS += -L`xcrun --show-sdk-path`/usr/lib" >> Makefile.config
+
+  echo "export MACOSX_DEPLOYMENT_TARGET=`otool -l ${LLVM_CONFIG_BIN} | grep minos | awk '{print $2}'`" >> shell.config
+  PARALLEL_THREADS=`sysctl -n hw.ncpu`
+else
+  PARALLEL_THREADS=`nproc`
+fi
+echo "PARALLEL_THREADS=$PARALLEL_THREADS" >> shell.config
```
phate, line 221: No, this is not okay. You created a dependency between the configure command for Jlm and the scripts we have for benchmarks and building.

The stuff that is written to shell.config can be just put in a file in scripts/shell.config and then included from all the other scripts if you do not want to duplicate the code.

scripts/build-circt.sh
```diff
@@ -11,8 +11,13 @@ CIRCT_BUILD=${JLM_ROOT_DIR}/build-circt
...
 
-LLVM_VERSION=18
-LLVM_CONFIG_BIN=llvm-config-${LLVM_VERSION}
+# Include global shell configuration
+if [ -f ${JLM_ROOT_DIR}/shell.config ]; then
+	source ${JLM_ROOT_DIR}/shell.config
+else
+	echo "Please use configure.sh to configure jlm."
+	exit
+fi
```
phate, line 20: Once you have the shell.config file, you can also simplify the if-then-else to a simple source.

scripts/run-polybench.sh
```diff
@@ -50,10 +58,13 @@ done
...
+else
+	git -C ${BENCHMARK_DIR} fetch origin
 fi
 
 export PATH=${JLM_BIN_DIR}:${PATH}
 cd ${BENCHMARK_DIR}
 git checkout ${GIT_COMMIT}
 make clean
-make -j `nproc` -O ${BENCHMARK_RUN_TARGET}
+echo "make -j ${PARALLEL_THREADS} -O ${BENCHMARK_RUN_TARGET}"
```
phate, line 69: This is noise, no?

.github/actions/BuildJlm/action.yml
```diff
@@ -78,6 +70,14 @@ runs:
...
       shell: bash
 
+    - name: "Install HLS dialect dependencies"
+      if: ${{inputs.enable-hls == 'true'}}
+      uses: ./.github/actions/BuildCirct
+
+    - name: "Install MLIR dialect dependencies"
+      if: ${{inputs.enable-mlir == 'true'}}
+      uses: ./.github/actions/BuildMlirDialect
+
```
phate, line 80: All these can be reverted once you just have the shell.config as not auto-generated file.

.github/workflows/ClangTidy.yml
```diff
@@ -17,16 +17,17 @@ jobs:
       - name: "Install clang tidy"
         uses: ./.github/actions/InstallPackages
         with:
+          install-llvm: true
           install-clang-tidy: true
 
+      - name: "Configure jlm with HLS and MLIR enabled"
+        run: ./configure.sh --enable-mlir=${{ github.workspace }}/lib/mlir-rvsdg --enable-hls=${{ github.workspace }}/build-circt/circt
+
```
phate, line 25: same here.

phate (changes requested):

configure.sh
sjalander: @caleridas Do you have an opinion regarding using configure.sh to not only create a Makefile.config but also a shell.config. This config can then be used by the jlm specific scripts to build MLIR and CIRCT and to run test suites?

My thought was that this creates a single point for configuring everything that's relevant for building and testing jlm, which ensures that everything is set up correctly.

An alternative would be to create a separate configure script that creates a shell.config. This would result in that the same information would have to be given to both scripts, e.g., which llvm-config to use.

Or do you have another suggestion on how to handle this?

caleridas: I think what is happening here is a bit messy -- generally, neither LD_LIBRARY_PATH settings nor toolchain-specific directory selection should be determined by the build configuration script e but rather provided from the outside and injected.

So, normally you would be expected to call e.g.

```./configure --includedir=`xcrun --show-sdk-path`/usr/include/ ```

etc (if needed at all) rather than the build configuration scripts having knowledge where/how to find toolchains. Normally it would actually even be on the toolchain install itself to "know" the correct location of standard includes (e.g. g++/clang installed in a custom non-system location are normally set up to pick up "their" private includes from the correct places). As for LD_LIBRARY_PATH, that is similarly environment-specific, I am actually very puzzled that these are not added to the standard lib paths in your environment? This is IMHO how this "should" be handled on somewhat regular installs -- after all, if llvm-config is added to PATH why are the libs not added likewise to the environment? For custom installs I would suggest either having a site-specific env-setup script, or adding it to .bashrc if it does not get in the way.

Are you certain that the paths produced by `xcrun` are correct on _every_ OSX setup? Or is this matching _your_ specific setup and way of installing the rest of the dependencies? I suspect the answer is the latter, maybe the more appropriate approach would be to have a "wrapper" that calls the `configure` script with paths appropriate for your setup and then role of the `configure` script solely remains to put the info you provide about your setup into the correct places?

What you do is not out of question (particularly since I am totally unaffected and therefore cannot really complain), but I am wondering whether it is putting too many site-specific assumptions into the script, and at some point you are rather going to regret having site-specific assumptions that cannot easily be overridden anymore.

In general, when you _do_ infer variable values by calling external scripts, I would strongly suggest evaluating the results of that in the configure script itself, and then write the _resulting_ values into the configuration. This way a) the configuration is fixed and not further dependent on changing at a whim, and b) the result is transparent and intro-spectable. (One may add c) it does not cause calling the tools on each single build target, although probably nobody cares these days).

scripts/run-polybench.sh
sjalander: No, PARALLEL_THREADS is a new variable.

configure.sh
sjalander: I've applied these settings locally, such that the configure script deosn't have to be changed.

phate (approved):

PR #763, opened 2025-01-21
Generate cached version of MLIR and CIRCT for master

sjalander:
The cache is not shared between branches, resulting in CIRCT and MLIR being built each time a new PR is created.

This creates a cached version of MLIR and CIRCT for master, which is shared with derived branches.


.github/workflows/DefaultCache.yml
```diff
@@ -0,0 +1,20 @@
+name: DefaultCache
+
+on:
+  push:
```
sjalander, line 4: This should only run when master is updated, so it isn't checked until it is merged.
But I tried it with 'pull_request' to verify that everything builds.

phate:
@sjalander What is the idea here? You create a single cached instance of CIRCT and MLIR on every push to master that is then used by future PRs instead of them building it?

What happens if the version, i.e., commit, of CIRCT/MLIR changes within a PR?

sjalander:
@phate The build should only happen if it hasn't already been cached. So it should only happen once the commit hash of MLIR/CIRCT has been updated as this is used for the key to check the cache, or if the cache gets garbage collected by github.

The PR checks the cache with the key based on the commit hash. So, if CIRCT or MLIR gets updated, then the key will not match and miss in the cache. It will then be built for the PR and then a second time when merged with master. After this, all new PRs should use the cached version built in master.

phate (approved):

PR #764, opened 2025-01-22
Bundle Lllvm/Jlm type conversion in single class

phate:
Refactors the type conversion of Llvm/Jlm types and bundles the logic in a single class instead of distributing it all over the translation passes.


haved (approved):
Very nice to have it all in one place

PR #765, opened 2025-01-22
Bugfix in HLS view and added unit test

sjalander:


jlm/hls/Makefile.sub
```diff
@@ -88,6 +88,7 @@ libhls_TESTS += \
 	tests/jlm/hls/backend/rvsdg2rhls/TestTheta \
 	tests/jlm/hls/backend/rvsdg2rhls/UnusedStateRemovalTests \
 	tests/jlm/hls/backend/rvsdg2rhls/test-loop-passthrough \
+	tests/jlm/hls/util/TestView \
```
phate, line 91: Please call the file `ViewTests`. If we prefix everything with `Test`, then alphabetical sorting is defeated.

tests/jlm/hls/util/TestView.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+  using namespace jlm;
+  using namespace jlm::hls;
+  using namespace jlm::llvm;
+
+  // Arrange
+  auto b32 = rvsdg::bittype::Create(32);
+  auto ft = rvsdg::FunctionType::Create({}, { b32 });
+
+  RvsdgModule rm(util::filepath(""), "", "");
```
phate, line 28: I believe that there is no need to create an RvsdgModule, as it is nowhere used. You could also just create a graph: `rvsdg::Graph graph`

tests/jlm/hls/util/TestView.cpp
```diff
@@ -0,0 +1,101 @@
...
+{
+  using namespace jlm;
+  using namespace jlm::hls;
+  using namespace jlm::llvm;
+
+  // Arrange
+  auto b32 = rvsdg::bittype::Create(32);
+  auto ft = rvsdg::FunctionType::Create({ b32, b32, b32 }, { b32, b32, b32 });
+
+  RvsdgModule rm(util::filepath(""), "", "");
```
phate, line 66: Same here.

phate (changes requested):
I would like to fix the file name, please.

jlm/hls/Makefile.sub
sjalander: Renamed

tests/jlm/hls/util/TestView.cpp
sjalander: Changed to graph

tests/jlm/hls/util/TestView.cpp
sjalander: Changed to graph

phate (approved):

PR #770, opened 2025-01-25
Simplify translation of integer operations in LLVM to JLM pass

phate:


haved (approved):
This increases readability while also reducing lines of code. Very nice

PR #780, opened 2025-01-28
Move and rename lambda node

caleridas:
Move lambda node to rvsdg, and rename it to LambdaNode / LambdaOperation. Provide "baseline" LambdaOperation and derived llvm-specific features. The name LambdaOperation is provisional and may later change to better capture its intent.

### DIFF
--- a/jlm/llvm/backend/rvsdg2jlm/rvsdg2jlm.cpp
+++ b/jlm/llvm/backend/rvsdg2jlm/rvsdg2jlm.cpp
@@ -118,7 +118,7 @@ convert_region(rvsdg::Region & region, context & ctx)
 }
 
 static inline std::unique_ptr<llvm::cfg>
-create_cfg(const lambda::node & lambda, context & ctx)
+create_cfg(const rvsdg::LambdaNode & lambda, context & ctx)
 {
   JLM_ASSERT(ctx.lpbb() == nullptr);
   std::unique_ptr<llvm::cfg> cfg(new llvm::cfg(ctx.module()));
@@ -134,7 +134,8 @@ create_cfg(const lambda::node & lambda, context & ctx)
     auto argument = llvm::argument::create(
         name,
         fctarg->Type(),
-        lambda.GetOperation().GetArgumentAttributes(fctarg->index()));
+        dynamic_cast<llvm::LlvmLambdaOperation &>(lambda.GetOperation())
+            .GetArgumentAttributes(fctarg->index()));
     auto v = cfg->entry()->append_argument(std::move(argument));
     ctx.insert(fctarg, v);
   }
@@ -410,12 +411,12 @@ convert_theta_node(const rvsdg::Node & node, context & ctx)
 static inline void
 convert_lambda_node(const rvsdg::Node & node, context & ctx)
 {
-  JLM_ASSERT(is<lambda::operation>(&node));
-  auto lambda = static_cast<const lambda::node *>(&node);
+  JLM_ASSERT(is<llvm::LlvmLambdaOperation>(&node));
+  auto lambda = static_cast<const rvsdg::LambdaNode *>(&node);
   auto & module = ctx.module();
   auto & clg = module.ipgraph();
 
-  const auto & op = lambda->GetOperation();
+  const auto & op = dynamic_cast<llvm::LlvmLambdaOperation &>(lambda->GetOperation());
   auto f = function_node::create(clg, op.name(), op.Type(), op.linkage(), op.attributes());
   auto v = module.create_variable(f);
 
@@ -445,9 +446,9 @@ convert_phi_node(const rvsdg::Node & node, context & ctx)
     JLM_ASSERT(subregion->argument(n)->input() == nullptr);
     auto node = rvsdg::output::GetNode(*subregion->result(n)->origin());
 
-    if (auto lambda = dynamic_cast<const lambda::node *>(node))
+    if (auto lambda = dynamic_cast<const rvsdg::LambdaNode *>(node))
     {
-      const auto & op = lambda->GetOperation();
+      const auto & op = dynamic_cast<llvm::LlvmLambdaOperation &>(lambda->GetOperation());
       auto f = function_node::create(ipg, op.name(), op.Type(), op.linkage(), op.attributes());
       ctx.insert(subregion->argument(n), module.create_variable(f));
     }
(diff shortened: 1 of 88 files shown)


jlm/llvm/ir/operators/lambda.hpp
```diff
@@ -21,46 +22,21 @@
...
 
-namespace lambda
-{
-
 /** \brief Lambda operation
  *
  * A lambda operation determines a lambda's name and \ref rvsdg::FunctionType "function type".
  */
-class operation final : public rvsdg::StructuralOperation
+class LlvmLambdaOperation final : public rvsdg::LambdaOperation
```
phate, line 29: The namespace `jlm::llvm` clarifies it that this is the LLVM lambda, no? It could also just be named `LambdaOperation`.

jlm/llvm/backend/rvsdg2jlm/rvsdg2jlm.cpp
```diff
@@ -410,12 +411,12 @@ convert_theta_node(const rvsdg::Node & node, context & ctx)
...
 {
-  JLM_ASSERT(is<lambda::operation>(&node));
-  auto lambda = static_cast<const lambda::node *>(&node);
+  JLM_ASSERT(is<llvm::LlvmLambdaOperation>(&node));
+  auto lambda = static_cast<const rvsdg::LambdaNode *>(&node);
   auto & module = ctx.module();
   auto & clg = module.ipgraph();
 
-  const auto & op = lambda->GetOperation();
+  const auto & op = dynamic_cast<llvm::LlvmLambdaOperation &>(lambda->GetOperation());
```
phate, line 419: This is now what I do not like. We get these casts everywhere where we need the more specialized form. Any change we could introduce a templated `GetOperation<>()` method that tries to cast it down to the concrete LambdaOperation type?

jlm/llvm/backend/rvsdg2jlm/rvsdg2jlm.cpp
```diff
@@ -515,21 +516,20 @@ convert_node(const rvsdg::Node & node, context & ctx)
...
-                { typeid(phi::operation), convert_phi_node },
-                { typeid(delta::operation), convert_delta_node } });
+          map({ { typeid(rvsdg::LambdaNode), convert_lambda_node },
+                { std::type_index(typeid(rvsdg::GammaNode)), convert_gamma_node },
+                { std::type_index(typeid(rvsdg::ThetaNode)), convert_theta_node },
+                { typeid(phi::node), convert_phi_node },
+                { typeid(delta::node), convert_delta_node } });
 
-  if (dynamic_cast<const rvsdg::SimpleOperation *>(&node.GetOperation()))
+  if (dynamic_cast<const rvsdg::SimpleNode *>(&node))
```
phate, line 525: Since you touch this line anyway: `is<>`

jlm/rvsdg/lambda.cpp
```diff
@@ -0,0 +1,212 @@
...
+LambdaOperation::~LambdaOperation() = default;
+
+LambdaOperation::LambdaOperation(std::shared_ptr<const FunctionType> type)
+    : type_(std::move(type))
+{}
+
+std::string
+LambdaOperation::debug_string() const
+{
+  return util::strfmt("LAMBDA[", Type()->debug_string(), "]");
```
phate, line 21: Since you touch it: `Lambda[]`

jlm/rvsdg/lambda.cpp
```diff
@@ -0,0 +1,212 @@
...
+  return op && op->type() == type();
+}
+
+std::unique_ptr<rvsdg::Operation>
+LambdaOperation::copy() const
+{
+  return std::make_unique<LambdaOperation>(*this);
+}
+
+/* lambda node class */
```
phate, line 37: Can be removed

jlm/rvsdg/lambda.hpp
```diff
@@ -0,0 +1,302 @@
...
+
+#include <optional>
+#include <utility>
+
+namespace jlm::rvsdg
+{
+
+/** \brief Lambda operation
+ *
+ * A lambda operation determines a lambda's name and \ref FunctionType "function type".
```
phate, line 24: Comment is incorrect. The name part is not true any longer.

jlm/rvsdg/lambda.hpp
```diff
@@ -0,0 +1,302 @@
...
+
+  LambdaOperation(const LambdaOperation & other) = default;
+
+  LambdaOperation(LambdaOperation && other) noexcept = default;
+
+  LambdaOperation &
+  operator=(const LambdaOperation & other) = default;
+
+  LambdaOperation &
+  operator=(LambdaOperation && other) noexcept = default;
```
phate, line 41: Could be removed, no? Or is this not possible as we do not have an empty default constructor?

phate (changes requested):
This biggest problem I have with it is that we get these casts to the LlvmLambdaOperation everywhere. Any chance we could do something about this? See my review comment. The rest is all minor stuff.

jlm/llvm/ir/operators/lambda.hpp
caleridas: Naming it "LambdaNode" causes name resolution issues when both jlm::llvm and jlm::rvsdg are in scope. It also causes issues with moving code between modules (hence namespaces), as now suddenly instead of a compilation  error you get changed semantics. Lastly, it is also very ambiguous when reading the code unless you are using an IDE that constantly clarifies for every visible symbol where it resolves.

I was orginally inclined to name it more like "LlvmFunctionAttributes" or so (that's what it effectively is), rather going the direction of making it more ambiguous.

jlm/llvm/backend/rvsdg2jlm/rvsdg2jlm.cpp
caleridas: No this does not work, is<> checks for the operation while I want code to really rather check for node type primarily.

Generalizing is<> to accept either might be possible but requires interesting C++ "concepts" which I would rather not introduce.

I would indeed generally like to remove the duality of dispatching on "op" vs. "node", and cleanly settle that "op" is only sensible for things that are SimpleNodes (you still need to cast the node to something specific for doing anything sensible with structural nodes, so there is no point in ever starting with op there).

phate (approved):

PR #781, opened 2025-01-30
MLIR roundtrip of Alloca, Load and Store operations (Non-volatile)

halvorlinder:
This PR adds roundtripping of the mentioned nodes together with tests for this functionality. 
Depends changes to the [dialect](https://github.com/EECS-NTNU/mlir_rvsdg/pull/4/commits/51889684abf9f64dac8d051ee6a71a0a78827860).

### DIFF
--- a/jlm/mlir/frontend/MlirToJlmConverter.cpp
+++ b/jlm/mlir/frontend/MlirToJlmConverter.cpp
@@ -16,6 +16,11 @@
 
 #include <jlm/llvm/ir/operators/operators.hpp>
 
+#include <jlm/llvm/ir/operators/alloca.hpp>
+#include <jlm/llvm/ir/operators/Load.hpp>
+#include <jlm/llvm/ir/operators/MemoryStateOperations.hpp>
+#include <jlm/llvm/ir/operators/Store.hpp>
+
 namespace jlm::mlir
 {
 
@@ -316,6 +321,55 @@ MlirToJlmConverter::ConvertOperation(
     return rvsdg::output::GetNode(*jlmUndefOutput);
   }
 
+  // Memory operations
+
+  else if (auto AllocaOp = ::mlir::dyn_cast<::mlir::jlm::Alloca>(&mlirOperation))
+  {
+    auto outputType = AllocaOp.getValueType();
+    std::shared_ptr<jlm::rvsdg::Type> jlmType = ConvertType(outputType);
+    auto jlmValueType = std::dynamic_pointer_cast<const rvsdg::ValueType>(jlmType);
+
+    auto jlmBitType = dynamic_cast<const jlm::rvsdg::bittype *>(&inputs[0]->type());
+    auto bitTypePrt = std::make_shared<const jlm::rvsdg::bittype>(jlmBitType->nbits());
+
+    auto allocaOp = jlm::llvm::alloca_op(jlmValueType, bitTypePrt, AllocaOp.getAlignment());
+
+    auto operands = std::vector(inputs.begin(), inputs.end());
+
+    return &rvsdg::SimpleNode::Create(rvsdgRegion, allocaOp, operands);
+  }
+  else if (auto MemstateMergeOp = ::mlir::dyn_cast<::mlir::rvsdg::MemStateMerge>(&mlirOperation))
+  {
+    auto operands = std::vector(inputs.begin(), inputs.end());
+    auto memoryStateMergeOutput = jlm::llvm::MemoryStateMergeOperation::Create(operands);
+    return rvsdg::output::GetNode(*memoryStateMergeOutput);
+  }
+  else if (auto StoreOp = ::mlir::dyn_cast<::mlir::jlm::Store>(&mlirOperation))
+  {
+    auto address = inputs[0];
+    auto value = inputs[1];
+    auto memoryStateInputs = std::vector(std::next(std::next(inputs.begin())), inputs.end());
+    auto & storeNode = jlm::llvm::StoreNonVolatileNode::CreateNode(
+        *address,
+        *value,
+        memoryStateInputs,
+        StoreOp.getAlignment());
+    return &storeNode;
+  }
+  else if (auto LoadOp = ::mlir::dyn_cast<::mlir::jlm::Load>(&mlirOperation))
+  {
+    auto address = inputs[0];
+    auto memoryStateInputs = std::vector(std::next(inputs.begin()), inputs.end());
+    auto outputType = LoadOp.getOutput().getType();
+    std::shared_ptr<jlm::rvsdg::Type> jlmType = ConvertType(outputType);
+    auto jlmValueType = std::dynamic_pointer_cast<const rvsdg::ValueType>(jlmType);
+    auto & loadNode = jlm::llvm::LoadNonVolatileNode::CreateNode(
+        *address,
+        memoryStateInputs,
+        jlmValueType,
+        LoadOp.getAlignment());
+    return &loadNode;
+  }
   // * region Structural nodes **
   else if (auto MlirCtrlConst = ::mlir::dyn_cast<::mlir::rvsdg::ConstantCtrl>(&mlirOperation))
   {
@@ -491,6 +545,10 @@ MlirToJlmConverter::ConvertType(::mlir::Type & type)
   else if (::mlir::isa<::mlir::rvsdg::IOStateEdgeType>(type))
   {
     return std::make_unique<llvm::IOStateType>();
+  }
+  else if (::mlir::isa<::mlir::rvsdg::RVSDGPointerType>(type))
+  {
+    return std::make_unique<llvm::PointerType>();
   }
   else
   {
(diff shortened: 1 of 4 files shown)


jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -316,6 +321,55 @@ MlirToJlmConverter::ConvertOperation(
...
+  // Memory operations
+
+  else if (auto AllocaOp = ::mlir::dyn_cast<::mlir::jlm::Alloca>(&mlirOperation))
+  {
+    auto outputType = AllocaOp.getValueType();
+    std::shared_ptr<jlm::rvsdg::Type> jlmType = ConvertType(outputType);
+    auto jlmValueType = std::dynamic_pointer_cast<const rvsdg::ValueType>(jlmType);
+
+    auto jlmBitType = dynamic_cast<const jlm::rvsdg::bittype *>(&inputs[0]->type());
+    auto bitTypePrt = std::make_shared<const jlm::rvsdg::bittype>(jlmBitType->nbits());
```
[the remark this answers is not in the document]
haved: If you use `Type()` (with uppercase T) on the line above, you get a `shared_ptr` to the type directly. You can use `std::dynamic_pointer_cast` on it.

I would add some ASSERTS, or actually even `throw jlm::util::errorr` to the dynamic casts if they fail. The MLIR is user input, so we can not really trust it.

haved (commented):
Some suggestions, I haven't looked at the rest yet

haved:
@phate @caleridas In a frontend, where the incoming data is user provided, do we want to use `throw jlm::util::___error`, asserts or unreachable? I think asserts are problematic as they may be compiled out, and should not be used to validate user-input.

Looking at the llvm frontend, there are mostly asserts, but they also mostly check our own invariants, and not the correctness of user input. We rely on the LLVM IR being well formed, which I think makes sense, as LLVM already has code for validating their IR, so there is less need for us to validate things.

For MLIR, there is only as much validation as we provide ourselves (in our dialect). I think the MLIR frontend needs to be a bit more careful about its assumptions, and use exceptions (or `JLM_UNREACHABLE`) when it assumes that the MLIR looks a certain way. I do not think lots of checks will be necessary, but when we do need to validate user-input, it is nice to agree on how we do it.

phate:
@haved In case of user inputs, it should be exceptions.

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -316,6 +321,55 @@ MlirToJlmConverter::ConvertOperation(
...
+  {
+    auto operands = std::vector(inputs.begin(), inputs.end());
+    auto memoryStateMergeOutput = jlm::llvm::MemoryStateMergeOperation::Create(operands);
+    return rvsdg::output::GetNode(*memoryStateMergeOutput);
+  }
+  else if (auto StoreOp = ::mlir::dyn_cast<::mlir::jlm::Store>(&mlirOperation))
+  {
+    auto address = inputs[0];
+    auto value = inputs[1];
+    auto memoryStateInputs = std::vector(std::next(std::next(inputs.begin())), inputs.end());
```
phate, line 351: `std::next(inputs.begin(), 2)`

tests/jlm/mlir/TestJlmToMlirToJlm.cpp
```diff
@@ -67,4 +69,300 @@ TestUndef()
...
+
+        auto valueType = mlir::cast<mlir::IntegerType>(mlirAllocaOp.getValueType());
+        assert(valueType);
+        assert(valueType.getWidth() == 64);
+        foundAlloca = true;
+      }
+    }
+    if (!foundAlloca)
+    {
+      return 1;
```
phate, line 121: `assert(false)`

tests/jlm/mlir/TestJlmToMlirToJlm.cpp
```diff
@@ -67,4 +69,300 @@ TestUndef()
...
+
+          assert(jlm::rvsdg::is<PointerType>(allocaOp->result(0)));
+          assert(jlm::rvsdg::is<jlm::llvm::MemoryStateType>(allocaOp->result(1)));
+
+          foundAlloca = true;
+        }
+      }
+      if (!foundAlloca)
+      {
+        return 1;
```
phate, line 165: `assert(false)`

tests/jlm/mlir/TestJlmToMlirToJlm.cpp
```diff
@@ -67,4 +69,300 @@ TestUndef()
...
+    std::cout << "Validate MLIR" << std::endl;
+    auto & omegaRegion = omega.getRegion();
+    auto & omegaBlock = omegaRegion.front();
+    auto & mlirLambda = omegaBlock.front();
+    auto & mlirLambdaRegion = mlirLambda.getRegion(0);
+    auto & mlirLambdaBlock = mlirLambdaRegion.front();
+    auto & mlirOp = mlirLambdaBlock.front();
+
+    if (!mlir::isa<mlir::jlm::Load>(mlirOp))
+      return 1;
```
phate, line 218: `assert(mlir::isa<mlir::jlm::Load>(mlirOp))`

tests/jlm/mlir/TestJlmToMlirToJlm.cpp
```diff
@@ -67,4 +69,300 @@ TestUndef()
...
+    std::cout << "Validate MLIR" << std::endl;
+    auto & omegaRegion = omega.getRegion();
+    auto & omegaBlock = omegaRegion.front();
+    auto & mlirLambda = omegaBlock.front();
+    auto & mlirLambdaRegion = mlirLambda.getRegion(0);
+    auto & mlirLambdaBlock = mlirLambdaRegion.front();
+    auto & mlirOp = mlirLambdaBlock.front();
+
+    if (!mlir::isa<mlir::jlm::Store>(mlirOp))
+      return 1;
```
phate, line 315: same here

tests/jlm/mlir/TestJlmToMlirToJlm.cpp
```diff
@@ -67,4 +69,300 @@ TestUndef()
...
+      assert(inputBitType->nbits() == 32);
+    }
+  }
+  return 0;
+}
+
 JLM_UNIT_TEST_REGISTER("jlm/mlir/TestMlirUndefGen", TestUndef)
+JLM_UNIT_TEST_REGISTER("jlm/mlir/TestMlirAllocaGen", TestAlloca)
+JLM_UNIT_TEST_REGISTER("jlm/mlir/TestMlirLoadGen", TestLoad)
+JLM_UNIT_TEST_REGISTER("jlm/mlir/TestMlirStoreGen", TestStore)
```
phate, line 368: It would be nice to put them directly under the respective test

phate (dismissed):

caleridas:
> @phate @caleridas In a frontend, where the incoming data is user provided, do we want to use `throw jlm::util::___error`, asserts or unreachable? I think asserts are problematic as they may be compiled out, and should not be used to validate user-input.

conventions that I can generally recommend:
- if it is error preventable by coding correctly (i.e. caller ensuring preconditions of called functions are met, maintaining data structure invariants, "not mere coding bugs"), then it should be std::logic_error or anything derived from it (std::bad_cast actually also belongs into the same category)
- if it is due to external conditions, it should be std::runtime_error (or anything derived from it); that includes validating user input which could legitimately throw std::runtime_error
- std::runtime_error may be caught, std::logic_error should _never_ be caught; there may be exceptions when a "driver" program of a library uses said library and that library makes gratuitious use std::logic_error for non-preventable things (seen that)

In general, _libraries_ should never ever issue std::abort / std::unreachable and sparingly make use of std::logic_error -- all of these should be reserved purely for own internal inconsistencies, and std::abort / std::unreachable really reserved for cases when there is no path to recover (e.g. because internal data structures variants are destroyed). Otherwise, this just enforces "whole program behavior" from within a library where the library author is generally never in a good position to "know" the scope of the program and generally should defer judgement to the main program author who can decide what to do (just quit, quit and log, try to recover, launch crashdump helper to display helpful message for user and/or send backtrace to developers...).

> 
> Looking at the llvm frontend, there are mostly asserts, but they also mostly check our own invariants, and not the correctness of user input. We rely on the LLVM IR being well formed, which I think makes sense, as LLVM already has code for validating their IR, so there is less need for us to validate things.
> 
> For MLIR, there is only as much validation as we provide ourselves (in our dialect). I think the MLIR frontend needs to be a bit more careful about its assumptions, and use exceptions (or `JLM_UNREACHABLE`) when it assumes that the MLIR looks a certain way. I do not think lots of checks will be necessary, but when we do need to validate user-input, it is nice to agree on how we do it.

haved (dismissed):

sjalander:
@halvorlinder There are tests that fails. Is there a need to update the MLIR dialect for it to work?

sjalander (approved):

PR #783, opened 2025-01-30
Add and use UnionWithAndClear and IntersectWithAndClear to HashSet

haved:


jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
```diff
@@ -1649,19 +1648,17 @@ PointerObjectConstraintSet::RunWorklistSolver(WorklistStatistics & statistics)
     const auto nonRoot = root == aRoot ? bRoot : aRoot;
 
     // Move constraints owned by the non-root to the root
-    supersetEdges[root].UnionWith(supersetEdges[nonRoot]);
-    supersetEdges[nonRoot].Clear();
+    supersetEdges[root].UnionWithAndClear(supersetEdges[nonRoot]);
+
     // Try to avoid self-edges, but indirect self-edges can still exist
     supersetEdges[root].Remove(root);
+    supersetEdges[root].Remove(nonRoot);
```
phate, line 1655: Why does this all in a sudden appear here? It was not there before (or have I missed it)?

phate (approved):
Looks good, just a minor question from my side.

jlm/llvm/opt/alias-analyses/PointerObjectSet.cpp
haved: Good catch, the change is not technically related to union+clear.

This change is a small optimization to the Andersen solver, which is unlikely to have any big impact, but feels nice to have to avoid extra work.

PR #788, opened 2025-02-04
Add LLVM-specific integer arithmetic operations

phate:
This PR replaces the old bitstring operations with new LLVM specific integer arithmetic operations. The implementations have currently still the following limitations:

1. The operations still use the bistring type. The type will potentially later replaced by an LLVM specific type.
2. No operation specific reductions have been implemented yet. This will also happen in a follow-up PR.


phate:
There is still quite a bit repetition in the code, but I would like you to look over that for now. We can refactor that similar to the bitstring operations once the dust has settled.

jlm/llvm/ir/operators/IntegerOperations.hpp
```diff
@@ -0,0 +1,472 @@
...
+
+#include <jlm/rvsdg/binary.hpp>
+#include <jlm/rvsdg/bitstring/type.hpp>
+
+namespace jlm::llvm
+{
+
+/**
+ * This operation is equivalent to LLVM's 'add' instruction.
+ */
```
phate, line 17: FIXME: Update the documentation as all the operations are only equivalent to the integer part of the instructions.

jlm/llvm/ir/operators/IntegerOperations.cpp
```diff
@@ -0,0 +1,572 @@
...
+
+IntegerBinaryOperation::~IntegerBinaryOperation() = default;
+
+IntegerAddOperation::~IntegerAddOperation() noexcept = default;
+
+bool
+IntegerAddOperation::operator==(const Operation & other) const noexcept
+{
+  const auto addOperation = dynamic_cast<const IntegerAddOperation *>(&other);
+  return addOperation && addOperation->result(0) == result(0);
```
haved, line 19: Could all the references to `result(0)` use the `Type()` method instead?

haved (commented):
The interface looks nice, with some duplication of course.

I have one question about the use of `result(0)`.

In order to reduce duplication for some of the methods, maybe use of an additional helper superclass based on a curiously recurring template could be nice?

jlm/llvm/ir/operators/IntegerOperations.cpp
phate: Fixed.

phate:
@haved I would not like to reduce the duplication yet, before we have not modeled all the semantics of the corresponding LLVM instructions. Once we have all the semantics in there, we can start to simplify. I hope that is okay for you.

haved (approved):

PR #790, opened 2025-02-10
add HashSet::DifferenceWith method

haved:
I had a use for this method in the Andersen-work, and thought I would backport it as I am likely to want to use it again.

With this operation, there are two possible implementations. I try to pick the one that should be the fastest, but I have not done benchmarking to figure out exactly where one implementation becomes faster than the other.

### DIFF
--- a/jlm/util/HashSet.hpp
+++ b/jlm/util/HashSet.hpp
@@ -294,6 +294,35 @@ public:
   }
 
   /**
+   * Modifies this HashSet object by removing any elements that are present in \p other.
+   *
+   * @param other the HashSet used as the negative side of the set difference.
+   */
+  void
+  DifferenceWith(const HashSet<ItemType> & other)
+  {
+    // If this HashSet is smaller, loop over it and remove elements in other.
+    // If other is smaller, loop over it and remove elements from this.
+
+    if (Size() <= other.Size())
+    {
+      // This branch also handles the unlikely case where this and other are the same set.
+
+      auto inOther = [&](const ItemType item)
+      {
+        return other.Contains(item);
+      };
+
+      RemoveWhere(inOther);
+    }
+    else
+    {
+      for (auto & item : other.Set_)
+        Remove(item);
+    }
+  }
+
+  /**
    * Removes the specified item from a HashSet object.
    *
    * @param item The item to remove.
--- a/jlm/util/HashSet.hpp
+++ b/jlm/util/HashSet.hpp
@@ -268,7 +268,7 @@ public:
   void
   IntersectWith(const HashSet<ItemType> & other)
   {
-    auto isContained = [&](const ItemType item)
+    auto isContained = [&](const ItemType & item)
     {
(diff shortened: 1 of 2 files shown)


jlm/util/HashSet.hpp
```diff
@@ -293,6 +293,35 @@ class HashSet
...
+  DifferenceWith(const HashSet<ItemType> & other)
+  {
+    // If this HashSet is smaller, loop over it and remove elements in other.
+    // If other is smaller, loop over it and remove elements from this.
+
+    if (Size() <= other.Size())
+    {
+      // This branch also handles the unlikely case where this and other are the same set.
+
+      auto inOther = [&](const ItemType & item)
```
phate, line 311: nitpick: maybe `isInOther()`

tests/jlm/util/TestHashSet.cpp
```diff
@@ -180,3 +180,41 @@ TestIntersectWith()
...
+  HashSet<int> set12({ 1, 2 });
+  HashSet<int> set123({ 1, 2, 3 });
+  HashSet<int> set45({ 4, 5 });
+
+  set123.DifferenceWith(set12); // {1, 2, 3} - {1, 2}
+  assert(set123.Size() == 1);
+  assert(set123.Contains(3));
+
+  // set12 was not touched
+  assert(set12.Size() == 2);
```
phate, line 198: The  better way to check this is to make a copy of set12 and then equal the copy to the set12 after the operation.

jlm/util/HashSet.hpp
```diff
@@ -293,6 +293,35 @@ class HashSet
...
+      {
+        return other.Contains(item);
+      };
+
+      RemoveWhere(inOther);
+    }
+    else
+    {
+      for (auto & item : other.Set_)
+        Remove(item);
```
phate, line 321: Could this not be: `Set_.erase(other.Set_.begin(), other.Set_.end())`. It might be a bit faster than iterating over all the elements individually.

jlm/util/HashSet.hpp
```diff
@@ -293,6 +293,35 @@ class HashSet
...
+    if (Size() <= other.Size())
+    {
+      // This branch also handles the unlikely case where this and other are the same set.
+
+      auto inOther = [&](const ItemType & item)
+      {
+        return other.Contains(item);
+      };
+
+      RemoveWhere(inOther);
```
phate, line 316: The same here, no? See comment below.

phate (approved):

jlm/util/HashSet.hpp
haved: No, the documentation for `erase(first, last)` specifics that it must be iterators within the same set.

phate: Meh....

haved:
@phate I fixed your comments, where I could

phate (approved):

PR #794, opened 2025-02-12
Fix incorrect cache key of MLIR Dialect built in CI

haved:
Takes a lot of inspiration from the existing action `BuildCirct/action.yml`

Also changes the installation directory from `JLM_ROOT/lib/mlir-rvsdg` into `JLM_ROOT/build-mlir/usr`, which is closer to what CIRCT uses (`JLM_ROOT/build-circt/circt`).


sjalander (approved):

haved:
@sjalander I need a new review as I had forgotten to update the path in the clang tidy workflow. (Giving the scripts coherent defaults and letting them use them could be nice)

sjalander (approved):

PR #795, opened 2025-02-13
Add support for IOBarrier [AndersenAgnostic][SteensgaardAgnostic]

### LINKED ISSUE #750: How should we model the side effects of modulo and division operations?
The issues with #526 are caused by the following situation:
```c
#include <stdlib.h>

int x;

void foo() {
    exit(0);
}

int main() {
    int a = x;
    if (a == 0)
        foo();
    return 6 % a;
}
```

Using a build of `jlm-opt` from  and executing the following commands:
```sh
clang -S -emit-llvm example.c -o example.ll -Xclang -disable-O0-optnone
opt -S --passes=mem2reg example.ll -o example-m2r.ll
jlm-opt example-m2r.ll -o example-jlm.ll
clang example-jlm.ll -o example.out
./example.out
```
The program terminates with SIGFPE (Floating Point Exception), as it attempts to do modulo by 0.

For reference, the definition of `main` in `example-m2r.ll` is 
```llvm
  %1 = load i32, ptr @x, align 4
  %2 = icmp eq i32 %1, 0
  br i1 %2, label %3, label %4

3:                                                ; preds = %0
  call void @foo()
  br label %4

4:                                                ; preds = %3, %0
  %5 = srem i32 6, %1
  ret i32 %5
}
```
We need to `exit` via the `foo` function to prevent LLVM from adding `unreachable` right below the call, but we could just as well imagine `foo` being an external function that somehow calls `exit`. 

The RVSDG produced by jlm-opt ends up looking like
image

As you can see, the modulo operation `BitSMod32` ends up neither higher nor lower in the DAG than the `gamma` containing the call. When the program is converted back into LLVM IR, it (more or less randomly) ends up like so:

```llvm
define i32 @main() #0 {
bb0:
  %0 = load i32, ptr @x, align 4
  %1 = icmp eq i32 %0, 0
  %2 = srem i32 6, %0
  br i1 %1, label %bb2, label %bb1

bb1:                                              ; preds = %bb2, %bb0
  ret i32 %2

bb2:                                              ; preds = %bb0
  call void @foo()
  br label %bb1
}
```

This means the modulo by 0, which normally would never occur, happens before the terminating call to `foo()`.

We somehow need to model the possible side effect of modulo and division by 0 sending a signal.

This raises a few questions:
 - Should we use the I/O state edge to enforce the original sequence of operations? A signal is I/O in a sense.
 - What about other things that raise signals? Dereferencing a NULL pointer typically raises a signal, but is at the same time undefined behavior, so it does not have to.
   - What are the rules for re-ordering memory operations around function calls? What do other compilers do?
   - Is division by 0 "undefined behavior" or is it mearly implementation defined?

@caleridas @phate @sjalander

phate:
This PR introduces the IOBarrier operation and adds support for it to the LLVM and HLS dialect. The IOBarrier operation is necessary to sequentialize operations with potentially undefined behavior, such as division and modulo by zero as well as load and store operations from null pointers, to ensure correct program semantics. For example, the following code requires a IOBarrier to avoid unwanted undefined behavior:

```
int f(int x, int y)
{
   opaque(); // calls internally exit(0)
   return x / y;
}
```
The program is a valid C program under all circumstances, even if y equals zero. In order to capture this semantic, it is necessary to sequentialize the division operation below the call to opaque in the RVSDG. This is accomplished using the IOBarrier operation. If it can be proven that an operation can never exhibit undefined behavior, then the respective IOBarrier can be removed.

Close #750

### DIFF
--- a/jlm/hls/Makefile.sub
+++ b/jlm/hls/Makefile.sub
@@ -37,6 +37,7 @@ libhls_SOURCES = \
     jlm/hls/ir/hls.cpp \
     \
     jlm/hls/opt/cne.cpp \
+    jlm/hls/opt/IOBarrierRemoval.cpp \
     \
     jlm/hls/util/view.cpp \
 
@@ -76,6 +77,7 @@ libhls_HEADERS = \
 	jlm/hls/ir/hls.hpp \
 	\
 	jlm/hls/opt/cne.hpp \
+	jlm/hls/opt/IOBarrierRemoval.hpp \
 	\
 	jlm/hls/util/view.hpp \
 
@@ -88,6 +90,7 @@ libhls_TESTS += \
 	tests/jlm/hls/backend/rvsdg2rhls/TestTheta \
 	tests/jlm/hls/backend/rvsdg2rhls/UnusedStateRemovalTests \
 	tests/jlm/hls/backend/rvsdg2rhls/test-loop-passthrough \
+	tests/jlm/hls/opt/IOBarrierRemovalTests \
 	tests/jlm/hls/util/ViewTests \
 
 libhls_TEST_LIBS += \
--- a/jlm/hls/backend/rvsdg2rhls/rvsdg2rhls.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/rvsdg2rhls.cpp
@@ -47,6 +47,7 @@
 #include <llvm/Support/raw_ostream.h>
 #include <llvm/Support/SourceMgr.h>
 
+#include <jlm/hls/opt/IOBarrierRemoval.hpp>
 #include <regex>
 
 namespace jlm::hls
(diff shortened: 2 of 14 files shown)


haved (approved):
I think this implementation looks very clean
 As you say it may need some extra handling in optimizations, and I know it needs to be added as a pass-through in the Alias Analysis.

The current implementation where it depends on the latest `ctx` iostate works, but keep in mind it will then also sequentialize after volatile loads and stores, which we can then loosen up later, as they never halt (I'm 99% sure)

phate:
@haved @sjalander @caleridas It is ready for review now.

phate:
This should also unblock PR #526.

haved (approved):
Looks very nice, the Andersen implementation is perfect

PR #798, opened 2025-02-15
Rename assignment_op class to AssignmentOperation

phate:


haved (approved):
Looks good, though I realize I have no idea what an assignment operation is. Is it a store?

phate:
@haved : It is an assignment, i.e., x = y. It is currently necessary when we go out of SSA form in the RVSDG construction.

PR #803, opened 2025-02-26
Delta conversion

halvorlinder:
Roundtrip delta node


jlm/mlir/backend/JlmToMlirConverter.cpp
```diff
@@ -631,6 +635,30 @@ JlmToMlirConverter::ConvertTheta(
...
+      Builder_->getType<::mlir::LLVM::LLVMPointerType>(),
+      inputs,
+      ::llvm::StringRef(deltaNode.name()),
+      ::llvm::StringRef(llvm::ToString(deltaNode.linkage())),
+      ::llvm::StringRef(deltaNode.Section()),
+      deltaNode.constant());
+  block.push_back(delta);
+  auto & deltaBlock = delta.getRegion().emplaceBlock();
+  auto regionResults = ConvertRegion(*deltaNode.subregion(), deltaBlock);
+  assert(regionResults.size() == 1); // Delta nodes have 1 output
```
haved, line 655: Use `JLM_ASSERT`

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -567,6 +567,33 @@ MlirToJlmConverter::ConvertOperation(
...
+        outputValueType,
+        mlirDeltaNode.getName().str(),
+        ConvertLinkage(linakgeString),
+        mlirDeltaNode.getSection().str(),
+        mlirDeltaNode.getConstant());
+
+    auto outputVector = ConvertRegion(mlirDeltaNode.getRegion(), *rvsdgDeltaNode->subregion());
+
+    if (outputVector.size() != 1)
+      JLM_UNREACHABLE("Expected 1 output for Delta operation.");
```
haved, line 591: I think using unreachable here makes sense. Validation can handle printing a nice error message, but the validation code is not even in this repository, so it is nice for us to have a cheap backup test that is never compiled away

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -634,6 +662,57 @@ MlirToJlmConverter::ConvertFPSize(unsigned int size)
   }
 }
 
+llvm::linkage
+MlirToJlmConverter::ConvertLinkage(std::string stringValue)
```
haved, line 666: I would suggest this function takes a `std::string_view` instead, as it does not need its own owned string.

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -634,6 +662,57 @@ MlirToJlmConverter::ConvertFPSize(unsigned int size)
...
+{
+  if (!stringValue.compare("external_linkage"))
```
haved, line 668: This code might be better placed in `jlm/llvm/ir/linkage.hpp`, as it is the inverse of the existing `ToString`. Could you create a `FromString` there? You can use an implementation based on a `static std::unordered_map<std::string_view, linkage> = {{"blabla", x}, {"blublu", y}}`.

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -634,6 +662,57 @@ MlirToJlmConverter::ConvertFPSize(unsigned int size)
...
+  else if (!stringValue.compare("external_weak_linkage"))
+  {
+    return llvm::linkage::external_weak_linkage;
+  }
+  else if (!stringValue.compare("common_linkage"))
+  {
+    return llvm::linkage::common_linkage;
+  }
+  auto message = util::strfmt("Unsupported linkage: ", stringValue, "\n");
+  JLM_UNREACHABLE(message.c_str());
```
haved, line 713: I agree with UNREACHABLE, keep that in the `FromString`

haved (changes requested):
Looks nice, I would like the string->linkage conversion to be next to the linkage->string conversion

jlm/mlir/frontend/MlirToJlmConverter.cpp
halvorlinder: Agree

jlm/mlir/frontend/MlirToJlmConverter.cpp
halvorlinder: Yes

Issue #612, opened 2024-09-08
assert is_valid(cfg) triggered after mem2reg on silly loop+switch


haved:
The assert in question is
```
jlm-opt: jlm/llvm/ir/cfg-structure.cpp:810: void jlm::llvm::prune(cfg&): Assertion `is_valid(cfg)' failed.
```

Code triggering this assert:
```c
int lzma_decode()
{
    int x = 1;
    switch (0) {
        while (0) {
            case 0:
            case 1:
                if (1) {
                    x++;
                }
        }
    }
    return x;
}
```

The assert is only tripped if the LLVM IR is passed through `mem2reg` first, for example using the following
```
clang-18 -emit-llvm -S -c lzma_decoder.c -o lzma_decoder.ll -Xclang -disable-O0-optnone
opt lzma_decoder.ll -o lzma_decoder-opt.ll --passes=mem2reg -S
jlm-opt lzma_decoder-opt.ll -o lzma_decoder-jlm-opt.ll
```

haved:
This issue keeps cropping up. It turns out that when `mem2reg` creates phi nodes, it has some surprising output when multiple cases of a switch statement have the same target basic block. The IR after mem2reg looks something like:

```llvm
define dso_local i32 @lzma_decode() {
  switch i32 0, label %7 [
    i32 0, label %4
    i32 1, label %4
  ]

4:                                                ; preds = %3, %0, %0
  %.0 = phi i32 [ 1, %0 ], [ 1, %0 ], [ %5, %3 ]
  %5 = add nsw i32 %.0, 1
  br label %7

7:                                                ; preds = %4, %0
  %.1 = phi i32 [ 1, %0 ], [ %5, %6 ]
  ret i32 %.1
}
```

This creates a strange phi node where the same predecessor basic block is mentioned multiple times.
In this case each instance of %0 has the same value (the number literal 1), but I can also create hand-written phi nodes like this:

```llvm
define i32 @lzma_decode() {
  br label %4

4:
  %.0 = phi i32 [ 32, %0 ], [ 65, %0 ]
  ret i32 %.0
}
```
and clang will happily consume it, with no warnings, and compile it to just
```
ret 32
```

It seems like LLVM is a bit more fast-and-loose with their phi nodes than we are. Here are to small examples that both work in clang, but causes `jlm-opt` to assert false:

Problem 1:
```llvm
define i32 @lzma_decode() {
  br i1 false, label %4, label %4

4:
[...]

PR #811, opened 2025-02-28
Handle cases of duplicate edges in the control flow graph

haved:
Fixes issue #612,

Phi nodes are now required to have exactly one operand per incoming cfg node. We still allow duplicated out-edges in the CFG. If we want to remove them, a separate pass can do that.


haved:
I see the llvm test suite is failing, which is honestly not completely surprising, as I have not checked every operation that can be performed on the cfg.

It would probably be a more robust solution to change the AddOutEdge to do de-duplication

jlm/llvm/frontend/LlvmInstructionConversion.cpp
```diff
@@ -417,30 +420,49 @@ convert_branch_instruction(::llvm::Instruction * instruction, tacsvector_t & tac
...
   auto i = ::llvm::cast<::llvm::SwitchInst>(instruction);
   auto bb = ctx.get(i->getParent());
 
-  size_t n = 0;
+  JLM_ASSERT(bb->NumOutEdges() == 0);
+
+  // It is possible for multiple cases to jump to the same basic block.
+  // In which case the same outgoing edge from this basic block should be re-used.
+  // Mapping from target basic block to out edge index
+  std::unordered_map<basic_block *, uint64_t> existing_out_edges;
```
phate, line 435: camelCase

jlm/llvm/frontend/LlvmInstructionConversion.cpp
```diff
@@ -417,30 +420,49 @@ convert_branch_instruction(::llvm::Instruction * instruction, tacsvector_t & tac
...
   {
     JLM_ASSERT(it != i->case_default());
-    mapping[it->getCaseValue()->getZExtValue()] = n++;
-    bb->add_outedge(ctx.get(it->getCaseSuccessor()));
+    mapping[it->getCaseValue()->getZExtValue()] = AddOutgoingEdgeTo(it->getCaseSuccessor());
   }
 
-  bb->add_outedge(ctx.get(i->case_default()->getCaseSuccessor()));
-  JLM_ASSERT(i->getNumSuccessors() == n + 1);
+  const auto default_outedge = AddOutgoingEdgeTo(i->case_default()->getCaseSuccessor());
```
phate, line 459: camelCase

jlm/llvm/frontend/LlvmModuleConversion.cpp
```diff
@@ -49,8 +49,13 @@ convert_instructions(::llvm::Function & function, context & ctx)
   return phis;
 }
 
+/**
+ * During conversion of LLVM instructions, phi instructions become empty SsaPhiOperations.
```
phate, line 53: "became empty" sounds off here. More like, "During the conversion of LLVM instructions, phi instruction where created without operands. Once all...."

jlm/llvm/frontend/LlvmModuleConversion.cpp
```diff
@@ -65,10 +70,20 @@ patch_phi_operands(const std::vector<::llvm::PHINode *> & phis, context & ctx)
         continue;
 
       auto bb = ctx.get(phi->getIncomingBlock(n));
+
+      // In LLVM, some phi instructions have multiple operands that reference the same basic block.
+      // When that has happened "in the wild" (see issue #612), the operands that refer to the
```
phate, line 75: Please do not reference issues in the comments

jlm/llvm/ir/cfg-structure.cpp
```diff
@@ -555,10 +557,8 @@ is_valid(const llvm::cfg & cfg)
     return false;
 
   /* check basic blocks */
```
phate, line 559: Since you touched the line below, maybe you can change this comment to `//`

phate:
What can happen in LLVM that I am (nowadays) aware of:

1. An LLVM basic block can have multiple "outgoing edges" to the same basic block.
2. An LLVM basic block can have multiple "incoming edges" from the same basic block. (Must be true if 1. is true).
3. A Phi instruction can have multiple operands where the value is coming from the same basic block (must be true if 2. is true), but then the value should be the same (everything else makes no sense). It should not be possible to have different values from the same basic block in a phi instruction. The counts as an invalid IR.

jlm/llvm/frontend/LlvmInstructionConversion.cpp
```diff
@@ -392,17 +392,20 @@ convert_return_instruction(::llvm::Instruction * instruction, tacsvector_t & tac
...
+
+  // Avoid duplicate out-edges by making "conditional" branches to the same target unconditional
+  if (i->isUnconditional() || i->getSuccessor(0) == i->getSuccessor(1))
   {
     bb->add_outedge(ctx.get(i->getSuccessor(0)));
-    return {};
+    return nullptr;
   }
 
   bb->add_outedge(ctx.get(i->getSuccessor(1))); /* false */
```
phate, line 411: Could you maybe change the false comment at the as well.

haved:
@phate I agree on all 3 rules. I think it was a mistake to be too eager to avoid duplicated edges. 

The only thing that actually makes #612 an issue is that you sometimes get a different number of phi operands and in-edges.

The simplest way to handle this is simply to allow it, and instead check that every possible predecessor basic block is covered by at least one phi operand

phate:
> The only thing that actually makes https://github.com/phate/jlm/issues/612 an issue is that you sometimes get a different number of phi operands and in-edges.

> The simplest way to handle this is simply to allow it, and instead check that every possible predecessor basic block is covered by at least one phi operand

Hmm no, this does not sound like a good idea either to me. The discrepancy arises as LLVM permits multiple incoming edges to a basic block from the same source basic block, while we do not allow this. The problem is that the inedges of a CFG node are modeled as a set: https://github.com/phate/jlm/blob/997b4aee2e203b13a94d5e25507e792865171797/jlm/llvm/ir/cfg-node.hpp#L202C1-L202C43

We could change it to a vector and allow multiple incoming edges from the same source, but that would have the downside that the removal of an inedge from a node would not be amortized constant any longer, but linear.

Alternatively, we could also just try to insert empty basic blocks to ensure that we never have multiple incoming edges from the same basic block. It could easily be done as:

1. A pre-pass in LLVM before translating to JLM
2. Or do this while translating LLVM to JLM

I woudl prefer option 2, but I am not sure how much complexity that would actually entail.

haved:
@phate
I agree that in-edges are a set, but they are a set of cfg_edge, not a set of predecessor cfg_nodes. That means you can have multiple incoming edges from the same node, as long as that node also has multiple outgoing edges going to you. 

I kind of want to restrict duplicate edges, but that means transformations that remove empty blocks and such must respect it. However, transformations already need to be mindful of phi nodes, so in practice the transformations we do on CFGs might not be that extreme anyways.

I will try to see how easy it is to fix this PR before scrapping anything

phate:
@haved 

> I agree that in-edges are a set, but they are a set of cfg_edge, not a set of predecessor cfg_nodes. That means you can have multiple incoming edges from the same node, as long as that node also has multiple outgoing edges going to you. This is the current behavior.

You are absolutely right. I missed that detail. Is the problem then not that the SsaPhiOperation takes CFG nodes as input, but should actually take CFG edges: https://github.com/phate/jlm/blob/6a49a21da336b0bae5647e0cb70f98df140ef2c9/jlm/llvm/ir/operators/operators.hpp#L31C1-L37C5

If it would take CFG edges, it would:

1. Allow multiple incoming edges from the same source basic block
2. Would always have as many operands as the incoming edges of its basic block and we could enforce this.

haved:
Looking back at the original cause for the issue, before I started modifying the LLVM IR to make the test case "minimal":

The problem is that when we have a phi node like 
```llvm
%1 = phi i32 [ 1, %bb0 ], [ 1, %bb0 ]
```

our validation tries to check that both incoming values from bb0 are the same value. This check is done using a pointer comparison on the tac variables, which does not work for constants.

haved:
I made a new attempt, @phate.

I tried going down the path of associating each operand with a `cfg_edge*` instead of a `cfg_node*`, but that kind of only introduced other issues, as there is no mapping in the LLVM IR, so we would have to make one up ourselves.

The changes made in this PR now are:
 - Phi nodes are allowed to have fewer operands than the number of in-edges, as long as each incoming node is covered by exactly one operand
 - I no longer try to enforce that there are no duplicated edges in the cfg.
   - This would be better added as a separate pass that removes duplicate edges
 - I no longer try to elide phi operations when they only have one operand
   - This too should be done in a "pass", if desired

I wanted to add to validation that a basic block ends in a `branch_op` if and only if it has more than one successor, but that invariant is not satisfied in many of the cfg tests, so that has been left out

PR #812, opened 2025-03-01
Refactor rvsdg to ip graph module conversion

phate:
This PR just moves the code in a converter class. The conversion code itself did not change.


haved (approved):
Nice renaming, makes it much clearer what is going on

PR #814, opened 2025-03-03
Refactor ip graph to lllvm module conversion

phate:
This PR just moves the code in a converter class. The conversion code
itself did not change.


haved (dismissed):

haved:
Looks like some code in the hls part still references some old names, @phate

PR #820, opened 2025-03-06
Rename fpneg_op class to FNegOperation

phate:


haved:
@phate
Did you mean FNeg, or FPNeg?

phate:
@haved The instruction is called FNeg: https://llvm.org/docs/LangRef.html#fneg-instruction. It is contrary to all the other instruction names in LLVM, which start with FP. However, I would like to have the names the same as in LLVM to avoid confusion (even though that is not consitent).

haved (approved):
@phate In that case the name makes complete sense

PR #822, opened 2025-03-12
Give the hls workflow permission to comment on the PR

haved:
I am not entirely sure why, but the HLS workflow on #803 fails when it tries to write a comment about changed cycle count.

This has never been an issue before, but it might be because the creator of the PR, or the last committer, is @halvorlinder who has fewer permissions.

I am hoping this change might help with this, but the permission documentation is not very clear

### DIFF
--- a/.github/workflows/hls.yml
+++ b/.github/workflows/hls.yml
@@ -18,7 +18,7 @@ jobs:
       issues: write
       pull-requests: write
 
-      steps:
+    steps:
     - uses: actions/checkout@v4
       with:
         fetch-depth: '0'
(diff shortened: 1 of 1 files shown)


.github/workflows/hls.yml
```diff
@@ -39,4 +45,4 @@ jobs:
             owner: context.repo.owner,
             repo: context.repo.repo,
             body: "```\n" + comment + "```"
-          });
\ No newline at end of file
+          });
```
sjalander, line 48: Noise

sjalander (approved):

.github/workflows/hls.yml
haved: Sure, but most text editors insert trailing newlines automatically, so I think it makes sense to have it be there, rather than having to keep in mind that you do not add it

PR #827, opened 2025-03-14
[AndersenAgnostic] Fix RegionAwareMemoryNodeProvider, now based on call graph and handling calls to external

haved:
Introduces a call graph creation step to the RegionAwareMemoryProvider, which ensures that all possible call paths are handled, even when going via external function calls.

SCCs are found in this call graph, and they are visited in reverse topological order.
This means that for any function call found during the "annotation" step, it either points to a function in an earlier SCC which is already finished, or in the same SCC. In the latter case a flag is used to remember that the call may be recursive.

RegionSummaries also remember if they may contain recursion. Once all functions have been annotated, a final pass goes through all summaries that may contain recursion, and add the whole SCC's summary of memory locations to it.

These changes made the encoding more precise in some of the tests, so they have been updated to account for that.

Testing the llvm-test-suite with `--AAAndersenRegionAware` passes on all tests, though I managed to get the encoder to fail when combining `--AAAndersenRegionAware` with other passes. I felt this PR was large enough already, so fixing encoder issues is not done here.

I am adding the [AndersenAgnostic] on this PR, but that is only because [AndersenRegionAware] does not exist yet.


phate:
@haved I am excited to review it, but is there any chance that we could move the iteration wrapper changes into its own PR?

haved:
@phate that is absolutely possible, I can do that

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.hpp
```diff
@@ -21,20 +22,21 @@ class RegionAwareMemoryNodeProvisioning;
  * states will be routed through structural nodes and renders them independent if they do not
  * reference the same memory location. The region-aware analysis proceeds as follows:
  *
- * 1. Annotation: Each region is annotated with its utilized memory locations, the contained
- * structural nodes, the function calls it contains, and the RVSDG nodes that reference unknown
- * memory locations.
+ * 1. Call Graph Creation: creates a call graph by looking at all CALLs.
```
phate, line 25: `at all call operations`

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.hpp
```diff
@@ -210,7 +191,19 @@ class RegionAwareMemoryNodeProvider final : public MemoryNodeProvider
...
+  // The SCCs of functions, in reverse topological order
+  // If function a() calls b(), and they are not in the same SCC, a()'s SCC comes after b()'s SCC.
+  std::vector<util::HashSet<const rvsdg::LambdaNode *>> FunctionSCCs_;
+  // Which SCC contains the node representing external functions
+  size_t ExternalNodeSccIndex_;
+  // A mapping from functions to the index of the SCC they belong to in the call graph
+  std::unordered_map<const rvsdg::LambdaNode *, size_t> FunctionToSccIndex_;
+
+  // The set of memory nodes that may be affected within a given SCC. Indexed by sccIndex.
+  std::vector<util::HashSet<const PointsToGraph::MemoryNode *>> SccSummaries_;
```
phate, line 206: Could we get some spaces between the individual variables and transform the comments to doxygen comments?

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.hpp
```diff
@@ -92,111 +94,90 @@ class RegionAwareMemoryNodeProvider final : public MemoryNodeProvider
...
+  CreateCallGraph(const rvsdg::RvsdgModule & rvsdgModule);
 
-  void
-  AnnotateSimpleNode(const rvsdg::SimpleNode & provider);
+  /**
+   * Collects all lambda nodes defined in the given module, in an unspecified order.
+   * @param rvsdgModule the module
+   * @return a list of all lambda nodes in the module
+   */
+  std::vector<const rvsdg::LambdaNode *>
```
phate, line 109: This method can be made static if I am not mistaken

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.cpp
```diff
@@ -675,159 +603,288 @@ RegionAwareMemoryNodeProvider::Create(
...
+    {
+      if (auto lambda = dynamic_cast<rvsdg::LambdaNode *>(&node))
+      {
+        result.push_back(lambda);
+      }
+      else if (auto structural = dynamic_cast<rvsdg::StructuralNode *>(&node))
+      {
+        for (size_t i = 0; i < structural->nsubregions(); i++)
+        {
+          recurse(*structural->subregion(0), recurse);
```
phate, line 737: It should be `structural->subregion(n)` and not `structural->subregion(0)`, right?

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.cpp
```diff
@@ -675,159 +603,288 @@ RegionAwareMemoryNodeProvider::Create(
...
+  ExternalNodeSccIndex_ = sccIndex[externalNodeIndex];
+}
+
+std::vector<const rvsdg::LambdaNode *>
+RegionAwareMemoryNodeProvider::CollectLambdaNodes(const rvsdg::RvsdgModule & rvsdgModule)
+{
+  std::vector<const rvsdg::LambdaNode *> result;
+
+  // Recursively traverses all structural nodes, but does not enter into lambdas
+  const auto CollectLambdasInRegion = [&](rvsdg::Region & region, auto recurse) -> void
```
phate, line 725: You can avoid handing in `recurse`. Just explicitly provide the type for the lambda expression (without auto) and you should be able to recursively call it.

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.cpp
```diff
@@ -675,159 +603,288 @@ RegionAwareMemoryNodeProvider::Create(
...
+          || dynamic_cast<const PointsToGraph::ImportNode *>(&callee))
+      {
+        // Add the edge caller -> node representing external functions
+        callGraphSuccessors[callerIndex].Insert(externalNodeIndex);
+      }
+    }
+  };
+
+  // Recursive function finding call nodes, adding edges to the call graph
+  const auto HandleCalls = [&](rvsdg::Region & region, size_t callerIndex, auto recurse) -> void
```
phate, line 660: The same here, recurse should not be necessary.

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.cpp
```diff
@@ -675,159 +603,288 @@ RegionAwareMemoryNodeProvider::Create(
...
+        // Look up which call graph node represents the target lambda
+        JLM_ASSERT(callGraphNodeIndex.find(&lambdaNode) != callGraphNodeIndex.end());
+        const auto calleeCallGraphNode = callGraphNodeIndex[&lambdaNode];
+
+        // Add the edge caller -> callee to the call graph
+        callGraphSuccessors[callerIndex].Insert(calleeCallGraphNode);
+      }
+      else if (
+          dynamic_cast<const PointsToGraph::ExternalMemoryNode *>(&callee)
+          || dynamic_cast<const PointsToGraph::ImportNode *>(&callee))
```
phate, line 651: Maybe use `PointsToGraph::Is<>()` here instead of `dynamic_cast<>()`

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.cpp
```diff
@@ -837,187 +894,104 @@ RegionAwareMemoryNodeProvider::AnnotateMemcpy(const rvsdg::SimpleNode & memcpyNo
...
+      const auto & lambdaNode = lambdaCallee->GetLambdaNode();
+
+      // Look up which SCC the callee belongs to
+      JLM_ASSERT(FunctionToSccIndex_.find(&lambdaNode) != FunctionToSccIndex_.end());
+      targetSccIndex = FunctionToSccIndex_[&lambdaNode];
     }
-    else if (auto phiNode = dynamic_cast<const phi::node *>(node))
+    else if (
+        dynamic_cast<const PointsToGraph::ExternalMemoryNode *>(&callee)
+        || dynamic_cast<const PointsToGraph::ImportNode *>(&callee))
```
phate, line 923: `PointsToGraph::Is<>()`

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.cpp
```diff
@@ -148,116 +133,205 @@ class RegionAwareMemoryNodeProvider::Statistics final : public util::Statistics
...
   util::HashSet<const PointsToGraph::MemoryNode *> MemoryNodes_;
-  util::HashSet<const rvsdg::SimpleNode *> UnknownMemoryNodeReferences_;
 
-  util::HashSet<const CallNode *> RecursiveCalls_;
-  util::HashSet<const CallNode *> NonRecursiveCalls_;
-  util::HashSet<const rvsdg::StructuralNode *> StructuralNodes_;
+  // Which SCC in the call graph does the function containing this region belong to
+  size_t CallGraphSccIndex_;
+  // If this region contains a call that may enter a new instance of this same region
+  bool PossiblyRecursive_;
```
phate, line 334: The name feels a bit misleading. How about `ContainsRecursiveCall_`

phate (commented):
The first batch of comments. I need to go over it one more time, as I was not able to finish it completely.

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.cpp
haved: Woah, good catch

haved: This for loop will in practice only ever visit phi nodes, which only have one subregion, so that explains why there were no issues. I fixed it anyways, of course.

phate: Well, delta nodes as well, but they also just have one subregion. We should introduce a subregion range in the structural nodes to avoid this problem in the future. It was on my TODO list since a while, but I never got around to it.

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.cpp
haved: I can do that, yes. There is more overhead from doing that (see [here](https://quick-bench.com/q/f49eDKZxXpISkSs6vk6vywOCAww)), but I highly doubt that runtime makes much of a difference in the grand scheme of things.

haved:
@phate I addressed your comments so far

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.cpp
phate: I do not think that it makes a difference i practice. We have bigger fishes to fry at other places.

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.hpp
```diff
@@ -210,7 +191,31 @@ class RegionAwareMemoryNodeProvider final : public MemoryNodeProvider
...
+
+  /**
+   * A mapping from functions to the index of the SCC they belong to in the call graph
+   */
+  std::unordered_map<const rvsdg::LambdaNode *, size_t> FunctionToSccIndex_;
+
+  /**
+   * The set of memory nodes that may be affected within a given SCC. Indexed by sccIndex.
+   */
+  std::vector<util::HashSet<const PointsToGraph::MemoryNode *>> SccSummaries_;
```
phate, line 218: One thing that just popped up in my mind: The Provision method is expected to be called multiple times and arrive at the same result. However, this would not happen right now as the private attributes are not set up right now. That is also the reason why I always put everything into Context and then just reallocate this.

phate (changes requested):

jlm/llvm/opt/alias-analyses/RegionAwareMemoryNodeProvider.hpp
haved: That is a very good point, I can create a Context wrapper

haved:
@phate I created a `Context` to make it easier to clear temporary state (and actually clear it).

Regarding `setjmp` and `longjmp`, they are currently handled, as we are currently very conservative within each SCC, and anything that calls either will end up in the big "external + almost everything else"-SCC.

phate (approved):

PR #830, opened 2025-03-16
Generalize PtrIterator to support other kinds of wrapping iterators

haved:


tests/jlm/util/TestIteratorWrapper.cpp
```diff
@@ -7,7 +7,7 @@
 #include <test-util.hpp>
```
phate, line 7: Let's call it IteratorWrapperTests.cpp, then alphabetical sorting still works.

jlm/util/Makefile.sub
```diff
@@ -19,7 +19,7 @@ libutil_HEADERS = \
     jlm/util/intrusive-list.hpp \
     jlm/util/iterator_range.hpp \
     jlm/util/Math.hpp \
-    jlm/util/PtrIterator.hpp \
+    jlm/util/IteratorWrapper.hpp \
```
phate, line 22: The files seem to be alphabetically ordered in this list. It is currently at the wrong position

phate (changes requested):
Just some organizational issues, nothing major.

phate (approved):

sjalander:
Why use Ptr instead of spelling it out to Pointer?

haved:
> Why use Ptr instead of spelling it out to Pointer?

I guess I did that originally because I was iterating over `unique_ptr`s, before realizing I could make it general. @sjalander

PR #832, opened 2025-03-18
Use after free bug fix in HLS

sjalander:


sjalander:
The bug never seems to have triggered on Linux but caused random HLS test errors on Mac

haved (approved):

PR #838, opened 2025-03-28
Roundtrip Graph imports

halvorlinder:
Requires https://github.com/EECS-NTNU/mlir_rvsdg/pull/8

Graph imports are reprsented by omega argument ops at the beginning of the omega block

### DIFF
--- a/jlm/mlir/backend/JlmToMlirConverter.cpp
+++ b/jlm/mlir/backend/JlmToMlirConverter.cpp
@@ -62,14 +62,25 @@ JlmToMlirConverter::Print(::mlir::rvsdg::OmegaNode & omega, const util::filepath
 ::mlir::rvsdg::OmegaNode
 JlmToMlirConverter::ConvertModule(const llvm::RvsdgModule & rvsdgModule)
 {
-  return ConvertOmega(rvsdgModule.Rvsdg());
-}
+  auto & graph = rvsdgModule.Rvsdg();
 
-::mlir::rvsdg::OmegaNode
-JlmToMlirConverter::ConvertOmega(const rvsdg::Graph & graph)
-{
   auto omega = Builder_->create<::mlir::rvsdg::OmegaNode>(Builder_->getUnknownLoc());
   auto & omegaBlock = omega.getRegion().emplaceBlock();
+
+  for (size_t i = 0; i < graph.GetRootRegion().narguments(); ++i)
+  {
+    auto arg = graph.GetRootRegion().argument(i);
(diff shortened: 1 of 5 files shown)


jlm/mlir/backend/JlmToMlirConverter.cpp
```diff
@@ -62,15 +62,27 @@ JlmToMlirConverter::Print(::mlir::rvsdg::OmegaNode & omega, const util::filepath
...
-::mlir::rvsdg::OmegaNode
-JlmToMlirConverter::ConvertOmega(const rvsdg::Graph & graph)
-{
   auto omega = Builder_->create<::mlir::rvsdg::OmegaNode>(Builder_->getUnknownLoc());
   auto & omegaBlock = omega.getRegion().emplaceBlock();
 
+  for (size_t i = 0; i < graph.GetRootRegion().narguments(); ++i)
+  {
+    auto arg = graph.GetRootRegion().argument(i);
+    auto imp = dynamic_cast<llvm::GraphImport *>(arg);
```
haved, line 73: Bruk `util::AssertedCast`

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -37,14 +37,48 @@ MlirToJlmConverter::ReadAndConvertMlir(const util::filepath & filePath)
   {
     JLM_ASSERT("Parsing MLIR input file failed.");
   }
-  return ConvertMlir(block);
+  auto & topNode = block->front();
+  auto omegaNode = ::mlir::dyn_cast<::mlir::rvsdg::OmegaNode>(topNode);
+  return ConvertOmega(omegaNode);
```
haved, line 42: Why has this call to `ConvertMlir()` been replaced? Isn't the new code identical to the body of `ConvertMlir()`?

haved (changes requested):
I have some small suggestions for the code.

One question I have is if the GraphImports you create during MLIR->RVSDG can actually be used. During conversions of blocks, there is a
```cpp
std::unordered_map<::mlir::Operation *, rvsdg::Node *> operationsMap;
```
for mapping operations, but this map never gets the `OmagaArgument` and corresponding `GraphImport`s added. Since `GraphImport` is not a real node, but actually an `output*`, we might have to change how the mapping is implemented.

To summarize, I do not think `MlirToJlmConverter::GetConvertedInputs` is currently able to convert usage of `OmegaArguments`.

Or maybe I am not understanding something?

PR #839, opened 2025-03-28
Region Sequentializer

phate:
This PR separates the sequentialization of the nodes from the LLVM converter. The reason for this change are as follows:

1. The idea is to implement a small debugging command line tool that takes an LLVM module as input, and outputs all possible sequentializations of nodes within the module. Each new combination is output as its own module, which can then be compiled and executed to see that all combinations result in the same output. This obviously only works for small modules.
2. We can eventually implement some heuristics to give a "better" sequentialization as the random one we are currently have.

This draft is mean to get feedback. I will later split it up in proper PRs and clean it up.

@haved @caleridas

### DIFF
--- a/jlm/llvm/backend/RegionSequentializer.cpp
+++ b/jlm/llvm/backend/RegionSequentializer.cpp
@@ -11,7 +11,7 @@
 namespace jlm::llvm
 {
 
-RegionSequentializer::~RegionSequentializer() = default;
+RegionSequentializer::~RegionSequentializer() noexcept = default;
 
 RegionSequentializer::RegionSequentializer(rvsdg::Region & region)
     : Region_(&region)


tests/jlm/llvm/backend/ExhaustiveSequentializerTests.cpp
```diff
@@ -0,0 +1,36 @@
...
+    }
+    std::cout << std::endl;
+
+    sequentializer.ComputeNextSequentializations();
+  }
+
+  return 0;
+}
+
+JLM_UNIT_TEST_REGISTER("jlm/llvm/backend/ExhaustiveSequentializerTests-Test", Test)
```
phate, line 36: I will implement proper unit tests once the region sequentializer interface is settled etc.

haved (commented):
Overall the design looks good and extensible. My only issue is with the API for iterating over sequentializations, as it is hard to keep track of when sequentializations start looping (silently when requesting a sequentialization).

Could we instead have a design where a sequentializer's `GetSequentialization()` always gives a valid sequentialization, and `ComputeNextSequentialization()` intead starts returning false when it is no longer changing the current sequentialization.

While it might be nice in some cases to query if there exist more sequentializations, without moving on to the next one, that does make it a bit harder for implementations to implement the interface. My suggested interface would instead be used like

```c
do {
  auto & seq= sequentializer.GetSequentialization();
} while(sequentializer.ComputeNextSequentialization())
```

@phate

PR #840, opened 2025-03-29
Fix operand order in select creation in RvsdgToIpGraphConverter

### LINKED ISSUE #586: Miscompilation of trivial loop after NodePushOut and NodePullIn

### LINKED ISSUE #836: RVSDG to LLVM backend emits "select" with the wrong operands
I seem to recall seeing this issue before, but I guess we never fixed it.

Given the following C code
```c
int main() {
    int l = 0;
    int res = -1;
    while (l < 1000) {
        res = l;
        if(l == 0) {
            break;
        }
        l++;
    }
    return res;
}
```

it is supposed to return 0, but we can get it to return -1 by compiling it like so:

```sh
clang -Xclang -disable-O0-optnone -S -emit-llvm minimal.c -o minimal.ll
opt -passes=mem2reg -S minimal.ll -o minimal-m2r.ll
jlm-opt minimal-m2r.ll -o minimal-jlm-opt.ll
clang minimal-jlm-opt.ll -o minimal-jlm-opt
# Execute it
./minimal-jlm-opt
```

When dumping the RVSDG to dot, everything looks correct.
The problem is that the RVSDG contains a gamma node that selects between two values:

Image

This turns into the LLVM instruction 
```llvm
%10 = select i1 %8, i32 %0, i32 %1
```

And as far as I can tell %0 and %1 should be the other way around. This causes the program to miscompile.

phate:
When handling an empty gamma node with 2 subregions, then the RvsdgToIpGraphConverter can create select operations instead of basic blocks. However, the operands of the select operations were switched. This PR fixes this problem.

Close #836
Close #586

### DIFF
--- a/jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
+++ b/jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
@@ -296,7 +296,7 @@ RvsdgToIpGraphConverter::ConvertEmptyGammaNode(const rvsdg::GammaNode & gammaNod
       const auto vo1 = Context_->GetVariable(output1);
       basicBlock->append_last(
           ctl2bits_op::create(Context_->GetVariable(predicate), rvsdg::bittype::Create(1)));
-      basicBlock->append_last(SelectOperation::create(basicBlock->last()->result(0), vo0, vo1));
+      basicBlock->append_last(SelectOperation::create(basicBlock->last()->result(0), vo1, vo0));
     }
 
     Context_->InsertVariable(output, basicBlock->last()->result(0));
--- a/jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
+++ b/jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
@@ -292,11 +292,14 @@ RvsdgToIpGraphConverter::ConvertEmptyGammaNode(const rvsdg::GammaNode & gammaNod
     }
     else
     {
-      const auto vo0 = Context_->GetVariable(output0);
-      const auto vo1 = Context_->GetVariable(output1);
+      const auto falseAlternative = Context_->GetVariable(output0);
+      const auto trueAlternative = Context_->GetVariable(output1);
       basicBlock->append_last(
           ctl2bits_op::create(Context_->GetVariable(predicate), rvsdg::bittype::Create(1)));
-      basicBlock->append_last(SelectOperation::create(basicBlock->last()->result(0), vo1, vo0));
+      basicBlock->append_last(SelectOperation::create(
+          basicBlock->last()->result(0),
+          trueAlternative,
+          falseAlternative));
     }
 
     Context_->InsertVariable(output, basicBlock->last()->result(0));
--- a/tests/jlm/llvm/backend/RvsdgToIpGraphConverterTests.cpp
+++ b/tests/jlm/llvm/backend/RvsdgToIpGraphConverterTests.cpp
@@ -123,6 +123,83 @@ JLM_UNIT_TEST_REGISTER(
     GammaWithoutMatch)
 
 static int
+EmptyGammaWithTwoSubregions()
+{
+  using namespace jlm::llvm;
+  using namespace jlm::tests;
+  using namespace jlm::util;
+
+  // Arrange
+  auto valueType = valuetype::Create();
+  auto functionType = jlm::rvsdg::FunctionType::Create(
+      { jlm::rvsdg::bittype::Create(32), valueType, valueType },
+      { valueType });
+
+  RvsdgModule rvsdgModule(filepath(""), "", "");
+
+  const auto lambdaNode = jlm::rvsdg::LambdaNode::Create(
+      rvsdgModule.Rvsdg().GetRootRegion(),
+      LlvmLambdaOperation::Create(functionType, "lambdaOutput", linkage::external_linkage));
+  const auto trueValue = lambdaNode->GetFunctionArguments()[1];
+  const auto falseValue = lambdaNode->GetFunctionArguments()[2];
+
+  const auto matchResult = match(32, { { 0, 0 } }, 1, 2, lambdaNode->GetFunctionArguments()[0]);
+
+  const auto gammaNode0 = jlm::rvsdg::GammaNode::create(matchResult, 2);
+  const auto & c0 = jlm::rvsdg::CreateOpNode<jlm::rvsdg::ctlconstant_op>(
+      *gammaNode0->subregion(0),
+      jlm::rvsdg::ctlvalue_repr(0, 2));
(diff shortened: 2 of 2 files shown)


haved (approved):
Great, nice to fix two issues with one PR

haved:
I am a bit unsure about the code for handling when the predicate comes from a match-node, as it only uses the default value of the match, and doesn't actually look at the match mapping. That is a separate piece of code to the one fixed here, however

phate:
@haved What do you think is wrong with the code where the predicate comes from a match-node?

haved:
@phate it only looks at the default value. This works if all matches with two alternatives always look like either

```
{1 -> 1}, default=0
```

or

```
{1 -> 0}, default=1
```

But can you never have situations like

```
{1 -> 0, 0 -> 1}, default = 0
```

or

```
{0 -> 0}, default = 0
```
?

Maybe you can never get situations like that in practice, but it was a bit surprising that only the default value was used

phate:
@haved 

> {1 -> 0, 0 -> 1}, default = 0

Firstly, this would be a gamma with 3 cases/subregions. The method only handles gammas with 2 subregions.
Secondly, I believe this should not be legal (I am not sure whether we check this when we create the match). This is effectively the same as: `{0 -> 1}, default = 0`.

`{0 -> 0}, default = 0`: That is not really legal either.

I am pretty sure we can tighten the code, but your two examples are not really contradictory to the code right now (at least I believe so). We can certainly discuss this further, but then I guess we should open an issue.

haved:
@phate I can open an issue, yes. I think if we are compiling a switch case, it is possible to end up with pretty arbitrary match operations. If it ever actually shows up in practice, I am not so sure

PR #841, opened 2025-03-29
Add color to dot output and refactor of view HLS

sjalander:
Written by: David Metz


jlm/hls/util/view.cpp
```diff
@@ -85,44 +97,44 @@ port_to_dot(const std::string & display_name, const std::string & dot_name)
...
         "            </TABLE>\n"
         "> tooltip=\""
-      + dot_name + "\"];\n";
+      + dot_name + "\" color=" + color + " fontcolor=" + color + "];\n";
   return dot;
 }
 
 std::string
-argument_to_dot(rvsdg::RegionArgument * argument)
+argument_to_dot(rvsdg::RegionArgument * argument, std::string color)
```
phate, line 105: `const std::string_view &`

jlm/hls/util/view.cpp
```diff
@@ -30,48 +30,60 @@ hex(size_t i)
...
   {
-    return jlm::util::strfmt("si", hex((intptr_t)input), ":", "default");
+    return util::strfmt("si", hex((intptr_t)input), ":", "default");
   }
   JLM_UNREACHABLE("not implemented");
 }
 
 std::string
-port_to_dot(const std::string & display_name, const std::string & dot_name)
+port_to_dot(const std::string & display_name, const std::string & dot_name, std::string & color)
```
phate, line 86: `const std::string_view &`

jlm/hls/util/view.cpp
```diff
@@ -85,44 +97,44 @@ port_to_dot(const std::string & display_name, const std::string & dot_name)
...
-  auto dot_name = jlm::util::strfmt("a", hex((intptr_t)argument));
-  return port_to_dot(display_name, dot_name);
+  auto display_name = util::strfmt("a", argument->index());
+  auto dot_name = util::strfmt("a", hex((intptr_t)argument));
+  return port_to_dot(display_name, dot_name, color);
 }
 
 std::string
-result_to_dot(rvsdg::RegionResult * result)
+result_to_dot(rvsdg::RegionResult * result, std::string color)
```
phate, line 113: same here

jlm/hls/util/view.cpp
```diff
@@ -85,44 +97,44 @@ port_to_dot(const std::string & display_name, const std::string & dot_name)
...
-  auto dot_name = jlm::util::strfmt("r", hex((intptr_t)result));
-  return port_to_dot(display_name, dot_name);
+  auto display_name = util::strfmt("r", result->index());
+  auto dot_name = util::strfmt("r", hex((intptr_t)result));
+  return port_to_dot(display_name, dot_name, color);
 }
 
 std::string
-structural_input_to_dot(rvsdg::StructuralInput * structuralInput)
+structural_input_to_dot(rvsdg::StructuralInput * structuralInput, std::string color)
```
phate, line 121: same here

jlm/hls/util/view.cpp
```diff
@@ -85,44 +97,44 @@ port_to_dot(const std::string & display_name, const std::string & dot_name)
...
-  auto dot_name = jlm::util::strfmt("si", hex((intptr_t)structuralInput));
-  return port_to_dot(display_name, dot_name);
+  auto display_name = util::strfmt("si", structuralInput->index());
+  auto dot_name = util::strfmt("si", hex((intptr_t)structuralInput));
+  return port_to_dot(display_name, dot_name, color);
 }
 
 std::string
-structural_output_to_dot(rvsdg::StructuralOutput * structuralOutput)
+structural_output_to_dot(rvsdg::StructuralOutput * structuralOutput, std::string color)
```
phate, line 129: same here

jlm/hls/util/view.cpp
```diff
@@ -162,7 +174,10 @@ isForbiddenChar(char c)
 }
 
 std::string
-structural_node_to_dot(rvsdg::StructuralNode * structuralNode)
+structural_node_to_dot(
+    rvsdg::StructuralNode * structuralNode,
+    std::unordered_map<rvsdg::output *, std::string> & o_color,
+    std::unordered_map<rvsdg::input *, std::string> & i_color)
```
phate, line 180: Naming: `inputColors` and `outputColors`

jlm/hls/util/view.cpp
```diff
@@ -210,34 +211,25 @@ structural_node_to_dot(rvsdg::StructuralNode * structuralNode)
...
 
   return dot.str();
 }
 
 std::string
-simple_node_to_dot(jlm::rvsdg::SimpleNode * simpleNode)
+simple_node_to_dot(
+    rvsdg::SimpleNode * simpleNode,
+    std::unordered_map<rvsdg::output *, std::string> & o_color,
+    std::unordered_map<rvsdg::input *, std::string> & i_color)
```
phate, line 232: same here

jlm/hls/util/view.cpp
```diff
@@ -310,46 +304,38 @@ simple_node_to_dot(jlm::rvsdg::SimpleNode * simpleNode)
 }
 
 std::string
-region_to_dot(rvsdg::Region * region)
+region_to_dot(
+    rvsdg::Region * region,
+    std::unordered_map<rvsdg::output *, std::string> & o_color,
+    std::unordered_map<rvsdg::input *, std::string> & i_color)
```
phate, line 310: same here

jlm/hls/util/view.cpp
```diff
@@ -360,64 +346,110 @@ region_to_dot(rvsdg::Region * region)
...
 
   return dot.str();
 }
 
 std::string
-to_dot(rvsdg::Region * region)
+to_dot(
+    rvsdg::Region * region,
+    std::unordered_map<rvsdg::output *, std::string> & o_color,
+    std::unordered_map<rvsdg::input *, std::string> & i_color)
```
phate, line 383: same here

jlm/hls/util/view.cpp
```diff
@@ -360,64 +346,110 @@ region_to_dot(rvsdg::Region * region)
...
 }
 
+void
+dot_to_svg(const std::string & file_name)
+{
+  auto cmd = "dot -Tsvg -O " + file_name;
+  if (system(cmd.c_str()))
+    exit(EXIT_FAILURE);
+}
+
```
phate, line 454: This can be done in bash, no? Why do we need this here in code, when it only does and external system call?

phate (changes requested):
Preferably, I would love if you would use an enum for the colors, and then only convert them to strings in the last moment when you write it out.

jlm/hls/util/view.cpp
sjalander: Changed to 
const ViewColors &

jlm/hls/util/view.cpp
sjalander: Changed to 
const ViewColors &

jlm/hls/util/view.cpp
sjalander: Changed to 
const ViewColors &

jlm/hls/util/view.cpp
sjalander: Changed to 
const ViewColors &

jlm/hls/util/view.cpp
sjalander: Changed

jlm/hls/util/view.cpp
sjalander: Changed

jlm/hls/util/view.cpp
sjalander: Changed

jlm/hls/util/view.cpp
```diff
@@ -19,6 +18,29 @@
...
+    break;
+
+  case RED:
+    return "red";
+    break;
+
+  default:
+    JLM_UNREACHABLE("HLS view color not defined");
+    break;
+  }
```
phate, line 41: Why not getting rid of the NONE color and just returning `""` be default?

jlm/hls/util/view.hpp
```diff
@@ -13,17 +13,61 @@
 namespace jlm::hls
 {
 
+enum ViewColors
+{
+  NONE,
```
phate, line 18: That is not really a color? Why is it necessary? What is the difference between NONE and not specifying a color?

jlm/hls/util/view.hpp
```diff
@@ -13,17 +13,61 @@
...
+  BLACK,
+  RED
+};
```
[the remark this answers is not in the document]
sjalander: They are neither methods nor variables.
They are unchangeable values.

jlm/hls/util/view.cpp
sjalander: Still need to define something representing the option in the enum, so then prefer to be explicit.
The default will only be triggered if one add another color to ViewColors and doesn't update this function.

jlm/hls/util/view.hpp
sjalander: @davidmetz
GetDefaultLabel() use "" as the default, while GetDefaultColor() use black.
What's the significance of using "" instead of a predefined color?

jlm/hls/util/view.cpp
```diff
@@ -28,123 +50,155 @@ hex(size_t i)
...
+  if (f == map.end())
+  {
+    return def;
+  }
+  return f->second;
+}
+
+template<class T>
+ViewColors
+GetDefaultLabel(std::unordered_map<T *, ViewColors> & map, T * elem, ViewColors def = NONE)
```
sjalander, line 91: This should be changed back to a string. Labels are not colors

jlm/hls/util/view.cpp
```diff
@@ -19,6 +18,29 @@
 namespace jlm::hls
 {
 
+std::string
+ViewcolorToString(const ViewColors & color)
```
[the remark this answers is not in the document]
sjalander: Changed

jlm/hls/util/view.hpp
sjalander: NONE has been removed

PR #842, opened 2025-03-30
Rename sitofp_op class to SIToFPOperation

phate:


jlm/llvm/backend/IpGraphToLlvmConverter.cpp
```diff
@@ -1373,7 +1373,7 @@ IpGraphToLlvmConverter::convert_operation(
   {
     return convert_cast<::llvm::Instruction::SExt>(op, arguments, builder);
   }
-  if (is<sitofp_op>(op))
+  if (is<SIToFPOperation>(op))
```
sjalander, line 1376: Personally, I think it would be easier to parse SI2FP as one now need to identify the lower case 'o' to then see To and realize that there are 3 parts to it.

sjalander (approved):
Personally, I think it would be easier to parse SI2FP as one now need to identify the lower case 'o' to then see To and realize that there are 3 parts to it.

jlm/llvm/backend/IpGraphToLlvmConverter.cpp
phate: The LLVM operator is called likes this. I want them to be the same for the sake of familarity.

PR #845, opened 2025-03-30
add PrecisionEvaluation to alias analysis

haved:
It uses a new `PairwiseAliasAnalysis` interface to check pairs of pointers.

Depending on mode, it either checks all pairs of pointers in each function, or it only checks clobbering between loads and stores. In the latter mode, all loads and stores are regarded as uses, while only stores are regarded as clobbers.

We might want to move some logic out of the `PrecisionAnalysis` file, as I later plan to write a different `PairwiseAliasAnalysis` that is not based on a PointsToGraph, for comparison.

The name `PairwiseAliasAnalysis`could also possibly be changed. The problem is that the current `AliasAnalysis` is actually a points-to-analysis, so we could juggle the names a bit to make it more clear.

The output looks like
```
AliasAnalysisPrecisionEvaluation spotsFilled.c PrecisionEvaluationMode:ClobberingStores PairwiseAliasAnalysisType:PointsToGraphAliasAnalysis #MayAliasQueries:18 DumpFile:./jlm-out/spotsFilled-JDT40v-AAPrecisionEvaluation-1.log ModuleAverageMayAliasRate:0.444444 PrecisionEvaluationTimer[ns]:27849
```

and function-wise numbers are written to a separate file (one line per function)

```
LAMBDA[takeSpot] [6 used pointers, 3 clobbering pointers]: 0.444444
```

@phate

### DIFF
--- a/jlm/llvm/opt/alias-analyses/PrecisionEvaluator.hpp
+++ b/jlm/llvm/opt/alias-analyses/PrecisionEvaluator.hpp
@@ -1,0 +1,229 @@
+/*
+ * Copyright 2025 Håvard Krogstie <krogstie.havard@gmail.com>
+ * See COPYING for terms of redistribution.
+ */
+
+#ifndef JLM_LLVM_OPT_ALIAS_ANALYSES_PRECISIONEVALUATOR_HPP
+#define JLM_LLVM_OPT_ALIAS_ANALYSES_PRECISIONEVALUATOR_HPP
+
+#include <jlm/llvm/opt/alias-analyses/PointsToGraph.hpp>
+#include <jlm/rvsdg/RvsdgModule.hpp>
+#include <jlm/util/Statistics.hpp>
+
+#include <unordered_map>
+
+namespace jlm::llvm::aa
+{
+
+/**
+ * Interface for making alias analysis queries about pairs of pointers
+ */
+class PairwiseAliasAnalysis
+{
+public:
+  PairwiseAliasAnalysis();
+  virtual ~PairwiseAliasAnalysis();
+
+  /**
+   * If this alias analysis is unable to determine NoAlias, it queries the optional backup analysis.
+   * @param backup the PairwiseAliasAnalysis instance to use as backup
+   */
+  void
+  SetBackup(PairwiseAliasAnalysis * backup) noexcept
+  {
+    Backup_ = backup;
+  }
+
+  PairwiseAliasAnalysis *
+  GetBackup() const noexcept
+  {
+    return Backup_;
+  }
+
+  /**
+   * Checks if the pointers represented by p1 and p2 may alias.
+   * If this analysis is unable to prove No Alias, and a backup is provided, the backup is asked.
+   * @param p1 the first pointer value
+   * @param p2 the second pointer value
+   * @return false if the analysis is able to prove that they do not alias, true otherwise
+   */
+  bool
+  MayAlias(const rvsdg::output & p1, const rvsdg::output & p2);
+
+  /**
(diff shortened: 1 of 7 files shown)


jlm/llvm/opt/alias-analyses/PrecisionEvaluator.hpp
```diff
@@ -0,0 +1,234 @@
...
+protected:
+  virtual bool
+  MayAliasImpl(const rvsdg::output & p1, const rvsdg::output & p2) = 0;
+
+  virtual std::string
+  ToStringImpl() const = 0;
+
+private:
+  // Another instance of PairwiseAliasAnalysis, queried if the current analysis says "May Alias"
+  PairwiseAliasAnalysis * Backup_ = nullptr;
```
phate, line 68: I think for the backup option, you should implement another class that statically takes two PairwiseAliasAnalyses and applies the one first and then the second. This would also enable to have more than one backup.

phate: Implementing this in another class would also permit to directly overload the `MayAlias()` and not require the `Impl()` variants.

jlm/llvm/opt/alias-analyses/PrecisionEvaluator.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+  /**
+   * Checks if the pointers represented by p1 and p2 may alias.
+   * If this analysis is unable to prove No Alias, and a backup is provided, the backup is asked.
+   * @param p1 the first pointer value
+   * @param p2 the second pointer value
+   * @return false if the analysis is able to prove that they do not alias, true otherwise
+   */
+  bool
+  MayAlias(const rvsdg::output & p1, const rvsdg::output & p2);
```
phate, line 51: Would it possible to generalize this where this might be: `AliasInformation GetAliasInformation(...)`, where AliasInformation is:

```
enum AliasInformation
{
   NoAlias,
   MayAlias
}
```

This could then also be extended with potential MustAlias at some point etc.

jlm/llvm/opt/alias-analyses/PrecisionEvaluator.hpp
```diff
@@ -0,0 +1,234 @@
...
+{
+  // In this mode, a pointer use is a store / load, and a clobber is a store
+  ClobberingStores,
+
+  // In this mode, all pointer outputs are both used, and considered to clobber.
+  // This is equivalent to checking all pairs of pointers in each function.
+  // Temporary pointers, such as the result of a pointer offset, are counted as separate pointers.
+  // This is consistent with Lattner's 2007 DSA paper.
+  AllPointerPairs
+};
```
phate, line 105: Might be subjective, but I would move this into the PrecisionEvaluator class and just call it Mode.

jlm/llvm/opt/alias-analyses/PrecisionEvaluator.hpp
```diff
@@ -0,0 +1,234 @@
...
+    }
+  };
+
+  struct Context
+  {
+    // Precision info per function in the evaluated module
+    std::unordered_map<const rvsdg::LambdaNode *, PrecisionInfo> PerFunctionPrecision;
+
+    /**
+     * During traversal of the current function, which pointers are used.
```
phate, line 216: English as speaks Yoda.

phate (commented):
Overall, this looked good. There is certainly nothing wrong with it. Nevertheless, I gave some ideas/suggestions for architecture, and some more minor comments.

PR #854, opened 2025-04-05
Fix select creation in RvsdgToIpGraphConverter

### LINKED ISSUE #846: Assert fails when compiling simple switch with char*
The following C code:

```c
const char* func(int colorid)
{
	const char *cp = "a";
	switch (colorid) {
		case 1:
			cp = "b"; break;
	}

	return cp;
}
```

when compiled with
```c
clang -Xclang -disable-O0-optnone -S -emit-llvm -c test.c -o test.ll
opt -passes=mem2reg test.ll -S -o test-m2r.ll
jlm-opt test-m2r.ll -o test-jlm-opt.ll
```

fails during `jlm-opt` with the error:
```
Type error - expected : bit32, received : bit1
```

Interestingly, it does not fail if the `const char*` is replaced by an `int`.

phate:
The RvsdgToIpGraphConverter was not handling the conversion of empty gamma nodes with match nodes to select operations correctly. This PR fixes the problems with it by creating select predicate operations in case there is no identity mapping in the match operation.

Close #846

### DIFF
--- a/jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
+++ b/jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
@@ -8,6 +8,7 @@
 #include <jlm/llvm/ir/cfg-structure.hpp>
 #include <jlm/llvm/ir/ipgraph-module.hpp>
 #include <jlm/llvm/ir/operators.hpp>
+#include <jlm/llvm/ir/operators/IntegerOperations.hpp>
 #include <jlm/llvm/ir/RvsdgModule.hpp>
 #include <jlm/rvsdg/gamma.hpp>
 #include <jlm/rvsdg/traverser.hpp>
@@ -253,7 +254,7 @@ RvsdgToIpGraphConverter::ConvertEmptyGammaNode(const rvsdg::GammaNode & gammaNod
   JLM_ASSERT(gammaNode.subregion(0)->nnodes() == 0 && gammaNode.subregion(1)->nnodes() == 0);
 
   // both regions are empty, create only select instructions
-  const auto predicate = gammaNode.predicate()->origin();
+  const auto gammaPredicate = gammaNode.predicate()->origin();
   const auto controlFlowGraph = Context_->GetControlFlowGraph();
 
   const auto basicBlock = basic_block::create(*controlFlowGraph);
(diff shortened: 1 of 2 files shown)


jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
```diff
@@ -277,25 +278,58 @@ RvsdgToIpGraphConverter::ConvertEmptyGammaNode(const rvsdg::GammaNode & gammaNod
...
+        // 2. 1 -> 1, default 0
+        // There is no need to insert operations for the select predicate
+        selectPredicate = matchOrigin;
+        trueAlternative = Context_->GetVariable(output1);
+        falseAlternative = Context_->GetVariable(output0);
+      }
+      else
+      {
+        // FIXME: This will recreate the select predicate operations for each gamma output for
+        // which we create a select.
```
phate, line 311: Please ignore this for now. My plan is to move the entire select creation out of the RvsdgToIpGraph convert into its own transformation pass, and invoke as one of the last passes in the optimization pipeline. The entire IfConversion story of gamma nodes does not really belong in the RvsdgToIpGraph converter.

The FIXME will be handled once this is done.

jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
```diff
@@ -277,25 +278,58 @@ RvsdgToIpGraphConverter::ConvertEmptyGammaNode(const rvsdg::GammaNode & gammaNod
...
-      const auto matchOperation = util::AssertedCast<const rvsdg::match_op>(&node->GetOperation());
-      const auto defaultAlternative = matchOperation->default_alternative();
-      const auto condition = Context_->GetVariable(node->input(0)->origin());
-      const auto trueAlternative =
-          defaultAlternative == 0 ? Context_->GetVariable(output1) : Context_->GetVariable(output0);
-      const auto falseAlternative =
-          defaultAlternative == 0 ? Context_->GetVariable(output0) : Context_->GetVariable(output1);
+      const auto matchOperation =
+          util::AssertedCast<const rvsdg::match_op>(&matchNode->GetOperation());
+      assert(matchOperation->nalternatives() == 2);
```
haved, line 286: Is there a reason for using assert and not `JLM_ASSERT`?

jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
```diff
@@ -277,25 +278,58 @@ RvsdgToIpGraphConverter::ConvertEmptyGammaNode(const rvsdg::GammaNode & gammaNod
...
+
+      const auto matchOrigin = Context_->GetVariable(matchNode->input(0)->origin());
+      const auto caseValue = matchOperation->begin()->first;
+      const auto caseSubregion = matchOperation->begin()->second;
```
haved, line 290: At this point we have asserted that the match only has two alternatives, but does that necessarily mean that it only has one (value->alt) mapping, and one default alternative? If it has multiple mapping then we might not be able to convert it into a select, but has that case been determined to never occur before reaching this point in the code?

haved (commented):
Some quick comments for now, will look more

jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
phate: It should, but the input from the match operation is not really well checked (to put it nicely). We should improve this. I put another `JLM_ASSERT()` there that ensures that we really have only two cases. At least, it will scream at us when this is not the case.

phate:

haved (approved):
Looks good, I thought for a second I saw something more, but I did not consider that the match [1->0] default 1 is actually a bit more tricky than just swapping true and false (consider what happens if the input is 2)

PR #864, opened 2025-04-09
Add IfConversion transformation

phate:
This PR moves the if-conversion from the RvsdgToIpGraphConverter into its own RVSDG transformation.


haved (approved):

haved:
Looks good!

If a gamma has multiple outputs that can be converted, and the match needs to be converted to a IntEq operation, it will create one IntEq per output.

That will be fixed by CNE, however, so no problem. 

(@phate GitHub ate my review message, so here it is)

PR #865, opened 2025-04-11
Remove LoadVolatileNode class [AndersenAgnostic][AndersenRegionAware][SteensgaardAgnostic]

phate:


jlm/llvm/ir/operators/Load.hpp
```diff
@@ -24,12 +24,52 @@ namespace jlm::llvm
...
+        : rvsdg::input::iterator<rvsdg::SimpleInput>(input)
+    {}
+
+    [[nodiscard]] rvsdg::SimpleInput *
+    next() const override
+    {
+      auto index = value()->index();
+      auto node = value()->node();
+
+      return node->ninputs() > index + 1 ? node->input(index + 1) : nullptr;
```
haved, line 41: This > operator feels a bit yoda-like, I personally think it reads more naturally the other way around

jlm/llvm/ir/operators/Load.hpp
```diff
@@ -24,12 +24,52 @@ namespace jlm::llvm
...
+        : rvsdg::output::iterator<rvsdg::SimpleOutput>(output)
+    {}
+
+    [[nodiscard]] rvsdg::SimpleOutput *
+    next() const override
+    {
+      auto index = value()->index();
+      auto node = value()->node();
+
+      return node->noutputs() > index + 1 ? node->output(index + 1) : nullptr;
```
haved, line 58: Same here

haved (approved):
Very nice, just a personal preference nitpick

jlm/llvm/frontend/InterProceduralGraphConversion.cpp
```diff
@@ -541,10 +541,6 @@ ConvertThreeAddressCode(
   {
     Convert<CallNode, CallOperation>(threeAddressCode, region, variableMap);
   }
-  else if (is<LoadVolatileOperation>(&threeAddressCode))
```
haved, line 544: I don't quite understand why this is removed entirely, do we not need to convert the operation?

haved:
I also realized, the Andersen code looks for nodes that extend `LoadNode`. With this change it should instead look for nodes where the operation is a subclass of `LoadOperation`.

See: https://github.com/phate/jlm/blob/master/jlm%2Fllvm%2Fopt%2Falias-analyses%2FAndersen.cpp#L620

phate:
@haved That might explain why anderson stuff is failing.

jlm/llvm/frontend/InterProceduralGraphConversion.cpp
phate: Look at the else case. A normal SimpleNode with the LoadVolatileOperation is created instead of a dedicated node.

jlm/llvm/ir/operators/Load.hpp
phate: Fixed. Does this work for you?

jlm/llvm/ir/operators/Load.hpp
phate: Fixed. Does this work for you?

haved:
Steensgard has the same thing, I probably got it from there originally @phate

haved (approved):

phate:
@haved Well great, we just merged something with a bug.

phate:
@haved I suspect the problem originates from here: https://github.com/phate/jlm/blob/3f17f632ce3c5c2100dade00fccd0825b3d76005/jlm/llvm/opt/alias-analyses/RegionAwareModRefSummarizer.cpp#L767C1-L770C4

This should have also been changed. I will make it happen in the next PR.

haved:
Wait what? How did it even allow the merge if the check didn't pass?

@phate

phate:
The checks for agnostic/region-aware are not mandatory and I had auto-merge
enabled :-/

PR #866, opened 2025-04-12
Fix bug in RegionAwareModRefSummarizer [AndersenRegionAware]

phate:
The last PR removed LoadVolatileNodes, but forgot to change some lines in the RegionAwareModRefSummarizer. This should fix this problem.


haved (approved):

haved:
Nice find, @phate

Once we remove the `LoadNode` in its entirety, any remnants like these ones will become compile errors, which is nice

PR #873, opened 2025-04-16
Adds Verilator AXI harness

sjalander:
This harness is used when performing synthesis


haved (approved):
While im not fully sure how it is used, it looks reasonable, and there is not much duplication from the existing harness

PR #881, opened 2025-04-16
Handle loop constant buffer in DNE

sjalander:


jlm/hls/backend/rvsdg2rhls/rhls-dne.cpp
```diff
@@ -267,6 +269,159 @@ dead_loop(rvsdg::Node * ndmux_node)
   return true;
 }
 
+bool
+dead_loop_lcb(rvsdg::Node * lcb_node)
```
sjalander, line 273: Removes unused state edges that have been converted into loop buffers.

jlm/hls/backend/rvsdg2rhls/rhls-dne.cpp
```diff
@@ -267,6 +269,159 @@ dead_loop(rvsdg::Node * ndmux_node)
...
+  }
+  // divert users
+  branch_in->node()->output(0)->divert_users(lcb_node->input(1)->origin());
+  remove(branch_in->node());
+  remove(lcb_node);
+  return true;
+}
+
+bool
+fix_mem_split(rvsdg::Node * split_node)
```
sjalander, line 326: Combine multiple memory state split nodes into a single node to reduce HW.
Removes unused split outputs, which are a result from running fix_mem_merge first.

jlm/hls/backend/rvsdg2rhls/rhls-dne.cpp
```diff
@@ -267,6 +269,159 @@ dead_loop(rvsdg::Node * ndmux_node)
...
+    {
+      combined_outputs[i]->divert_users(new_outputs[i]);
+    }
+    return true;
+  }
+  return false;
+}
+
+bool
+fix_mem_merge(rvsdg::Node * merge_node)
```
sjalander, line 371: Removes merges that only have one state edge.
Replaces multiple parallel state edges with a single edge.
Combines multiple merge nodes into a single node.

jlm/hls/backend/rvsdg2rhls/rhls-dne.cpp
sjalander: Should likely be part of a node reduction pass insteast of DNE

jlm/hls/backend/rvsdg2rhls/rhls-dne.cpp
sjalander: Should likely be part of a node reduction pass instead of DNE

PR #882, opened 2025-04-16
Adds stream-conv pass

sjalander:


jlm/hls/backend/rvsdg2rhls/stream-conv.cpp
```diff
@@ -0,0 +1,102 @@
...
+    JLM_ASSERT(stream_deqs.empty());
+    return;
+  }
+  std::vector<rvsdg::SimpleNode *> enq_calls, deq_calls;
+  std::unordered_set<rvsdg::output *> visited;
+  for (auto stream_enq : stream_enqs)
+  {
+    JLM_ASSERT(stream_enq.inner);
+    trace_function_calls(stream_enq.inner, enq_calls, visited);
+    visited.erase(visited.begin(), visited.end());
```
haved, line 36: Use the `clear()` method instead of erasing using iterators

jlm/hls/backend/rvsdg2rhls/stream-conv.cpp
```diff
@@ -0,0 +1,102 @@
...
+  }
+  for (auto stream_deq : stream_deqs)
+  {
+    trace_function_calls(stream_deq.inner, deq_calls, visited);
+    visited.erase(visited.begin(), visited.end());
```
haved, line 41: Same here

jlm/hls/backend/rvsdg2rhls/stream-conv.cpp
```diff
@@ -0,0 +1,102 @@
...
+  JLM_ASSERT(!deq_calls.empty());
+  for (auto & enq_call : enq_calls)
+  {
+    auto enq_constant = trace_constant(enq_call->input(1)->origin());
+    for (auto & deq_call : deq_calls)
+    {
+      auto deq_constant = trace_constant(deq_call->input(1)->origin());
+      if (*enq_constant == *deq_constant)
+      {
+        int buffer_capacity = 10;
```
haved, line 53: This magic number should at the very least be moved to a global constant at the top of the file. It seems to only be a default if another size is not provided

jlm/hls/backend/rvsdg2rhls/stream-conv.cpp
```diff
@@ -0,0 +1,102 @@
...
+  }
+  JLM_ASSERT(!enq_calls.empty());
+  JLM_ASSERT(!deq_calls.empty());
+  for (auto & enq_call : enq_calls)
+  {
+    auto enq_constant = trace_constant(enq_call->input(1)->origin());
+    for (auto & deq_call : deq_calls)
+    {
+      auto deq_constant = trace_constant(deq_call->input(1)->origin());
+      if (*enq_constant == *deq_constant)
```
haved, line 51: This if is doing a lot of heavy lifting, could everything inside it be moved to a separate function that handles a specific (enq, deq) pair?

jlm/hls/backend/rvsdg2rhls/stream-conv.cpp
```diff
@@ -0,0 +1,102 @@
...
+        for (size_t i = 0; i < enq_call->ninputs(); ++i)
+        {
+          if (dynamic_cast<const rvsdg::StateType *>(&enq_call->input(i)->type()))
+          {
+            int oi = enq_call->noutputs() - enq_call->ninputs() + i;
+            enq_call->output(oi)->divert_users(enq_call->input(i)->origin());
+          }
+        }
+        remove(enq_call);
+        // remove deq_call from list
```
haved, line 84: It is not necessary to remove the deq_call, and modifying sets while iterating is bad form, even if we break immediately after

haved (changes requested):
The code makes sense, I have some suggestions for improvements @sjalander

jlm/hls/backend/rvsdg2rhls/stream-conv.cpp
sjalander: Moved to separate function

jlm/hls/backend/rvsdg2rhls/stream-conv.cpp
sjalander: Changed

jlm/hls/backend/rvsdg2rhls/stream-conv.cpp
sjalander: Changed

jlm/hls/backend/rvsdg2rhls/stream-conv.cpp
sjalander: Created constant

jlm/hls/backend/rvsdg2rhls/stream-conv.cpp
sjalander: The hls-test-suite fails if removed

haved (approved):

PR #883, opened 2025-04-16
Handle empty gammas in graph converter

sjalander:


phate (changes requested):
This seems to be the same code that I have just thrown out in PR #864. Why is this re-introduced again?

sjalander:
@phate This is probably my mistake where code that has been removed from master looks like it's a new addition for the DAE work.

PR #884, opened 2025-04-16
Modified loop inversion logic

sjalander:


phate (commented):
Why do we want to limit to transformation to only loops that originally where for-loops?

phate (changes requested):
Blocking merging until I have a better understanding.

sjalander:
@phate 
The current logic is too simple causing a significant code increase for a number of cases.
It in some cases causes the loop to be duplicated in each gamma region and with nested loops this can happen multiple times.

phate (approved):
Opened issue #893 to clean this eventually up. Approving for now such that it is not blocking.

@sjalander Could you point me to some examples where this pass acts up and duplicates too much code?

davidmetz:
I actually managed to find an example where TGI produces wrong behavior.
The examples I previously had just duplicated the loop into both gamma branches, but this is worse.
In this kernel a store happens in a loop - except in the last iteration.
The current TGI wrongly pulls out the gamma, since it shares the predicate with the theta.
```
void kernel(int *a) {
    int i = 0;
    do {
        i++;
        if (i != 10)
            a[i] = i;
    } while (i != 10);
}

int main(int argc, char **argv) {
    int a[11] = {0};
    kernel(a);
    assert(a[0] == 0);
    for (int i = 1; i < 10; ++i) {
        assert(a[i] == i);
    }
    assert(a[10] == 0);
    return 0;
}
```
before TGI:
image
after TGI:
image

sjalander:
@phate David has posted an example where the inversion fails.

phate:
@davidmetz @sjalander Thanks for reporting. I created issue #895 and will take a look at it.

PR #885, opened 2025-04-16
Pass arguments from jhls to jlm-hls

sjalander:


sjalander:
This PR depend on #886 being merged.

PR #886, opened 2025-04-16
Improved buffer insertion pass

sjalander:
This improvement significantly reduces the size of generated accelerators.


sjalander:
This PR will reduce once #879, #880, #881, and #912 is merged.
But it will remain being a large PR.

PR #888, opened 2025-04-23
Roundtrip graph imports

halvorlinder:
Requires https://github.com/EECS-NTNU/mlir_rvsdg/pull/9
Graph imports are reprsented by omega argument ops at the beginning of the omega block


sjalander:
@halvorlinder I've approved and merged the MLIR dialect PRs so you should be able to update the commit hash used for building MLIR.

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -37,20 +37,34 @@ MlirToJlmConverter::ReadAndConvertMlir(const util::filepath & filePath)
   {
     JLM_ASSERT("Parsing MLIR input file failed.");
   }
-  return ConvertMlir(block);
+  auto & topNode = block->front();
+  auto omegaNode = ::mlir::dyn_cast<::mlir::rvsdg::OmegaNode>(topNode);
+  return ConvertOmega(omegaNode);
```
haved, line 42: These four lines look like they do the same as ConvertMlir. Can it be called instead?

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -91,22 +94,55 @@ MlirToJlmConverter::ConvertBlock(::mlir::Block & block, rvsdg::Region & rvsdgReg
 
   // Create an RVSDG node for each MLIR operation and store each pair in a
   // hash map for easy lookup of corresponding RVSDG nodes
```
haved, line 96: This comment should be changed to reflect the hash map being on value-granularity, not operation granularity

haved (dismissed):
Looks great, just two small comments

haved (approved):

PR #889, opened 2025-04-23
Roundtrip new integer and float operations

halvorlinder:
Roundtrips a bunch of floating point and integer operations. 
Some integer operations are double implemented in jlm. The mlir frontend converts into the newer version of these operations

### DIFF
--- a/jlm/hls/backend/rvsdg2rhls/add-prints.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/add-prints.cpp
@@ -126,7 +126,7 @@ convert_prints(
         JLM_ASSERT(bt);
         val = &llvm::ZExtOperation::Create(*val, rvsdg::bittype::Create(64));
       }
-      llvm::CallNode::Create(printf_local, functionType, { constantNode.output(0), val });
+      llvm::CallOperation::Create(printf_local, functionType, { constantNode.output(0), val });
       node->output(0)->divert_users(node->input(0)->origin());
       jlm::rvsdg::remove(node);
     }
--- a/jlm/hls/backend/rvsdg2rhls/instrument-ref.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/instrument-ref.cpp
@@ -179,7 +179,7 @@ instrument_ref(
         addr = jlm::llvm::bitcast_op::create(addr, void_ptr);
       }
       auto memstate = node->input(1)->origin();
-      auto callOp = jlm::llvm::CallNode::Create(
+      auto callOp = jlm::llvm::CallOperation::Create(
           load_func,
           loadFunctionType,
           { addr, widthNode.output(0), ioState, memstate });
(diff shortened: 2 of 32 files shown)


jlm/mlir/backend/JlmToMlirConverter.cpp
```diff
@@ -62,17 +63,13 @@ JlmToMlirConverter::Print(::mlir::rvsdg::OmegaNode & omega, const util::filepath
 ::mlir::rvsdg::OmegaNode
 JlmToMlirConverter::ConvertModule(const llvm::RvsdgModule & rvsdgModule)
 {
-  return ConvertOmega(rvsdgModule.Rvsdg());
```
sjalander, line 65: Need to get https://github.com/phate/jlm/pull/888 merged before merging this one

jlm/mlir/backend/JlmToMlirConverter.hpp
```diff
@@ -116,6 +117,23 @@ class JlmToMlirConverter final
   ::mlir::Operation *
   ConvertFpBinaryNode(const jlm::llvm::fpbin_op & op, ::llvm::SmallVector<::mlir::Value> inputs);
 
+  /**
+   * Converts an fpcmp an mlir::arith::CmpFPredicate.
```
haved, line 121: Typo here, missing a "to"

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -237,106 +270,160 @@ MlirToJlmConverter::ConvertFPBinaryNode(
       { inputs[0], inputs[1] });
 }
 
+llvm::fpcmp
+MlirToJlmConverter::ConvertFPCMP(const ::mlir::arith::CmpFPredicate & op)
```
haved, line 274: It would be nice to combine this with the converter that goes the opposite direction, and use a `util::BijectiveMap` that is used in both directions. The common map and converter functions could be placed in a `MlirConverterCommon.[ch]pp` placed outside both the fronted and backend folder

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -237,106 +270,160 @@ MlirToJlmConverter::ConvertFPBinaryNode(
...
+    return (llvm::fpcmp::une);
+  case ::mlir::arith::CmpFPredicate::UNO:
+    return (llvm::fpcmp::uno);
+  default:
+    JLM_UNREACHABLE("Unknown fp comp");
+  }
+}
+
 rvsdg::Node *
 MlirToJlmConverter::ConvertBitBinaryNode(
```
haved, line 316: Since this function may fail, I think it should be named `TryConvertBitBinaryNode`

haved (changes requested):
Looks very nice, I really like the use of templates and macros to make the tests extensive. I have a few small comments

jlm/mlir/frontend/MlirToJlmConverter.cpp
haved: In the new cpp file you make

PR #896, opened 2025-04-26
Make DeadNodeEliminiation extensible

phate:
We currently have two DNE transformations in the code base: one for LLVM dialect and one for HLS dialect. The reason why we duplicated them in the first place was because the HLS dialect required to handle additional structural nodes (loop node) that was not handle in the LLVM DNE.

This PR introduces a way to extend DNE for custom structural nodes such that the two transformations can be unified and we can remove the duplicated code. This PR  is meant as a POC and only shows the idea for the gamma node. For the other node, it would be possible to refactor it in a similar fashion. However, I would like to get some feedback to ensure that this is on the right track before I invest the time to convert the rest of the transformation.

### DIFF
--- a/jlm/hls/backend/rvsdg2rhls/rvsdg2rhls.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/rvsdg2rhls.cpp
@@ -58,7 +58,7 @@ void
 split_opt(llvm::RvsdgModule & rm)
 {
   // TODO: figure out which optimizations to use here
-  jlm::llvm::DeadNodeElimination dne;
+  jlm::llvm::DeadNodeElimination dne({ llvm::DNEGammaNodeHandler::GetInstance() });
   jlm::hls::cne cne;
   jlm::llvm::InvariantValueRedirection ivr;
   jlm::llvm::tginversion tgi;
@@ -76,7 +76,7 @@ void
 pre_opt(jlm::llvm::RvsdgModule & rm)
 {
   // TODO: figure out which optimizations to use here
-  jlm::llvm::DeadNodeElimination dne;
+  jlm::llvm::DeadNodeElimination dne({ llvm::DNEGammaNodeHandler::GetInstance() });
   jlm::hls::cne cne;
   jlm::llvm::InvariantValueRedirection ivr;
   jlm::llvm::tginversion tgi;
@@ -438,7 +438,7 @@ rvsdg2rhls(llvm::RvsdgModule & rhls, util::StatisticsCollector & collector)
 
   merge_gamma(rhls);
 
-  llvm::DeadNodeElimination llvmDne;
+  llvm::DeadNodeElimination llvmDne({ llvm::DNEGammaNodeHandler::GetInstance() });
   llvmDne.Run(rhls, collector);
 
   mem_sep_argument(rhls);
--- a/jlm/llvm/opt/DeadNodeElimination.cpp
+++ b/jlm/llvm/opt/DeadNodeElimination.cpp
@@ -163,9 +163,15 @@ public:
 
 DeadNodeElimination::~DeadNodeElimination() noexcept = default;
 
-DeadNodeElimination::DeadNodeElimination(std::vector<const DNEStructuralNodeHandler *> handlers)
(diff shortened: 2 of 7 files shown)


caleridas:
> @haved @caleridas @sjalander Could you provide me with some feedback such that I know whether this goes in the right direction, please?

That might be a necessity, but TBH the separate type of loop is really terrible to begin with. There is almost no point to it as everything hls_loop needs, Theta already does -- hls_loop is a "constrained subset" of normal theta capabilities, I would really like if users of hls_loop could look into formulating it this way.

sjalander:
@caleridas I don't fully understand hls_loop being a subset of theta.

The loop explicitly models the backedge with its own result and argument. So each loop variable has 1 input, 2 arguments, 2 results, and 1 output. This is such that in hardware one can select between the input and the backedge with a multiplexer. And between backedge and output.

Is this possible to model with a version of the existing theta?

caleridas:
> @caleridas I don't fully understand hls_loop being a subset of theta.
> 
> The loop explicitly models the backedge with its own result and argument. So each loop variable has 1 input, 2 arguments, 2 results, and 1 output. This is such that in hardware one can select between the input and the backedge with a multiplexer. And between backedge and output.
> 
> Is this possible to model with a version of the existing theta?

Yes, in the following way: for each "logical" loop variable X, create two theta loop variables I and B.
- food the input into I, feed "undefined" or "zero" into B before loop entry
- at beginning of loop iteration, demultiplex between I and B as you do now
- at end of loop iteration feed the output into B, feed I into itself (this is "passthrough")
- after loop exit, take the output from B, ignore/discard I

so I effectively is a "loop invariant constant". I had already entertained the thought whether we want to explicitly model loop invariant variables separate from loop-carried inputs, but may that's a story for another time.

The big primary "issue" is that the structure of loop input / loop carried / loop output variables conforming to the above must either be preserved by all transformation steps, or there must be some facility to restore this structure. Neither is conceptually really hard, but I admit that it is not terribly joyful work to iron out all corner cases. One conceptually good way to do it would be to treat such "special form" requirements as additional data to the ThetaNode, hooking this up correctly would in particularly allow to detect any constraint violation instantly at each transformation step. On the plus side, going this route probably allows to immediately leverage all loop transformations across backends.

I am not opposed to parameterizing DNE in a way such that it can be consolidated today already as the above is non-negligible work. This is not mutually exclusive with reworking the loop node longer-term, but I think it is worth keeping this in mind going forward as an option at least.

haved:
@caleridas The suggestions you bring up are things I have considered as well, but never quite come up with a nice way of representing.

For example, it is typical for a Theta to have a bunch of results that are always `Undef`, except for when the predicate is false and the loop is exiting. These become region arguments that should not be used. Conversely there are results that are always `Undef` when the loop is actually exiting, which leads to Theta output that should never be used.
I am not sure if this is actually a problem when writing compiler passes, but as a human looking at the graph it is a bit extra work to remember which theta results are actually "only" exit variables, and which are "only" loop variables.

I do not know how one would go about changing that, however, without doing large changes. If one were to introduce separate context variable, loop variable and exit variable, it would break all the symmetry between indices, which I am not sure it is worth?

One final question I have is regarding the special `hls_loop`. It handles memory port routing very differently to traditional loop variables. The arguments can produce multiple memory responses each iteration, and the results can consume multiple memory requests per iteration. Neither port is synced up with the actual looping of the theta. Is this also representable using a regular Theta node? Outside the realm of hls, are there other situations where we would like to represent such "sideways" flow of data in and out of regions?

phate:
@haved @caleridas @sjalander This spiraled a little bit out of control. The discussion is less about the current approach, but whether it should be taken in the first place due to the HLS loop. This is definitely worth a discussion, but HLS loop won't go anywhere any time soon. Thus, unless any one of still has any objections to the approach, I would just move ahead and produce a proper PR. @caleridas already sanctioned the approach. Any objections from your side @haved and @sjalander ?

sjalander:
@phate 
This optimization lives in llvm, but is intended to be made generic such that it can be applied to both llvm and hls.
Would it make sense to have the generic part live in jlm/opt or jlm/rvsdg/opt instead?

PR #897, opened 2025-04-27
Introduce UnitType

caleridas:
Introduce the UnitType to allow representing a type that holds no data (akin to the C "void" type). This can formally be used in places where "something" must be produced, but it has no content -- this will later be used in pattern matching over types without a content in generalized gamma constructs.

Prepare places that currently balk on existency of this type so its usage can be introduced without breakage later.

### DIFF
--- a/jlm/llvm/backend/dot/DotWriter.cpp
+++ b/jlm/llvm/backend/dot/DotWriter.cpp
@@ -12,6 +12,7 @@
 #include <jlm/rvsdg/structural-node.hpp>
 #include <jlm/rvsdg/traverser.hpp>
 #include <jlm/rvsdg/type.hpp>
+#include <jlm/rvsdg/UnitType.hpp>
 
 namespace jlm::llvm::dot
 {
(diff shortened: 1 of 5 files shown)


haved (approved):
I think the code looks good. I guess it is up to individual frontends and backends to decide what to do with the type.

How would one for example create a value of the unit type? In C and C++ I'm not sure if that is something you ever do explicitly, while In rust you can write `()`. If that actually translates into a unit type in LLVM or is completely compiled out before it reaches the IR, I do not know.

@caleridas

caleridas:
> I think the code looks good. I guess it is up to individual frontends and backends to decide what to do with the type.
> 
> How would one for example create a value of the unit type? In C and C++ I'm not sure if that is something you ever do explicitly, while In rust you can write `()`. If that actually translates into a unit type in LLVM or is completely compiled out before it reaches the IR, I do not know.
> 
> @caleridas

The place where I will introduce this first is in extending gamma to allow pattern matching + destructuring -- but if you feed in effectively only "bools" (as the current llvm frontend does), then they destructure into something empty which is the unit type, and that then also disappears in translation (it is, after all, "dead" since nobody uses its value).

If some need to translate it arises, it would be correct to map it to "void" or something like an empty struct (as "struct {}"), and everything will work.

PR #898, opened 2025-04-27
GammaNode: EntryVar/ExitVar removal API, and consistent usage of safe…

caleridas:
Add methods that allow safe removal of entryvars / exitvars, and generally make use of the safe EntryVar/ExitVar API in all places that need to map between inputs/arguments and results/outputs. This is in preparation to changes how gamma arguments are allocated when generalizing gamma to support pattern matching.

### DIFF
--- a/jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
@@ -129,48 +129,53 @@ RemovePassthroughArgument(const rvsdg::RegionArgument & argument)
   region->node()->RemoveOutput(outputIndex);
 }
 
+// If this output has a single user and that single user happens to be
+// the exit variable of this gamma node, then return it.
+static std::optional<rvsdg::GammaNode::ExitVar>
+TryGetSingleUserExitVar(rvsdg::GammaNode & gammaNode, rvsdg::output & argument)
+{
+  if (argument.nusers() == 1)
+  {
+    rvsdg::input * user = *argument.begin();
+    if (rvsdg::TryGetRegionParentNode<rvsdg::GammaNode>(*user) == &gammaNode)
+    {
+      return gammaNode.MapBranchResultExitVar(*user);
+    }
+  }
+  return std::nullopt;
+}
+
 static void
 RemoveUnusedStatesFromGammaNode(rvsdg::GammaNode & gammaNode)
 {
-  auto entryvars = gammaNode.GetEntryVars();
-  for (int i = entryvars.size() - 1; i >= 0; --i)
-  {
-    size_t resultIndex = 0;
-    auto argument = entryvars[i].branchArgument[0];
-    if (argument->nusers() == 1)
-    {
-      auto result = dynamic_cast<rvsdg::RegionResult *>(*argument->begin());
-      resultIndex = result ? result->index() : resultIndex;
-    }
+  std::vector<rvsdg::GammaNode::EntryVar> deadEntryVars;
+  std::vector<rvsdg::GammaNode::ExitVar> deadExitVars;
 
-    bool shouldRemove = true;
-    for (size_t n = 0; n < gammaNode.nsubregions(); n++)
-    {
-      auto subregion = gammaNode.subregion(n);
-      shouldRemove &=
-          IsPassthroughArgument(*subregion->argument(i))
-          && dynamic_cast<jlm::rvsdg::RegionResult *>(*subregion->argument(i)->begin())->index()
-                 == resultIndex;
-    }
+  for (const auto & entryvar : gammaNode.GetEntryVars())
+  {
+    std::optional<rvsdg::GammaNode::ExitVar> exitvar0 =
+        TryGetSingleUserExitVar(gammaNode, *entryvar.branchArgument[0]);
+
+    bool shouldRemove = exitvar0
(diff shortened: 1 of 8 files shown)


jlm/rvsdg/gamma.cpp
```diff
@@ -344,6 +346,56 @@ GammaNode::MapBranchResultExitVar(const rvsdg::input & input) const
...
+        return x > y;
+      });
+  for (std::size_t index : indices)
+  {
+    for (std::size_t r = 0; r < nsubregions(); ++r)
+    {
+      subregion(r)->RemoveResult(index);
+    }
+    RemoveOutput(index);
+  }
```
haved, line 371: I understand that the indices are sorted to avoid shifting that invalidates later indices, but we could also avoid the O(N*M) removal here, and replace it with O(N+M), if we do a single sweep.

jlm/rvsdg/gamma.cpp
```diff
@@ -344,6 +346,56 @@ GammaNode::MapBranchResultExitVar(const rvsdg::input & input) const
...
+}
+
+void
+GammaNode::RemoveEntryVars(const std::vector<EntryVar> & entryvars)
+{
+  std::vector<std::size_t> indices;
+  for (const auto & entryvar : entryvars)
+  {
+    indices.push_back(entryvar.input->index());
```
haved, line 380: Same here, `JLM_ASSERT` would be nice

haved (commented):
Looks very nice, I have some small suggestions and questions

jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
```diff
@@ -129,48 +129,53 @@ RemovePassthroughArgument(const rvsdg::RegionArgument & argument)
...
-      }
-      gammaNode.RemoveInput(i + 1);
+      exitvar0->output->divert_users(entryvar.input->origin());
+      deadEntryVars.push_back(entryvar);
+      deadExitVars.push_back(*exitvar0);
     }
   }
+
+  gammaNode.RemoveExitVars(deadExitVars);
+  gammaNode.RemoveEntryVars(deadEntryVars);
```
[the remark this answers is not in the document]
caleridas: I think we should at some point look into consolidating the transforms, and also moving them towards common code (I think they are generically applicable, but may need minor tweaks), then this should definitely be revisited. For now, I primarily want to get these abstractions in so I have more libertry changing structures.

jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
```diff
@@ -261,15 +261,18 @@ RvsdgToIpGraphConverter::ConvertGammaNode(const rvsdg::GammaNode & gammaNode)
...
   std::vector<cfg_node *> phi_nodes;
   entryBlock->append_last(BranchOperation::create(numSubregions, Context_->GetVariable(predicate)));
+  auto entryvars = gammaNode.GetEntryVars();
   for (size_t n = 0; n < gammaNode.nsubregions(); n++)
   {
     const auto subregion = gammaNode.subregion(n);
 
     // add arguments to context
-    for (size_t i = 0; i < subregion->narguments(); i++)
+    for (std::size_t i = 0; i < entryvars.size(); ++i)
```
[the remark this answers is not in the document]
caleridas: I default to std::size_t for C++ as opposed to C -- but simply reverting this, it doesn't matter.

jlm/rvsdg/gamma.cpp
caleridas: I think it needs to be O(#vars_removed * #regions), that's what the inner loop amounts to. So I am not entirely certain what you mean by "single sweep" here?

jlm/rvsdg/gamma.cpp
```diff
@@ -344,6 +346,56 @@ GammaNode::MapBranchResultExitVar(const rvsdg::input & input) const
   return ExitVar{ std::move(branchResults), Node::output(input.index()) };
 }
 
+void
+GammaNode::RemoveExitVars(const std::vector<ExitVar> & exitvars)
+{
+  std::vector<std::size_t> indices;
+  for (const auto & exitvar : exitvars)
+  {
+    indices.push_back(exitvar.output->index());
```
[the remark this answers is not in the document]
caleridas: added assert, as well as ensuring uniqueness (just to be sure)

jlm/rvsdg/gamma.cpp
caleridas: same as above, added check

haved (approved):
Looks good. My comment about quadratic removal time has nothing to do with the number of regions, but is in the case when you are removing multiple variables, and have to shift the indices of all later variables by one. If you are removing 100 variables, followed by 100 variables you do not remove, then the number of move operations becomes 100*100. That might not actually be a problem in practice, though.

@caleridas

caleridas:
> Looks good. My comment about quadratic removal time has nothing to do with the number of regions, but is in the case when you are removing multiple variables, and have to shift the indices of all later variables by one. If you are removing 100 variables, followed by 100 variables you do not remove, then the number of move operations becomes 100*100. That might not actually be a problem in practice, though.
> 
> @caleridas

I see -- yes that is true, but is inherent to the "single port" removal interface currently offered by Node/Region. Agree to address this eventually, but it is in a different place.

PR #900, opened 2025-04-30
GammaNode: Add pattern matching support to gamma

caleridas:
Add an argument to gamma branch subregions that represents the "content" of the discriminator object that gamma is matching over. For booleans/control decisions, they do not have a content (although one could argue that they carry an "assertion" that a certain proposition holds that in turn some operations in the contained region may be dependent on -- e.g. a divisor not being zero, a pointer being non-null etc.) -- so use UnitType here by default.

This allows Gamma to serve as general pattern matching operator which makes the contents of e.g. an algebraic type being matched over available in its sub branches.

Adapt all places using Gamma API to deal with the fact that there is another argument now not corresponding to any EntryVar.

### DIFF
--- a/jlm/rvsdg/gamma.cpp
+++ b/jlm/rvsdg/gamma.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010 2011 2012 2013 2014 2025 Helge Bahmann <hcb@chaoticmind.net>
+ * Copyright 2010 2011 2012 2013 2014 Helge Bahmann <hcb@chaoticmind.net>
  * Copyright 2013 2014 2015 2016 Nico Reißmann <nico.reissmann@gmail.com>
  * See COPYING for terms of redistribution.
  */
@@ -9,6 +9,7 @@
 #include <jlm/rvsdg/control.hpp>
 #include <jlm/rvsdg/gamma.hpp>
 #include <jlm/rvsdg/substitution.hpp>
+#include <jlm/rvsdg/UnitType.hpp>
 
 namespace jlm::rvsdg
 {
@@ -208,19 +209,33 @@ bool
 GammaOperation::operator==(const Operation & other) const noexcept
 {
   auto op = dynamic_cast<const GammaOperation *>(&other);
-  return op && op->nalternatives_ == nalternatives_;
+  return op && op->numAlternatives_ == numAlternatives_;
+}
+
+std::shared_ptr<const Type>
+GammaOperation::GetMatchContentType(std::size_t alternative) const
+{
+  return alternative < MatchContentTypes_.size() ? MatchContentTypes_[alternative]
+                                                 : UnitType::Create();
 }
 
 /* gamma node */
 
 GammaNode::~GammaNode() noexcept = default;
 
-GammaNode::GammaNode(rvsdg::output * predicate, size_t nalternatives)
+GammaNode::GammaNode(
+    rvsdg::output * predicate,
+    size_t nalternatives,
+    std::vector<std::shared_ptr<const Type>> match_content_types)
     : StructuralNode(predicate->region(), nalternatives),
-      Operation_(nalternatives)
+      Operation_(nalternatives, std::move(match_content_types))
 {
   add_input(std::unique_ptr<node_input>(
       new StructuralInput(this, predicate, ControlType::Create(nalternatives))));
+  for (std::size_t n = 0; n < nalternatives; ++n)
+  {
+    RegionArgument::Create(*subregion(n), nullptr, Operation_.GetMatchContentType(n));
+  }
 }
 
 [[nodiscard]] const GammaOperation &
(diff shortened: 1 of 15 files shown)


jlm/rvsdg/gamma.hpp
```diff
@@ -23,9 +23,16 @@ class GammaOperation final : public StructuralOperation
...
   ~GammaOperation() noexcept override;
 
-  explicit constexpr GammaOperation(size_t nalternatives) noexcept
+  explicit GammaOperation(size_t nalternatives) noexcept
+      : GammaOperation(nalternatives, {})
+  {}
+
+  GammaOperation(
+      size_t nalternatives,
+      std::vector<std::shared_ptr<const Type>> match_content_types) noexcept
```
phate, line 32: camelCase

jlm/rvsdg/gamma.hpp
```diff
@@ -23,9 +23,16 @@ class GammaOperation final : public StructuralOperation
 public:
   ~GammaOperation() noexcept override;
 
-  explicit constexpr GammaOperation(size_t nalternatives) noexcept
+  explicit GammaOperation(size_t nalternatives) noexcept
```
phate, line 26: We generally use `num...` instead of `n...`. Thus, it should be `numAlternatives`

jlm/rvsdg/gamma.hpp
```diff
@@ -23,9 +23,16 @@ class GammaOperation final : public StructuralOperation
...
+      : GammaOperation(nalternatives, {})
+  {}
+
+  GammaOperation(
+      size_t nalternatives,
```
phate, line 31: `numAlternatives`

jlm/rvsdg/gamma.hpp
```diff
@@ -43,21 +50,80 @@ class GammaOperation final : public StructuralOperation
...
   operator==(const Operation & other) const noexcept override;
 
+  /**
+   * \brief Returns the type of pattern matching content produced in a branch
+   */
+  std::shared_ptr<const Type>
+  GetMatchContentType(std::size_t alternative) const;
+
 private:
   size_t nalternatives_;
```
phate, line 60: Maybe you can change that name as well once you are at it.

jlm/rvsdg/gamma.hpp
```diff
@@ -43,21 +50,80 @@ class GammaOperation final : public StructuralOperation
...
+  std::vector<std::shared_ptr<const Type>> MatchContentTypes_;
 };
 
 /* gamma node */
```
phate, line 64: Could be removed

jlm/rvsdg/gamma.hpp
```diff
@@ -43,21 +50,80 @@ class GammaOperation final : public StructuralOperation
...
 {
 public:
   ~GammaNode() noexcept override;
 
 private:
-  GammaNode(rvsdg::output * predicate, size_t nalternatives);
+  GammaNode(
+      rvsdg::output * predicate,
+      size_t nalternatives,
+      std::vector<std::shared_ptr<const Type>> match_content_types);
```
phate, line 109: camelCase

jlm/rvsdg/gamma.hpp
```diff
@@ -94,7 +160,16 @@ class GammaNode : public StructuralNode
...
   {
-    return new GammaNode(predicate, nalternatives);
+    return new GammaNode(predicate, nalternatives, {});
+  }
+
+  static GammaNode &
+  Create(
+      jlm::rvsdg::output * predicate,
+      size_t nalternatives,
+      std::vector<std::shared_ptr<const Type>> match_content_types)
```
phate, line 170: camelCase

jlm/rvsdg/gamma.hpp
```diff
@@ -94,7 +160,16 @@ class GammaNode : public StructuralNode
...
   create(jlm::rvsdg::output * predicate, size_t nalternatives)
   {
-    return new GammaNode(predicate, nalternatives);
+    return new GammaNode(predicate, nalternatives, {});
+  }
+
+  static GammaNode &
+  Create(
+      jlm::rvsdg::output * predicate,
+      size_t nalternatives,
```
phate, line 169: numAlternatives

jlm/rvsdg/gamma.cpp
```diff
@@ -211,16 +212,33 @@ GammaOperation::operator==(const Operation & other) const noexcept
...
-      Operation_(nalternatives)
+      Operation_(nalternatives, std::move(match_content_types))
 {
   add_input(std::unique_ptr<node_input>(
       new StructuralInput(this, predicate, ControlType::Create(nalternatives))));
+  for (std::size_t n = 0; n < nalternatives; ++n)
+  {
+    RegionArgument::Create(
+        *subregion(n),
+        nullptr /* input(0) */,
```
phate, line 239: Use `//`

phate (dismissed):
I found only minor stuff

jlm/rvsdg/gamma.hpp
caleridas: done

jlm/rvsdg/gamma.hpp
caleridas: done

jlm/rvsdg/gamma.hpp
caleridas: done

jlm/rvsdg/gamma.hpp
caleridas: renamed

jlm/rvsdg/gamma.hpp
caleridas: done

jlm/rvsdg/gamma.cpp
caleridas: removed superfluous comment

phate (approved):

PR #901, opened 2025-04-30
Updated RemoveUnusedStatesFromGammaNode

sjalander:


phate (changes requested):
This code was just changed in PR #898. Is this a mistake or is there a serious reason to reverse this again?

jlm/hls/backend/rvsdg2rhls/UnusedStateRemoval.cpp
```diff
@@ -129,53 +129,48 @@ RemovePassthroughArgument(const rvsdg::RegionArgument & argument)
...
-                            {
-                              auto exitvar = TryGetSingleUserExitVar(gammaNode, *argument);
-                              return exitvar && exitvar->output == exitvar0->output;
-                            });
+    bool shouldRemove = true;
+    for (size_t n = 0; n < gammaNode.nsubregions(); n++)
+    {
+      auto subregion = gammaNode.subregion(n);
+      shouldRemove &=
+          IsPassthroughArgument(*subregion->argument(i))
```
haved, line 151: Should `entryvars.branchArguments[n]` be used instead of looking up the argument from the subregion?

haved (commented):
The change makes sense, one comment for now @sjalander 

Edit: I saw Nico's comment now, I didn't really look at the red, only the green

sjalander:
I guess I have too many moving parts so I'm making mistakes.

PR #902, opened 2025-04-30
Minor refactoring of add-sinks

sjalander:


jlm/hls/backend/rvsdg2rhls/add-sinks.cpp
```diff
@@ -3,6 +3,7 @@
  * See COPYING for terms of redistribution.
  */
 
+#include "jlm/rvsdg/lambda.hpp"
```
phate, line 6: Please use `<>`

phate (changes requested):

jlm/hls/backend/rvsdg2rhls/add-sinks.cpp
```diff
@@ -47,7 +48,8 @@ add_sinks(llvm::RvsdgModule & rm)
 {
   auto & graph = rm.Rvsdg();
   auto root = &graph.GetRootRegion();
-  add_sinks(root);
+  auto lambda = dynamic_cast<rvsdg::LambdaNode *>(root->Nodes().begin().ptr());
+  add_sinks(lambda->subregion());
```
phate, line 52: Let's use an asserted cast here.

jlm/hls/backend/rvsdg2rhls/add-sinks.cpp
```diff
@@ -47,7 +48,8 @@ add_sinks(llvm::RvsdgModule & rm)
 {
   auto & graph = rm.Rvsdg();
   auto root = &graph.GetRootRegion();
-  add_sinks(root);
+  auto lambda = dynamic_cast<rvsdg::LambdaNode *>(root->Nodes().begin().ptr());
```
phate, line 51: Maybe also add an assert that the root region really only contains a single node.

jlm/hls/backend/rvsdg2rhls/add-sinks.cpp
sjalander: Changed

jlm/hls/backend/rvsdg2rhls/add-sinks.cpp
sjalander: Added

jlm/hls/backend/rvsdg2rhls/add-sinks.cpp
sjalander: Changed

phate (approved):

PR #904, opened 2025-05-01
Removal outdated input::GetNode() method [SteensgaardAgnostic][AndersenAgnostic][AndersenRegionAware]

phate:


haved (approved):
Looks good, I was wondering if we could change `TryGetOwnerNode<rvsdg::Node>` to just `TryGetOwnerNode`, since `rvsdg::Node` is a sane default.

It does make the code a little less explicit, however, so it is a matter of taste.

phate:
@haved Yeah, we can do a specialization for it I guess. I would make the code a bit less verbose.

haved:
I didn't think a specialization, just a default type for the template argument @phate

haved (approved):

phate:
@haved Ah, I see. It is a feature I have not used very much. I will open another PR once this is merged.

PR #906, opened 2025-05-01
Introduce default template type to (Try-/Assert-)GetOwnerNode() methods

phate:


haved (approved):
Very nice!

caleridas (commented):
I am not sure - the places just asking for Node are IMHO generally not doing "the right thing" anyways: if you need to interpret the origin of a value, then you will also need to distinguish by specific kind of node. I suspect that using unadorned Node is more of a relic than anything else and wonder whether it should really be treated as the "default" case.

phate:
@calerida I see your point. The only reason I could see is that  we would want to check whether the input/output belong to a node and not a region, but then it would also be possible to just call GetOwner(). So I take it from your comment that you are not in favor of merging this?

@haved Any opinions?

haved:
@phate @caleridas I see the point as well. We currently have a "pattern" where a node is retrieved only to get its operation. In such cases, we do not need the SimpleNode, since all nodes have operations. This pattern should probably be handled in a cleaner way, however, so I can agree that simply getting the `Node` is often an indication of an anti-pattern, and can be left explicit in the few cases it is actually needed

phate:
Closed. See discussion.

PR #909, opened 2025-05-01
Remove type() method of output class

phate:


jlm/hls/backend/rhls2firrtl/RhlsToFirrtlConverter.cpp
```diff
@@ -1633,11 +1633,11 @@ RhlsToFirrtlConverter::MlirGenMem(const jlm::rvsdg::SimpleNode * node)
     Connect(body, memReqWrite, zeroBitValue);
     auto invalid = GetInvalid(body, 32);
     Connect(body, memReqData, invalid);
-    if (auto bitType = dynamic_cast<const jlm::rvsdg::bittype *>(&node->output(0)->type()))
+    if (auto bitType = std::dynamic_pointer_cast<const rvsdg::bittype>(node->output(0)->Type()))
     {
       bitWidth = bitType->nbits();
     }
-    else if (dynamic_cast<const llvm::PointerType *>(&node->output(0)->type()))
+    else if (rvsdg::is<const llvm::PointerType>(node->output(0)->Type()))
```
haved, line 1640: Does `rvsdg::is` need the `const` part?

haved (dismissed):
Looks good, just one question about `rvsdg::is` @phate

jlm/hls/backend/rhls2firrtl/RhlsToFirrtlConverter.cpp
phate: No, it does not need it. I went through the code again and removed from all places where it was for the is() method.

haved (approved):

PR #912, opened 2025-05-03
Adds HLS NodeReduction transformation

sjalander:
The tronsformation tries to reduce multiple memory state split or multiple merge nodes into a single node.
It also reduces multiple parallel memory state edges into a single state edge.


phate:
@sjalander The reductions on the MemoryStateSplitOperation and the MemoryStateMergeOperation should go into the llvm namespace as this has nothing really to do with HLS. The bits of code that handle the loop node stuff should eventually end up in the real DNE. The rest can be removed as far as I can see.

phate (changes requested):
Blocking this PR as it should not be merged. The code for the memory state split and memory state merge operations should go into the LLVM namespace. There it should be implemented as a proper node reduction and not as its own pass.

PR #913, opened 2025-05-04
Prefer TryGetOwnerNode instead of casting input / output types

caleridas:
Some places in code still cast input / output to node_input / node_output in order to determine ownership. Replace this with the TryGetOwnerNode API.

### DIFF
--- a/tests/jlm/mlir/frontend/TestMlirToJlmConverter.cpp
+++ b/tests/jlm/mlir/frontend/TestMlirToJlmConverter.cpp
@@ -290,11 +290,9 @@ TestDivOperation()
       assert(std::dynamic_pointer_cast<const bittype>(DivInput0->Type())->nbits() == 32);
 
       // Check second input
-      jlm::rvsdg::node_output * DivInput1NodeOuput;
-      assert(
-          DivInput1NodeOuput =
-              dynamic_cast<jlm::rvsdg::node_output *>(lambdaResultOriginNode->input(1)->origin()));
-      Node * DivInput1Node = DivInput1NodeOuput->node();
+      auto DivInput1Node = jlm::rvsdg::TryGetOwnerNode<jlm::rvsdg::SimpleNode>(
+          *lambdaResultOriginNode->input(1)->origin());
+      assert(DivInput1Node);
       assert(is<jlm::llvm::IntegerConstantOperation>(DivInput1Node->GetOperation()));
       auto DivInput1Constant =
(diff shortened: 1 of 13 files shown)


tests/jlm/mlir/frontend/TestMlirToJlmConverter.cpp
```diff
@@ -450,11 +448,9 @@ TestCompZeroExt()
...
       // Traverse the rvsgd graph upwards to check connections
       std::cout << "Testing lambdaResultOriginNodeOuput\n";
-      jlm::rvsdg::node_output * lambdaResultOriginNodeOuput;
-      assert(
-          lambdaResultOriginNodeOuput = dynamic_cast<jlm::rvsdg::node_output *>(
-              convertedLambda->subregion()->result(0)->origin()));
-      Node * ZExtNode = lambdaResultOriginNodeOuput->node();
+      auto ZExtNode = jlm::rvsdg::TryGetOwnerNode<jlm::rvsdg::SimpleNode>(
+          *convertedLambda->subregion()->result(0)->origin());
+      assert(ZExtNode);
```
phate, line 453: The assert() is not really necessary here. The pointer is used further down.

tests/jlm/mlir/frontend/TestMlirToJlmConverter.cpp
```diff
@@ -290,11 +290,9 @@ TestDivOperation()
...
 
       // Check second input
-      jlm::rvsdg::node_output * DivInput1NodeOuput;
-      assert(
-          DivInput1NodeOuput =
-              dynamic_cast<jlm::rvsdg::node_output *>(lambdaResultOriginNode->input(1)->origin()));
-      Node * DivInput1Node = DivInput1NodeOuput->node();
+      auto DivInput1Node = jlm::rvsdg::TryGetOwnerNode<jlm::rvsdg::SimpleNode>(
+          *lambdaResultOriginNode->input(1)->origin());
+      assert(DivInput1Node);
```
phate, line 295: same here

phate (approved):

tests/jlm/mlir/frontend/TestMlirToJlmConverter.cpp
caleridas: I think the code was asserting before (in slightly different ways), and I thought meaningful assert failure instead of default does not harm in any case.

PR #915, opened 2025-05-05
Add TryGetSimpleNodeAndOp() methods

phate:
A lot of transformation in the code simply have to check whether the owners of inputs/outputs are SimpleNode of a specific operation type. Those methods simplify these checks.

### DIFF
--- a/jlm/hls/backend/rvsdg2rhls/decouple-mem-state.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/decouple-mem-state.cpp
@@ -307,7 +307,8 @@ follow_state_edge(
     {
       return rr->output();
     }
-    else if (rvsdg::TryGetOwnerNode<loop_node>(*state_edge))
+
+    if (rvsdg::TryGetOwnerNode<loop_node>(*state_edge))
     {
       std::vector<rvsdg::SimpleNode *> loop_mem_ops;
       auto si = jlm::util::AssertedCast<rvsdg::StructuralInput>(state_edge);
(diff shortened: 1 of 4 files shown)


caleridas:
There is this one already: [TryGetOwnerOp(const rvsdg::input & input) noexcept](https://github.com/phate/jlm/blob/9457369e14bf0d3864d44a888b4c7d56184d03e0/jlm/hls/backend/rvsdg2rhls/hls-function-util.hpp#L54)

I think this can be consolidated immediately. I am not overly fond of non-orthogonal utility functions, but this one seems to fulfill a need that is ubiquituous. I am not certain you need both node and op -- on the one hand, that's the most closed and sensible API. On the other hand, most places don't need it.

phate:
@caleridas I removed the `TryGetOwnerOp()` method from the code

haved:
@phate is this PR supposed to touch decouple-mem-state?

Regarding the API, I think it is a bit weird to get only the node, if the operation happens to be wrong, but it is still a SimpleNode. However, I do not really have a great alternative, and users of the function are not likely to actually use that behavior for anything

phate:
@haved Well, @caleridas asked me to remove the TryGetOwnerOp(), but I agree, it should maybe not be in this PR. I can split them up, we merge the TryGetSimpleNodeAndOp() part first, and then the removal of the TryGetOwnerOp() part.

@caleridas @haved Would this be okay for both of you?

haved:
@phate oh ok then I understand, no worries, then, you do not need to split it up

phate:
@caleridas @haved Could I get an approval if it looks all right, please?

haved (approved):

PR #916, opened 2025-05-05
Separate SimpleOperation / StructuralOperation

caleridas:
Separate the inheritance hierarchies of SimpleOperation / StructuralOperation. This makes "SimpleOperation" the root class for all operational nodes. The inheritance hierarchy for StructuralOperation is kept for the moment, eventually the "contents" of structural operations can be entirely disassociated. They should evolve into something that can carry information specific to the structure, particularly backend-specific customizations of structural nodes.


phate:
@caleridas I admire your determination, but maybe it is better to split this in multiple PRs. In particular, for the HLS transformations.

PR #930, opened 2025-06-02
Add memory state operation reductions

phate:
This PR does the following:
1. Implement several normalizations for the MemoryStateMerge operation
2. Implement several normalizations for the MemoryStateSplit operation
3. Enables these normalizations by including them in the node reduction transformation


phate:
The HLS problem seems to arise from the NormalizeSingleOperand() reduction.

phate:
@haved This PR can be reviewed now.

sjalander:
@phate @haved 
I see that it's ready. Thought I would update the cycle count as several HLas benchmarks are improved.

jlm/llvm/ir/operators/MemoryStateOperations.cpp
```diff
@@ -29,6 +30,86 @@ MemoryStateMergeOperation::copy() const
...
+
+  return std::nullopt;
+}
+
+std::optional<std::vector<rvsdg::Output *>>
+MemoryStateMergeOperation::NormalizeDuplicateOperands(
+    const MemoryStateMergeOperation &,
+    const std::vector<rvsdg::Output *> & operands)
+{
+  const util::HashSet<rvsdg::Output *> uniqueOperands(operands.begin(), operands.end());
```
haved, line 49: This operation will shuffle the operands when it is applied. Would it be worth it to use a vector in combination?

For the future maybe we could implement a insertion order maintaining hash set

haved (approved):
@phate looks good, I just have some comments that are probably not important at all

sjalander (approved):

jlm/llvm/ir/operators/MemoryStateOperations.cpp
phate: @haved Well, I need a set in order to determine in linear time whether we have duplicates. This would not work with a vector (with the same runtime). Is it a problem that this would shuffle the operands? There is no order defined on those operands anyway. Semantically, there is no problem either.

jlm/llvm/ir/operators/MemoryStateOperations.cpp
```diff
@@ -29,6 +30,86 @@ MemoryStateMergeOperation::copy() const
...
+
+  if (operands == newOperands)
+    return std::nullopt;
+
+  auto result = Create(std::move(newOperands));
+  return { { result } };
+}
+
+std::optional<std::vector<rvsdg::Output *>>
+MemoryStateMergeOperation::NormalizeNestedSplits(
```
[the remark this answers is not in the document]
phate: @haved Other suggestion for name: NormalizeMergeSplit!?

Regarding semantics: From an RVSDG perspective it is not illegal to have mutiple users for state outputs, as it is sequentialization mechanism. From a program perspective, it depends on the state type. In case of memory state, it is not a problem per se and we do not have any mechanism enforced to always have an explicit fork right now. What exactly is your concern here?

jlm/llvm/ir/operators/MemoryStateOperations.cpp
haved: @phate yes, I will just make a mental note about this being a place that introduces non-deterministic passes

haved:
@phate you answered my comments, just click resolve when you have read my short responses

haved (approved):

jlm/llvm/ir/operators/MemoryStateOperations.cpp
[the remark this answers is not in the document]
phate: @haved I renamed the function. I need annother approval.

sjalander (approved):

sjalander (approved):

PR #935, opened 2025-06-04
Rename entryaggnode class to EntryAggregationNode

phate:


jlm/llvm/ir/aggregation.cpp
```diff
@@ -55,25 +55,22 @@ AggregationNode::normalize(AggregationNode & node)
     normalize(child);
 }
 
-/* entryaggnode class */
+EntryAggregationNode::~EntryAggregationNode() noexcept = default;
```
sjalander, line 58: What does this do differently from what's already in the hope?

sjalander (approved):

jlm/llvm/ir/aggregation.cpp
phate: I do not understand the question?

sjalander: @phate Håvard explained the benefit of adding this to the CPP and not only have it in the HPP.

PR #976, opened 2025-06-15
Fix bug in InvariantValueRedirection pass

### LINKED ISSUE #937: RegionAware with InvariantValueRedirection fails
This is a reduced version of a file in the LLVM test suite.

**Recreate**
`jlc -JAAAndersenRegionAware -JInvariantValueRedirection fpcmp.c`

**Output**

> jlc: jlm/llvm/opt/InvariantValueRedirection.cpp:218: static void jlm::llvm::InvariantValueRedirection::RedirectCallOutputs(rvsdg::SimpleNode &): Assertion `lambdaEntrySplit && lambdaExitMerge && callEntryMerge' failed.

[fpcmp.c.tgz](https://github.com/user-attachments/files/20592327/fpcmp.c.tgz)

phate:
The pass falsely assumed that if the CallMemoryStateExitSplit node is present for a call, then all the other memory state nodes (i.e., CallMemoryStateEntryMerge,  LambdaMemoryStateEntrySplit, LambdaMemoryStateExitMerge) are present as well. This does not have to be the case in general. For example, it might be that after region-aware state encoding, the LambdaMemoryStateEntrySplit node is missing, while all other nodes are present.

Close #937

### DIFF
--- a/jlm/llvm/opt/InvariantValueRedirection.cpp
+++ b/jlm/llvm/opt/InvariantValueRedirection.cpp
@@ -3,6 +3,9 @@
  * See COPYING for terms of redistribution.
  */
 
+#include "jlm/llvm/backend/dot/DotWriter.hpp"
+#include "jlm/util/GraphWriter.hpp"
+#include <fstream>
 #include <jlm/llvm/ir/LambdaMemoryState.hpp>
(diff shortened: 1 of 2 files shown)


haved (approved):
Looks good, nice test

PR #990, opened 2025-06-18
Add the option of handling unification during TarjanSCC.

haved:
The Topological Worklist Policy and Online Cycle Detector has not been handling unification properly due to the TarjanSCC algorithm not handling unification. This PR adds the ability for Tarjan to handle graphs with node unification.

As a consequence, it became more practical to provide the topological order in reverse, which is why some other functions have been changed


tests/jlm/util/TestTarjanScc.cpp
```diff
@@ -8,71 +8,84 @@
...
 #include <cassert>
-#include <cstdint>
 #include <optional>
 #include <tuple>
-#include <unordered_map>
-#include <unordered_set>
 #include <vector>
 
 #include <iostream>
+#include <jlm/util/HashSet.hpp>
```
phate, line 16: The header should be moved up to the other jlm headers.

phate (approved):
Sorry that it took so long.

haved:
@phate I merged the latest master into this PR, so I need a new approval. Otherwise nothing has changed since the last time you looked at it

PR #992, opened 2025-06-21
Add LambdaExitMemoryStateMergeOperation normalizations

phate:
Adds three LambdaExitMemoryStateMergeOperation normalizations that together render load, store, and alloca operations independent from the lambda node, leading to them potentially becoming dead code. This should be able to remove a significant amount of load, stores, and allocas that are normally eliminated by mem2reg.

### DIFF
--- a/jlm/llvm/ir/operators/Load.hpp
+++ b/jlm/llvm/ir/operators/Load.hpp
@@ -149,6 +149,9 @@ public:
              MemoryStateOutputIterator(nullptr) };
   }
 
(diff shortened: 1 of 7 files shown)


haved (approved):
Great work, it will be very interesting to see how much it is able to handle.

It might make for a faster pass if it recursively kept traversing up through the memory state edge, but this implementation is cleaner.

It is not able to handle loads and stores inside gammas and thetas, which is another thing to think about

PR #1005, opened 2025-07-01
Rework HLS fork insertion pass

phate:


jlm/hls/backend/rvsdg2rhls/add-forks.cpp
```diff
@@ -5,62 +5,89 @@
...
       }
     }
   }
 }
 
 void
-add_forks(llvm::RvsdgModule & rvsdgModule)
+ForkInsertion::AddForkToOutput(rvsdg::Output & output)
+{
+  JLM_ASSERT(output.nusers() > 1 && output.nusers() != 0);
```
haved, line 63: This assert seems a bit "extra"

haved (dismissed):
Very nice refactor, thank you. I have one comment

jlm/hls/backend/rvsdg2rhls/add-forks.cpp
haved: I mean the second part of the &&, I found it confusing that it checks both > 1 and != 0

@phate

phate: @haved Ah yeah, sorry, it took me a while to realize that the second part was always true if the first part is true. My bad.

PR #1016, opened 2025-07-03
Rework HLS sink insertion pass

phate:


haved (approved):

phate:
@haved I resolved the problem. The issue was that the root region has two dead imports on which the reworked pass also puts hls sink nodes (see graph below). The follow up passes did not like this as they expect the root region to only have a single node.

Now I had two orthogonal options to resolve this:
1. Remove these dead imports by running DNE before. The problem is that I do not know whether they are actually needed (even though they are dead).
2. Avoiding to insert sink nodes in the root region as the original pass did.

I opted for option 2 as it was the more defensive and safer option. However, we should find out whether these two dead imports are actually needed or whether they can be removed.

image

@sjalander FYI

haved:
They should not be used any longer, as all calls to those functions have been replaced by memory queue operations. @phate

haved (approved):

PR #1017, opened 2025-07-03
Simplify HLS merge gamma pass

phate:
Contrary to what the comment in the code says, DNE is perfectly capable to remove the dead gammas created by the merge gamma pass.


phate:
This code path triggers, but it only triggers because the merge_gamma pass does some transformations on the gamma nodes, which ultimately leads to a dead gamma. It has nothing to do with DNE missing a dead gamma or something.

PR #1034, opened 2025-07-11
Rename fpbin_op class to FBinaryOperation

phate:


haved (dismissed):

jlm/llvm/ir/operators/operators.cpp
```diff
@@ -543,45 +543,46 @@ PoisonValueOperation::copy() const
...
+FBinaryOperation::debug_string() const
 {
-  static std::unordered_map<llvm::fpop, std::string> map({ { fpop::add, "add" },
-                                                           { fpop::sub, "sub" },
-                                                           { fpop::mul, "mul" },
-                                                           { fpop::div, "div" },
-                                                           { fpop::mod, "mod" } });
+  static std::unordered_map<llvm::fpop, std::string> map({ { fpop::add, "Add" },
+                                                           { fpop::sub, "Sub" },
+                                                           { fpop::mul, "Mul" },
```
sjalander: I'm not at a computer so I've not been able to check. 
But I'm guessing these name changes can be the reason for the HLS to fail. The floating-point operations do not get generated. They are instead pre-built macros and I guess this change affects the name so they are not found.

phate: Yeah, I figured the same. I changed them back and it seems to do better. This being implicitly bound to the debug string is not great, it should be an explicit mapping. Changing that is for another PR though.

sjalander (approved):

PR #1053, opened 2025-07-15
Rename domnode class to DominatorTreeNode

phate:


jlm/llvm/ir/domtree.cpp
```diff
@@ -13,10 +13,10 @@
 namespace jlm::llvm
 {
 
-/* domnode class */
+/* DominatorTreeNode class */
```
sjalander, line 16: Don't you want //
Or no comment at all

PR #1059, opened 2025-07-16
Remove outdated begin() and end() methods from Output class

phate:


haved (approved):
Very nice, the `SingleUser` is also a great addition

PR #1060, opened 2025-07-17
Add owner attribute to Input class

phate:


caleridas (approved):
Essentially devirtualizes access to the owner of the input. LGTM.

PR #1068, opened 2025-07-19
Refactor SimpleInput dispatching

phate:
We would like to remove the SimpleInput subclass, but before we can do this we need to refactor all the dispatching that is based on it.


jlm/hls/backend/rvsdg2rhls/decouple-mem-state.cpp
```diff
@@ -84,10 +84,11 @@ trace_edge(
...
     }
-    auto si = util::AssertedCast<rvsdg::SimpleInput>(state_edge);
-    auto sn = si->node();
-    auto new_si = util::AssertedCast<rvsdg::SimpleInput>(new_edge_user);
-    auto new_sn = new_si->node();
+
+    auto si = state_edge;
+    auto sn = rvsdg::TryGetOwnerNode<rvsdg::SimpleNode>(*si);
+    auto new_si = new_edge_user;
+    auto new_sn = rvsdg::TryGetOwnerNode<rvsdg::SimpleNode>(*new_si);
```
haved, line 91: This `sn` and `new_sn` should use `AssertGetOwnerNode`, no?

jlm/hls/backend/rvsdg2rhls/mem-conv.cpp
```diff
@@ -64,8 +64,8 @@ TraceEdgeToMerge(rvsdg::Input * state_edge)
     {
       JLM_UNREACHABLE("there should be no new loops");
     }
-    auto si = util::AssertedCast<rvsdg::SimpleInput>(state_edge);
-    auto sn = si->node();
+    auto si = state_edge;
+    auto sn = rvsdg::TryGetOwnerNode<rvsdg::SimpleNode>(*si);
```
haved, line 68: Here as well. `AssertGetOwnerNode`?

jlm/hls/backend/rvsdg2rhls/mem-queue.cpp
```diff
@@ -215,9 +214,8 @@ separate_load_edge(
             // But we are not allowed to discard the vector and can't have unused variables
             // So adding a meaningless assert to get it to compile
             JLM_ASSERT(dummy_user_tmp.size() == 0);
-            auto dummy_user =
-                dynamic_cast<jlm::rvsdg::SimpleInput *>(&*load_branch_out[i]->Users().begin())
-                    ->node();
+            auto dummy_user = jlm::rvsdg::TryGetOwnerNode<jlm::rvsdg::SimpleNode>(
```
haved, line 217: Here as well. `AssertGetOwnerNode`.

haved (approved):
Look good, just some `Try` I think should be `Assert`. @phate

jlm/hls/backend/rvsdg2rhls/decouple-mem-state.cpp
phate: Fixed

PR #1069, opened 2025-07-20
HLS: Overhaul RedundantBufferElimination pass

phate:


haved (dismissed):
Very nice. The docstring could maybe specify that it only applies to memory state edges?

phate:
@haved I fixed the docstring.

PR #1080, opened 2025-07-22
Remove outdated Create() method from SimpleNode class

phate:
The PR does the following:

1. Remove a Create() method from SimpleNode class. The method lead to a lot of unnecessary additional operation allocation and copies. Remove it such that it can not be misused any longer.
2. Extend the second Create() method of the SimpleNode class to also take a Operation instead of just a SimpleOperation. This leads to a more usable interface for the user and allows to combine the method easily with the copy() method of the Operation class.


haved (approved):
Very nice interface changes

PR #1086, opened 2025-07-23
Overhaul Input::iterator class

phate:
The GetOwner() method of the Input class allows us to compute the next input directly in the iterator, simplifying all the iterator code a lot.


haved (approved):
Nice find

PR #1096, opened 2025-07-26
Add normalization to NodeReduction pass

phate:
The LoadNonVolatileOperation::NormalizeIOBarrierAllocaAddress was not part of the NodeReduction pass.


phate:
Well that is interesting....I guess some debugging is necessary.

PR #1104, opened 2025-07-27
Handle RawConstantIndices for MLIR GEPOp

sjalander:


jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -702,9 +702,22 @@ MlirToJlmConverter::ConvertOperation(
 
     auto pointeeValueType = std::dynamic_pointer_cast<const rvsdg::ValueType>(pointeeType);
 
+    // Constant indices are not part of the inputs to a GEPOp,
+    // but they are required as explicit nodes in RVSDG
+    std::vector<rvsdg::Output *> inputsWithConstants(std::next(inputs.begin()), inputs.end());
```
phate, line 707: The vector could be better named....maybe "indices"?

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -702,9 +702,22 @@ MlirToJlmConverter::ConvertOperation(
...
+    // TODO: There should exist a better way to check if there are constants
+    if (GepOp.getNumOperands() == 1 && GepOp.getRawConstantIndices().size() != 0)
+    {
+      for (size_t constant : GepOp.getRawConstantIndices())
+      {
+        inputsWithConstants.push_back(
+            jlm::llvm::IntegerConstantOperation::Create(rvsdgRegion, 32, constant).output(0));
+      }
+    }
+
```
phate, line 717: I do not like it that it is implicit that in case the raw constants are not there, then the inputsWithConstants vector is empty. Could we rather make this explicit with an if-then-else statement:

```
std::vecotr<rvsdg::Output*> indices;
if (raw constants presents)
{
  indices = // add constants here
}
else
{
  indices = {std::next(inputs.begin()), inputs.end()}
}
```

phate (approved):

phate (approved):

PR #1105, opened 2025-07-27
Add LocalAliasAnalysis and test

haved:
This PR adds the `LocalAliasAnalysis` class, which contains quite a bit of tricks that try to establish guarantees about aliasing. Hopefully the comments go a long way to explain what is going on at each step.


jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
```diff
@@ -0,0 +1,825 @@
...
+  }
+  else if (const auto outerTheta = rvsdg::TryGetRegionParentNode<rvsdg::ThetaNode>(output))
+  {
+    const auto loopVar = outerTheta->GetLoopVars()[output.index()];
+
+    // If the loop variable is invariant, continue normalizing
+    if (ThetaLoopVarIsInvariant(loopVar))
+      return NormalizeOutput(*loopVar.input->origin());
+  }
+  else if (const auto outerLambda = rvsdg::TryGetRegionParentNode<rvsdg::LambdaNode>(output))
```
phate, line 775: Why outerLambda?

jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
```diff
@@ -0,0 +1,825 @@
...
+  {
+    const auto ctxVar = outerLambda->MapBinderContextVar(output);
+
+    // If the argument is a contex variable, continue normalizing
+    if (ctxVar)
+      return NormalizeOutput(*ctxVar->input->origin());
```
phate, line 781: `if (const auto ctxVar = ...)`

jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
```diff
@@ -0,0 +1,825 @@
...
+  }
+  else if (const auto phiNode = rvsdg::TryGetRegionParentNode<rvsdg::PhiNode>(output))
+  {
+    const auto argument = phiNode->MapArgument(output);
+    if (const auto cvArg = std::get_if<rvsdg::PhiNode::ContextVar>(&argument))
```
phate, line 786: Let's call it `ctxVar` and not `cvArg` to be consistent

jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
```diff
@@ -0,0 +1,825 @@
...
+    {
+      // Follow the context variable to outside the phi
+      return NormalizeOutput(*cvArg->input->origin());
+    }
+    if (const auto fixArg = std::get_if<rvsdg::PhiNode::FixVar>(&argument))
```
phate, line 791: `fixVar`

jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
```diff
@@ -0,0 +1,825 @@
...
+{
+  const auto & p1Norm = NormalizePointerValue(p1);
+  const auto & p2Norm = NormalizePointerValue(p2);
+
+  // If the two pointers are the same value, they must alias
+  if (&p1Norm == &p2Norm)
+    return MustAlias;
+
+  // Trace through GEP operations to get closer to the origins of the pointers
+  // Only traces through GEPs where the offset is known at compile time,
```
phate, line 84: Only trace... (remove s)

jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
```diff
@@ -0,0 +1,825 @@
...
+  int64_t offset = 0;
+
+  while (true)
+  {
+    // Use normalization function to get past all trivially invariant operations
+    base = &NormalizePointerValue(*base);
+
+    if (auto node = rvsdg::TryGetOwnerNode<rvsdg::SimpleNode>(*base))
+    {
+      if (is<GetElementPtrOperation>(node->GetOperation()))
```
phate, line 282: Use `TrySimpleNodeAndOp()`

jlm/llvm/opt/alias-analyses/AliasAnalysis.hpp
```diff
@@ -0,0 +1,343 @@
...
+   * Traces through GEP operations, including those with offsets that are not known at compile time.
+   * Also traces through gamma and theta nodes, building a set of multiple possibilities.
+   * Tracing stops at "top origins", for example an ALLOCA, a LOAD, or the return value of a CALL.
+   *
+   * @param p the pointer to trace from
+   * @param traceCollection the collection of trace points being created
+   * @return false if the trace collection reached its maximum allowed size, and tracing aborted
+   */
+  [[nodiscard]] static bool
+  TraceAllPointerOrigins(TracedPointerOrigin p, TraceCollection & traceCollection);
```
phate, line 143: Could this interface be simplified a bit to: `std::optional<TraceCollection> TraceAllPointerOrigins(TracedPointerOrigin p)`. In other words, we only get a TraceCollection if we did not reac the maximum allowed size?

phate: Nah, the function is recursive....not possible. Ignore this comment.

jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
```diff
@@ -0,0 +1,825 @@
...
+  return false;
+}
+
+bool
+LocalAliasAnalysis::IsOriginalOriginFullyTraceable(const rvsdg::Output & pointer)
+{
+  // The only original origins that can be fully traced for escaping are ALLOCAs
+  const auto originalNode = rvsdg::TryGetOwnerNode<rvsdg::SimpleNode>(pointer);
+  if (!is<AllocaOperation>(originalNode))
+    return false;
```
phate, line 608: `TrySimpleNodeAndOp`

jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
```diff
@@ -0,0 +1,825 @@
...
+      if (elementCount.has_value())
+        return *elementCount * GetTypeSize(*allocaOp->ValueType());
+    }
+    if (is<MallocOperation>(node))
+    {
+      const auto mallocSize = GetConstantIntegerValue(*node->input(0)->origin());
+      if (mallocSize.has_value())
+        return *mallocSize;
+    }
+  }
```
phate, line 484: `TrySimpleNodeAndOp`

phate (commented):
Here is the first set of comments. I only found minor stuff. The code is very readable and understandable, it still takes quite a while to get through it and tires one out. It is non-trivial stuff. I will need another round on it at some point.

Great work!

haved:
I tended to your comments @phate. I am working on an additional test.

In the meantime:

In a subsequent PR I am thinking of adding something like:
 - `IsOwnerNodeOp<AllocaOperation>(input / output)`
 - or `IsOwnerOp<AllocaOperation>(input / output)`
 - or `IsSimpleNodeAndOp<AllocaOperation>(input / output)`

which is a simple `bool`-returning helper to avoid having to use `TryGetSimpleNodeAndOp` only to discard the node and check if the op is `nullptr` or not.

I am not sure which name I like best, however.
Personally I might prefer the top option.

"Owner" alone is a bit imprecise, as it can also refer to the region of an argument or result.
I only really want this function to handle actual node inputs and outputs.

One way to sidestep the issue of arguments/results is to require that we only work for `SimpleNode`s, which never have subregions. Using a name like `IsSimpleNodeAndOp` nicely mirrors `TryGetSimpleNodeAndOp`, but in this case the word "Is" is being misused.

We can of course discuss the naming once I actually make the PR, but if any of you have some early opinions on naming I would appreciate them. @phate @caleridas @sjalander

phate (changes requested):
Considering that the PrecisionEvaluator is not part of the newly added files in the Makefile, I suspect it ended up in here by accident.

I did otherwise not find much anymore. LGTM

jlm/llvm/opt/alias-analyses/AliasAnalysis.hpp
```diff
@@ -0,0 +1,105 @@
...
+ *   }[x]
+ *   ...
+ *
+ * Normalizing p3 yields p1
+ *
+ * @param output the output to trace from
+ * @return the most normalized source of the given output
+ */
+[[nodiscard]] const rvsdg::Output &
+NormalizeOutput(const rvsdg::Output & output);
```
caleridas, line 90: there is a similar notion in the "producer" function here: https://github.com/phate/jlm/blob/master/jlm/rvsdg/node.cpp#L405

I am not too happy about either name (neither "producer" nor "NormalizeOutput" catches it), but I think the concept is generally / universally applicable.

jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    JLM_UNREACHABLE("Unknown phi argument type");
+  }
+
+  return output;
+}
+
+const rvsdg::Output &
+NormalizePointerValue(const rvsdg::Output & pointer)
+{
```
caleridas, line 109: Is it really warranted to have an API function that does nothing on top of the more generic one besides adding an assertion?

It would be more orthogonal from an API point of view to have a function that is an "identity mapping" (i.e. returns the same output), but adds an assertion failure that can then be _combined_ with the generic origin tracer function.

jlm/llvm/opt/alias-analyses/LocalAliasAnalysis.hpp
```diff
@@ -0,0 +1,266 @@
...
+/**
+ * Gets the value of the given \p output as a compile time constant, if possible.
+ * The constant is interpreted as a signed value, and sign extended to int64 if needed.
+ * This function does not perform any constant folding.
+ *
+ * @param output the output whose constant value is requested
+ * @return the value of the output, or nullopt if it could not be determined.
+ */
+[[nodiscard]] std::optional<int64_t>
+GetConstantSignedIntegerValue(const rvsdg::Output & output);
```
caleridas, line 262: Does it need to be exposed? If it is private to LocalAliasAnalysis, then maybe not expose in header. Other than that, the concept sounds "fairly generic" and not related to alias analysis at all to me.

tests/TestRvsdgs.hpp
```diff
@@ -2537,4 +2537,136 @@ class VariadicFunctionTest2 final : public RvsdgTest
...
+ *     *p = &alloca2;
+ *
+ *     char* bytePtr = getPtr();
+ *     char* bytePtrPlus2 = bytePtr + 2;
+ *
+ *     // All alias queries happen here
+ *   }
+ * \endcode
+ */
+class LocalAliasAnalysisTest1 final : public RvsdgTest
```
caleridas, line 2574: This is not actionable right now, but ... TestRvsdgs has become too much of a place to collect all kinds of graphs. It is a justified place if the graphs here are subject to "multiple kinds of tests", but otherwise it is becoming quite unwieldy.

caleridas (approved):
I don't have anything to say about the analysis logic itself, it looks all good. Small questions only on "where to put things", nice if you want to address it, but not blocking for this PR I think.

tests/TestRvsdgs.hpp
haved: Yes I agree, these graphs are only really used for one test each, so it does not really make sense to place them there

PR #1115, opened 2025-07-29
Remove outdated TestGraphResult class

phate:


jlm/rvsdg/region.cpp
```diff
@@ -115,7 +115,6 @@ RegionResult::Create(
     StructuralOutput * output,
     std::shared_ptr<const rvsdg::Type> type)
 {
-  JLM_ASSERT(origin.region() == &region);
```
phate, line 118: This JLM_ASSERT() should never have been there in the first place. The regions of are checked further up in the Input class constructor.

haved (approved):

PR #1116, opened 2025-07-30
Add NormalizeCallEntryMemoryStateMerge reduction

phate:
Adds the NormalizeCallEntryMemoryStateMerge reduction to the LambdaEntryMemoryStateSplitOperation class and enables it in the NodeReduction pass

### DIFF
--- a/jlm/llvm/ir/aggregation.hpp
+++ b/jlm/llvm/ir/aggregation.hpp
@@ -17,105 +17,11 @@ class ControlFlowGraph;
 
 class AggregationNode
 {
-  class iterator final
-  {
-  public:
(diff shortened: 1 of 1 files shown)


haved (approved):
Looks good. I think the semantics of these nodes are a bit underspecified, which might make it a bit tricky to modify them, but it makes sense to try and go ahead and then find out as we go along

phate:
Well, interesting....

phate:
I looked at the problem, and in my opinion, this JLM_ASSERT() is just plain wrong. I think I understood the problem and will set up a unit test, fixing the issue. The added transformation was just the enabler for this to fail.

PR #1118, opened 2025-07-31
Move DotWriter.[cpp/hpp] out of backend folder

phate:


phate:
I am tired of manually inserting code to print out dot files between transformations. I want some functionality of printing out dot graphs after/before transformations similar to LLVM's --print-after-all etc. This is the first step in this direction.

haved (approved):

PR #1122, opened 2025-08-01
Fix Common Node Elimination

phate:
This PR does the following:
1. It could be that we check congruence for values that are the arguments from different theta nodes. This caused an JLM_ASSERT() to be thrown in the original code. Let's ensure that we handle this case gracefully.
2. Overhaul the CNE congruence check to reflect the latest API changes.


phate:
This hopefully fixes the issue from #1116

haved (approved):

PR #1129, opened 2025-08-02
Remove outdated jlm::tests::GraphImport class

phate:


haved:
Very weird CI failure, did you hit some sort of rate limit or did GitHub just have a hiccup? @phate

jlm/rvsdg/graph.cpp
```diff
@@ -24,6 +24,20 @@ GraphImport::debug_string() const
...
+GraphImport::Copy(Region & region, StructuralInput *)
+{
+  return Create(*region.graph(), Type(), Name());
+}
+
+GraphImport &
+GraphImport::Create(Graph & graph, std::shared_ptr<const rvsdg::Type> type, std::string name)
+{
+  auto graphImport = new GraphImport(graph, std::move(type), std::move(name));
+  graph.GetRootRegion().append_argument(graphImport);
```
haved, line 37: Is it necessary to append the argument even when the `GraphImport` constructor passes the root region to the `RegionArgument` constructor? If so, why is it not also done in the `Copy` method?

haved (approved):
Looks good, just one question

phate:
@haved I suspect hiccup.

haved (approved):

PR #1131, opened 2025-08-04
Add unique node identifier

phate:


phate:
@caleridas @haved While this is not much, but I would like to have both your review on this as this is rather important to get right.

haved:
@phate @caleridas is it a bit overkill to make the Id globally unique? It kind of removes the ability to do anything multithreaded ever. If the only thing we care about is the relative order within regions, we can get away with making it atomically increasing, but I think it is a bit more hygienic to avoid global variables altogether and have one counter per region. This would still be deterministic, assuming passes do not place nodes from different regions into the same priority queue of `std::pair<id, node*>`s. This should in practice be easy to avoid. As long as traversers are deterministic, most passes automatically become deterministic, as long as they avoid placing `node*/output*` in sets and then iterating over them.

jlm/rvsdg/node.cpp
```diff
@@ -252,8 +252,16 @@ node_output::node_output(Node * node, std::shared_ptr<const rvsdg::Type> type)
       node_(node)
 {}
 
+static Node::Id
+CreateNodeId() noexcept
+{
+  static Node::Id id = 0;
```
caleridas, line 258: Well this immediately causes thread-safety issues for the entire library (e.g. two threads using the rvsdg library concurrently, on _entirely_ distinct compilation units), so this is not advisable. I would suggest not making a mutable library-wide global. Per region should be fine.

caleridas (changes requested):
Library-wide global is definitely problematic. It could be made "atomic" to at least be correct, but I would rather not create "library-wide mutable global state" to begin with.

phate:
@caleridas Are you fine with it as well?

PR #1133, opened 2025-08-05
Move delta node definition to rvsdg

caleridas:
Make delta node a generic rvsdg node. Let llvm backend extend / override the delta operation bits.


phate (changes requested):
Nothing major, I would still like to rename the files.

phate (approved):

PR #1136, opened 2025-08-06
Remove global tracker registry

phate:
Replaces the global tracker registry with a per-region registry as trackers work nowadays solely on a per-region basis and not on a per-graph basis any longer.


phate:
@caleridas @haved Anyone, please!?

haved (dismissed):
Looks good, I just had to dig around the code for a bit to realize that Trackers are a separate concept to the callbacks we register on the Notifiers in `rvsdg/notifiers.cpp`. The latter are still global, which explains why this PR was smaller than I first thought it would be.

PR #1148, opened 2025-08-13
Add check to RegionAwareModRefSummary

phate:


haved (dismissed):
The check makes sense, at least for as long as we do not have any context sensitivity

phate:
@caleridas Here is the renaming.

PR #1150, opened 2025-08-13
LambdaBuilder and one API clarification

caleridas:
Add a lambda builder object. This allows incrementally building a lambda construct. It offers an API that allows to determine the full type signature (including return types) of the function at the very end.

Remove one superfluous std::optional specification in a delta API.

### DIFF
--- a/jlm/rvsdg/lambda.hpp
+++ b/jlm/rvsdg/lambda.hpp
@@ -19,6 +19,8 @@
 namespace jlm::rvsdg
 {
 
+class LambdaBuilder;
+
 /** \brief Lambda operation
  *
  * A lambda operation determines a lambda's \ref FunctionType "function type".
@@ -115,6 +117,17 @@ public:
     rvsdg::Output * inner;
   };
 
+  /**
+   * \brief Formal argument variable
+   */
+  struct ArgumentVar
+  {
+    /**
+     * \brief Access argument object in subregion.
+     */
+    rvsdg::Output * arg;
+  };
+
   [[nodiscard]] std::vector<rvsdg::Output *>
   GetFunctionArguments() const;
 
@@ -197,6 +210,18 @@ public:
   GetContextVars() const noexcept;
 
   /**
+   * \brief Maps region argument to its disposition (formal argument or context var).
+   *
+   * \returns
+   *   A description of the role that this arguments plays.
+   *
+   * \pre
+   *   \p output must be an argument to the subregion of this node.
+   */
+  std::variant<ArgumentVar, ContextVar>
+  MapArgument(const rvsdg::Output & output) const;
+
+  /**
    * Remove lambda inputs and their respective arguments.
    *
    * An input must match the condition specified by \p match and its argument must be dead.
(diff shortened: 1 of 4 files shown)


jlm/rvsdg/lambda.hpp
```diff
@@ -287,6 +314,81 @@ LambdaNode::RemoveLambdaInputsWhere(const F & match)
...
+   *
+   * \return
+   *   Region arguments that represent formal parameters of the function.
+   *
+   * \pre
+   *   \p this must not have been finalized yet.
+   *
+   */
+  std::vector<Output *>
+  Arguments();
```
phate, line 340:
```
[[nodiscard]] std::vector<Output*>
Arguments() const noexcept
```

jlm/rvsdg/lambda.hpp
```diff
@@ -287,6 +314,81 @@ LambdaNode::RemoveLambdaInputsWhere(const F & match)
...
+   * \return
+   *   The region
+   *
+   * \pre
+   *   \p this must not have been finalized yet.
+   *
+   */
+  rvsdg::Region *
+  GetRegion();
+
```
phate, line 354:
```
[[nodiscard]] rvsdg::Region &
GetRegion() const noexcept
```

jlm/rvsdg/lambda.hpp
```diff
@@ -287,6 +314,81 @@ LambdaNode::RemoveLambdaInputsWhere(const F & match)
...
+   * \param op
+   *   The operations struct for this lambda, including the formal type signature of the function.
+   *
+   * \pre
+   *   \p this must not have been finalized yet.
+   *
+   * \return
+   *   The output representing the bound lambda object.
+   */
+  Output *
```
phate, line 385: Return type should be `rvsdg::Output &`

jlm/rvsdg/lambda.hpp
```diff
@@ -287,6 +314,81 @@ LambdaNode::RemoveLambdaInputsWhere(const F & match)
...
+  Finalize(const std::vector<jlm::rvsdg::Output *> & results, std::unique_ptr<LambdaOperation> op);
+
+private:
+  LambdaNode * node_;
```
phate, line 389: `Node_`

jlm/rvsdg/lambda.hpp
```diff
@@ -287,6 +314,81 @@ LambdaNode::RemoveLambdaInputsWhere(const F & match)
   return numRemovedInputs;
 }
 
+class LambdaBuilder
```
phate, line 317: Maybe a little doxygen comment on top why this builder is here and what is the difference of it compared to the LambdaNode::Create() and LambdaNode::Finalize() methods.

phate (changes requested):

jlm/rvsdg/lambda.hpp
caleridas: This should eventually replace all other means of constructing lambdas -- this PR is for illustrating and discussing whether you are fine with the shape before I start converting anything.

jlm/rvsdg/lambda.hpp
caleridas: done

jlm/rvsdg/lambda.hpp
caleridas: done

PR #1153, opened 2025-08-14
Add RouteToRegion() function

phate:
Adds a general RouteToRegion() function and replaces all usages of the old instances with the new function.


caleridas:
Interestingly I have an even more general (stateful with memoization and caching) concept of this in the cow to rvsdg compiler (to resolve bindings which cross layers implicitly in cow but need explicit routing in rvsdg). There are several nuances to this as linearity of states must also be preserved, states cannot be bound lambda contexts etc.

phate:
@caleridas Do your comment apply that you would like to see your version here and this should not be merged? Or was this just a FYI for now?

caleridas:
> @caleridas Do your comment apply that you would like to see your version here and this should not be merged? Or was this just a FYI for now?

I will get what I have out ASAP so you can take a look, I am not certain how to proceed

phate:
@caleridas I will need something like this function soon. Either we merge this one and later replace it with your "better" version, or you maybe could prepare what you got so we can take a look at it.

phate:
@haved Could you take a look at this PR please?

PR #1157, opened 2025-08-17
Add memory node identifiers to LambdaEntryMemoryStateSplitOperation class

phate:
…lass


phate:
@haved Every way I tried to properly introduce the memory node identifiers and also change the usage sides at once blew up in my face. Thus, I decided to introduce them first, and then adjust the usage sides individually, just so we have some progress. The usages that are mostly in my way are on the HLS side, but they are irrelevant for now.

haved (approved):
Looks good, reasonable to add this piecewise

PR #1162, opened 2025-08-20
Fix CallExitMemoryStateSplitOperation normalizations

phate:


jlm/llvm/ir/operators/MemoryStateOperations.cpp
```diff
@@ -279,6 +282,30 @@ LambdaExitMemoryStateMergeOperation::copy() const
   return std::make_unique<LambdaExitMemoryStateMergeOperation>(*this);
 }
 
+rvsdg::Input *
+LambdaExitMemoryStateMergeOperation::MapMemoryNodeIdToInput(
+    const rvsdg::Node & node,
+    const MemoryNodeId memoryNodeId)
+{
+  const auto operation =
+      dynamic_cast<const LambdaExitMemoryStateMergeOperation *>(&node.GetOperation());
```
haved, line 291: I think maybe Helge was moving away from nodes having operations (only `SimpleNode` having them), but we can handle that when we get there

jlm/llvm/ir/operators/MemoryStateOperations.cpp
```diff
@@ -279,6 +282,30 @@ LambdaExitMemoryStateMergeOperation::copy() const
...
+  if (!operation)
+  {
+    return nullptr;
+  }
+
+  try
+  {
+    const auto index = operation->MemoryNodeIdToIndex_.LookupKey(memoryNodeId);
```
haved, line 299: I guess doing separate calls to `HasKey` and `LookupKey` would be slower. I am not really used to using `try`/`catch`, but I guess it makes sense here

haved (approved):
@phate Looks good, I only have two comments that are not really actionable, just remarks

jlm/llvm/ir/operators/MemoryStateOperations.cpp
phate: I can change it. I just wanted to avoid the duplicated access to the map.

phate: Changed.

PR #1163, opened 2025-08-20
Support for handling MLIR index type and operations in frontend

sjalander:


jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -333,11 +333,20 @@ MlirToJlmConverter::ConvertBitBinaryNode(
...
-  if (!type.isa<::mlir::IntegerType>())
+  size_t width = 0;
+  if (type.isa<::mlir::IntegerType>())
+  {
+    auto integerType = type.cast<::mlir::IntegerType>();
+    width = integerType.getWidth();
+  }
+  else if (type.isIndex())
+  {
+    width = 32;
```
haved, line 344: Why 32? Ideally we would get the size of `uintptr` from somewhere central, but I would at least expect it to be 64, if we are hard coding it for now

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -508,6 +517,58 @@ MlirToJlmConverter::ConvertOperation(
         &rvsdg::CreateOpNode<llvm::ConstantFP>(rvsdgRegion, size, constant.value()));
   }
 
+  // RVSDG does not have an index type. Indices are therefore converted to integers.
+  // The width of indices are machine dependent and 32-bits are used for x86.
+
+  else if (auto constant = ::mlir::dyn_cast<::mlir::arith::ConstantIndexOp>(&mlirOperation))
```
haved, line 523: A lot of constants `32` here. I would at least put them in a constant variable to make it easy to change / grep after

haved (commented):
I would like to see all the 32's referring to a single global variable, and also I think 64 is more correct on most systems, so maybe use that instead. In the future we of course want a central way of defining info about the target architecture.

jlm/mlir/frontend/MlirToJlmConverter.cpp
sjalander: The reason is that the constant index for the GepOp seems to be hard coded to 32, so I stuck with the same size for these.

jlm/mlir/frontend/MlirToJlmConverter.cpp
sjalander: I created a static GetIndexBitWidth(), which then later can be modified to provide an architecture dependent width.

PR #1164, opened 2025-08-20
Add PointsToGraphAliasAnalysis and tests

haved:
It also includes the `ChainedAliasAnalysis` because it is a very small class.

I also had to add some `#includes` to make things compile on my machine after a system update.

`TryGetConstantSignedInteger` is moved to `AliasAnalysis.{cpp,hpp}` because it is needed by both analyses, but ideally it should live somewhere even more general. It uses `NormalizeOutputs` however, which should also be moved somewhere more general.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
+++ b/jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
@@ -47,7 +47,7 @@ ChainedAliasAnalysis::ChainedAliasAnalysis(AliasAnalysis & first, AliasAnalysis 
       Second_(second)
 {}
 
-ChainedAliasAnalysis::~ChainedAliasAnalysis() = default;
+ChainedAliasAnalysis::~ChainedAliasAnalysis() noexcept = default;
 
 AliasAnalysis::AliasQueryResponse
 ChainedAliasAnalysis::Query(
--- a/jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
+++ b/jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
@@ -28,7 +28,7 @@ AliasAnalysis::~AliasAnalysis() noexcept = default;
  * @param b second alias query response to check
  * @return true if the responses are compatible, false otherwise
  */
-bool
+static bool
 AreAliasResponsesCompatible(
     AliasAnalysis::AliasQueryResponse a,
     AliasAnalysis::AliasQueryResponse b)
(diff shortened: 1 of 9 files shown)


jlm/llvm/opt/alias-analyses/PointsToGraphAliasAnalysis.cpp
```diff
@@ -0,0 +1,147 @@
...
+        continue;
+
+      if (singleTarget.has_value())
+        return std::nullopt;
+
+      singleTarget = &target;
+    }
+
+    return singleTarget;
+  };
```
phate, line 61: You can make this lambda a private function to improve readability in this method.

tests/jlm/llvm/opt/alias-analyses/PointsToGraphAliasAnalysisTests.cpp
```diff
@@ -0,0 +1,482 @@
+/*
+ * Copyright 2025 Håvard Krogstie <krogstie.havard@gmail.com>
+ * See COPYING for terms of redistribution.
+ */
+
+#include "jlm/llvm/opt/alias-analyses/Andersen.hpp"
```
phate, line 6: Use `<>`

phate (changes requested):

jlm/llvm/opt/alias-analyses/AliasAnalysis.hpp
```diff
@@ -62,6 +64,28 @@ class AliasAnalysis
 };
 
 /**
+ * Class using two instances of AliasAnalysis to answer alias analysis queries.
+ * If the first analysis responds "May Alias", the second analysis is queried.
+ */
+class ChainedAliasAnalysis final : public AliasAnalysis
+{
+public:
+  explicit ChainedAliasAnalysis(AliasAnalysis & first, AliasAnalysis & second);
```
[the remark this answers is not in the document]
phate: All right, thanks for the explanation.

jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
```diff
@@ -17,6 +19,58 @@ AliasAnalysis::AliasAnalysis() = default;
...
+/**
+ * Checks if two alias query responses are compatible with each other.
+ * NoAlias is incompatible with MustAlias, and vice versa.
+ * MayAlias is compatible with all responses.
+ *
+ * @param a first alias query response to check
+ * @param b second alias query response to check
+ * @return true if the responses are compatible, false otherwise
+ */
+bool
```
phate, line 31: static

phate (approved):
One minor nitpick

jlm/llvm/opt/alias-analyses/AliasAnalysis.cpp
haved: Done

phate (approved):

PR #1168, opened 2025-08-22
Handle memory node identifiers in MemoryStateEncoder

phate:


haved (approved):
Looks good! I notice that some of the functions rely on the iteration order of a `HashMap` being the same every time, but I think that is a safe assumption when nothing has happened to the map in between. @phate

PR #1170, opened 2025-08-22
Replace CFG iterators with PtrIterator

phate:


jlm/llvm/ir/cfg.hpp
```diff
@@ -156,111 +156,10 @@ class ExitNode final : public ControlFlowGraphNode
...
-      return it_->get();
-    }
-
-  private:
-    std::unordered_set<std::unique_ptr<BasicBlock>>::const_iterator it_;
-  };
+  using iterator =
+      util::PtrIterator<BasicBlock, std::unordered_set<std::unique_ptr<BasicBlock>>::iterator>;
+  using const_iterator = util::
+      PtrIterator<BasicBlock, std::unordered_set<std::unique_ptr<BasicBlock>>::const_iterator>;
```
haved, line 162: This is not how you make a const iterator.

When we use `std::unordered_set<std::unique_ptr<BasicBlock>>::const_iterator`, that constness is about not being able to change what the `unique_ptr` is pointing to.
The iterator never exposes the `unique_ptr` anyways, so it is impossible to re-assign it.

To make a const iterator over `const BasicBlock &`, specify `const BasicBlock` as the type.

haved (dismissed):
Great to be able to remove code. One comment about const iterators. @phate

jlm/llvm/ir/cfg.hpp
phate: Added a const

PR #1173, opened 2025-08-22
Rename worklist class to Worklist

phate:


haved:
For a second I was wondering why I used `worklist` for my class, but I now see it is a different class entirely.

We could use `FifoWorklist<Node*>` here from `/jlm/util/Worklist.hpp`, and it would behave exactly the same.

PR #1174, opened 2025-08-22
Replace AggregationNode iterators with PtrIterator

phate:


jlm/llvm/ir/aggregation.hpp
```diff
@@ -17,105 +17,10 @@ class ControlFlowGraph;
...
-    {
-      return it_->get();
-    }
-
-  private:
-    std::vector<std::unique_ptr<AggregationNode>>::const_iterator it_;
-  };
+  using iterator =
+      util::PtrIterator<AggregationNode, std::vector<std::unique_ptr<AggregationNode>>::iterator>;
+  using const_iterator = util::
```
haved, line 22: Same here, `const AggregationNode` to get const iterator

phate: Fixed.

PR #1190, opened 2025-08-28
Add conversions of FMulAddIntrinsicOp to/from MLIR

haved:
We have previously linked together tests and tools (`jlm-opt`, `jlc`) with `libllvm.a` in front of `libmlir.a`.

This has for some reason worked perfectly fine before, and I have no idea why it suddenly stopped working, as the conversions I added are pretty much identical to existing conversions, and your class looks just like other operations.

However, I got a lot of "undefined reference to typeinfo for FMulAddIntrinsicOperation", "undefined reference to vtable for FMulAddIntrinsicOperation"

To appease the linker, I had to place the dependency (`libllvm.a`) after the dependant (`libmlir.a`). This is something I have previously had to do with `-l` libraries, but apparently it can also be necessary for regular input files... I wish I understood better. Do you know anything about this @caleridas?

### DIFF
--- a/jlm/mlir/frontend/MlirToJlmConverter.cpp
+++ b/jlm/mlir/frontend/MlirToJlmConverter.cpp
@@ -4,29 +4,22 @@
  * See COPYING for terms of redistribution.
  */
 
-#include <jlm/mlir/frontend/MlirToJlmConverter.hpp>
-#include <jlm/mlir/MLIRConverterCommon.hpp>
-
-#include <llvm/Support/raw_os_ostream.h>
-#include <mlir/Parser/Parser.h>
-#include <mlir/Transforms/TopologicalSortUtils.h>
-
-#include <jlm/llvm/ir/operators/sext.hpp>
-#include <jlm/rvsdg/bitstring/arithmetic.hpp>
-#include <jlm/rvsdg/bitstring/comparison.hpp>
-#include <jlm/rvsdg/bitstring/constant.hpp>
-
-#include <jlm/llvm/ir/operators/operators.hpp>
-
 #include <jlm/llvm/ir/operators/alloca.hpp>
 #include <jlm/llvm/ir/operators/call.hpp>
 #include <jlm/llvm/ir/operators/GetElementPtr.hpp>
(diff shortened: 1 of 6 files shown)


haved:
Now what in the world is happening during linking

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -690,10 +690,6 @@ MlirToJlmConverter::ConvertOperation(
         std::vector(inputs.begin(), inputs.end()),
         ConvertType(type)));
   }
-  else if (auto MallocOp = ::mlir::dyn_cast<::mlir::jlm::Malloc>(&mlirOperation))
-  {
-    return jlm::llvm::MallocOperation::create(inputs[0]);
-  }
```
phate, line 696: Why was this removed?

haved: Because there is another one five lines above it, that is identical. This if is thus dead.

I wasn't originally intending to include it in this PR, I staged it by accident.

tests/jlm/mlir/TestJlmToMlirToJlm.cpp
```diff
@@ -640,6 +640,69 @@ TestFpBinary()
...
 
+static void
+TestFMulAddOp()
+{
+  using namespace jlm::llvm;
+  using namespace mlir::rvsdg;
+
+  auto rvsdgModule = RvsdgModule::Create(jlm::util::FilePath(""), "", "");
+  auto graph = &rvsdgModule->Rvsdg();
+  {
```
phate, line 651: This compound statement is kind of useless, no?

phate (approved):

tests/jlm/mlir/TestJlmToMlirToJlm.cpp
haved: Yes it does not do anything, it was just copied from the other test I based mine on. I can remove it, or we can leave it for a later PR that cleans up all the tests if desired.

What do you think? @phate

phate: Your choice. I am fine either way.

PR #1191, opened 2025-08-29
Add TraceOutputIntraProcedurally function

phate:


haved (approved):
Very nice. We can also create an LLVM specific one that uses this to handle gammas and thetas

PR #1198, opened 2025-08-29
Separate Lllvm-specific from generic parts in dot writer

phate:
This PR tries to separate the LLVM specific parts in the dot writer from the generic RVSDG parts. This means that the generic DotWriter can then be moved into the rsvdg part of the compiler and be utilized there to create dot files.

### DIFF
--- a/jlm/llvm/DotWriter.cpp
+++ b/jlm/llvm/DotWriter.cpp
@@ -17,25 +17,15 @@
 
 namespace jlm::llvm::dot
 {
-/**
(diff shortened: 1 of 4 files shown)


phate:
@haved I understand that the result is rather suboptimal, but I tried to get away with as minimal changes as possible in order to not make the PR confusing and unblock us. Some more refactoring is needed at a later point in time to make this all a bit nicer.

haved (approved):
Very clean separation! I am almost surprised it got to be that clean

PR #1208, opened 2025-09-01
Add name to Transformations

phate:


phate:
@haved I need a name to separate the output files before/after each pass printing.

haved (approved):
Looks good. I pushed a commit to fix the clang-tidy-complaint, and also removed the default value of `Name_`, as it is set in the constructor anyways, and looks confusing to also have an unused default. @phate

PR #1209, opened 2025-09-02
Add support for dumping RVSDGs to jlm-opt

phate:
Adds jlm-opt support for dumping RVSDGs as dot graphs after each performed transformation.

### DIFF
--- a/jlm/rvsdg/Transformation.cpp
+++ b/jlm/rvsdg/Transformation.cpp
@@ -3,9 +3,11 @@
  * See COPYING for terms of redistribution.
  */
 
-#include "RvsdgModule.hpp"
 #include <jlm/rvsdg/graph.hpp>
+#include <jlm/rvsdg/RvsdgModule.hpp>
 #include <jlm/rvsdg/Transformation.hpp>
+
+#include <fstream>
 
 namespace jlm::rvsdg
 {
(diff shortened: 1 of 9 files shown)


jlm/rvsdg/Transformation.cpp
```diff
@@ -52,13 +54,55 @@ TransformationSequence::Run(
   auto statistics = Statistics::Create(rvsdgModule.SourceFilePath().value());
   statistics->StartMeasuring(rvsdgModule.Rvsdg());
 
-  for (const auto & optimization : Transformations_)
+  size_t numPasses = 0;
+  if (DumpRvsdgDotGraphs_)
+  {
+    DumpDotGraphs(
+        rvsdgModule,
+        statisticsCollector.GetSettings().GetOutputDirectory(),
```
haved, line 62: Instead of `GetSettings().GetOutputDirectory()`, could we add a method to the `StatisticsCollector` called `GetOrCreateOutputDirectory()`, in which we put the logic for creating the statistics output folder (which is currently in `StatisticsCollector::CreateOutputFile()`?

The current `StatisticsCollector::CreateOutputFile()` always includes the module name and a random string in the filename, so I get the point of making the file path manually instead. It would be nice to only create the output folder if anyone actually attempts to put anything there, though.

jlm/rvsdg/Transformation.cpp
```diff
@@ -52,13 +54,55 @@ TransformationSequence::Run(
...
+
+  util::graph::Writer graphWriter;
+  DotWriter_.WriteGraphs(graphWriter, rvsdgModule.Rvsdg().GetRootRegion(), true);
+
+  std::stringstream filePath;
+  filePath << outputDir.to_str() << "/" << std::setw(3) << std::setfill('0') << numPass << "-"
+           << passName << ".dot";
+
+  std::ofstream outputFile;
+  outputFile.open(filePath.str().c_str());
```
haved, line 103: Do we need `.c_str()` here? `ofstream::open()` can take a `const std::string &`.

jlm/tooling/CommandLine.cpp
```diff
@@ -688,6 +688,8 @@ JlmOptCommandLineParser::ParseCommandLineArguments(int argc, const char * const
       cl::value_desc("file"));
 
   const auto statisticsDirectoryDefault = util::FilePath::TempDirectoryPath().Join("jlm");
+  statisticsDirectoryDefault.CreateDirectory();
```
haved, line 691: This should not be here, no?
This path is only the default output folder. If the user specifies a different one we should make that one instead. Also, I think it is cleaner to only make the directory if there is a pass that wants to write something there. See my comment above

phate: Removed.

jlm/rvsdg/Transformation.cpp
phate: I fully agree. Done

jlm/util/Statistics.hpp
```diff
@@ -339,12 +339,15 @@ class StatisticsCollectorSettings final
...
   /**
    * @return the directory used for outputting statistics and debug output files.
-   * If no output directory is given, an assertion failure occurs.
+   *
+   * \note If no output directory path is given, an assertion failure occurs. If the directory
+   * does not exist yet, it is created.
    */
   [[nodiscard]] const FilePath &
-  GetOutputDirectory() const noexcept
+  GetOrCreateOutputDirectory() const noexcept
```
haved, line 347: I think it might still make sense to keep a pure getter, if only to handle the comment I wrote above

haved (commented):
I think completely replacing the getter with `GetOrCreate` might be a bit too much. See comment @phate

jlm/util/Statistics.hpp
phate: Added it back.

haved:
@phate looks good, but I do not dare touch the update branch button

PR #1211, opened 2025-09-02
Add Alias Analysis Precision Evaluator

haved:
This is the PrecisionEvaluator for providing statistics about a given AliasAnalysis and its responses.

It also contains configuration options for dumping precision metrics per function to a file, as well as dumping a DotWriter graph that contains nodes for pointers, and edges representing MustAlias and MayAlias-relations between nodes.

This code should probably have some tests, as there are some moving parts regarding de-duplication of pointers. Even when pointer deduplication is not requested, it still aggregates pointers to make the runtime not explode, but tracks the aggregated values such that the final results are not affected.

### DIFF
--- a/jlm/llvm/ir/operators/Load.cpp
+++ b/jlm/llvm/ir/operators/Load.cpp
@@ -495,12 +495,13 @@ LoadNonVolatileOperation::NormalizeIOBarrierAllocaAddress(
   if (!ioBarrierOperation)
     return std::nullopt;
 
-  const auto barredAddress = IOBarrierOperation::BarredInput(*ioBarrierNode).origin();
-  if (!rvsdg::IsOwnerNodeOperation<AllocaOperation>(*barredAddress))
+  auto & barredAddress = *IOBarrierOperation::BarredInput(*ioBarrierNode).origin();
+  const auto & tracedAddress = rvsdg::TraceOutputIntraProcedurally(barredAddress);
+  if (!rvsdg::IsOwnerNodeOperation<AllocaOperation>(tracedAddress))
     return std::nullopt;
 
   auto & loadNode = CreateNode(
-      *barredAddress,
+      barredAddress,
       { std::next(operands.begin()), operands.end() },
       operation.GetLoadedType(),
       operation.GetAlignment());
--- a/jlm/llvm/ir/operators/Store.cpp
+++ b/jlm/llvm/ir/operators/Store.cpp
@@ -254,12 +254,13 @@ StoreNonVolatileOperation::NormalizeIOBarrierAllocaAddress(
   if (!ioBarrierOperation)
     return std::nullopt;
 
-  const auto barredAddress = IOBarrierOperation::BarredInput(*ioBarrierNode).origin();
-  if (!rvsdg::IsOwnerNodeOperation<AllocaOperation>(*barredAddress))
+  auto & barredAddress = *IOBarrierOperation::BarredInput(*ioBarrierNode).origin();
+  const auto & tracedAddress = rvsdg::TraceOutputIntraProcedurally(barredAddress);
+  if (!rvsdg::IsOwnerNodeOperation<AllocaOperation>(tracedAddress))
     return std::nullopt;
 
   auto & storeNode = CreateNode(
-      *barredAddress,
+      barredAddress,
       *value,
       { std::next(operands.begin(), 2), operands.end() },
       operation.GetAlignment());
--- a/tests/jlm/llvm/ir/operators/LoadTests.cpp
+++ b/tests/jlm/llvm/ir/operators/LoadTests.cpp
@@ -13,6 +13,7 @@
 #include <jlm/llvm/ir/operators/MemoryStateOperations.hpp>
 #include <jlm/llvm/ir/operators/Store.hpp>
 #include <jlm/llvm/ir/RvsdgModule.hpp>
+#include <jlm/rvsdg/gamma.hpp>
 #include <jlm/rvsdg/NodeNormalization.hpp>
 #include <jlm/rvsdg/view.hpp>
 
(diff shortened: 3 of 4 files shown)


jlm/llvm/opt/alias-analyses/AliasAnalysisPrecisionEvaluator.hpp
```diff
@@ -0,0 +1,296 @@
...
+   * Default: false
+   * @param loadsConsideredClobbers if true, loads will be clobbers.
+   */
+  void
+  SetLoadsConsideredClobbers(bool loadsConsideredClobbers) noexcept
+  {
+    LoadsConsideredClobbers_ = loadsConsideredClobbers;
+  }
+
+  bool
```
phate, line 53: [[nodiscard]]

jlm/llvm/opt/alias-analyses/AliasAnalysisPrecisionEvaluator.hpp
```diff
@@ -0,0 +1,296 @@
...
+   *
+   * @param perFunctionOutputEnabled if true, evaluation will print output per function to a file.
+   */
+  void
+  SetPerFunctionOutputEnabled(bool perFunctionOutputEnabled) noexcept
+  {
+    PerFunctionOutputEnabled_ = perFunctionOutputEnabled;
+  }
+
+  bool
```
phate, line 111: [[nodiscard]]

jlm/llvm/opt/alias-analyses/AliasAnalysisPrecisionEvaluator.cpp
```diff
@@ -0,0 +1,501 @@
+/*
+ * Copyright 2025 Håvard Krogstie <krogstie.havard@gmail.com>
+ * See COPYING for terms of redistribution.
+ */
+
+#include "jlm/llvm/ir/operators/Load.hpp"
+#include "jlm/llvm/ir/operators/Store.hpp"
```
phate, line 7: <>

jlm/llvm/opt/alias-analyses/AliasAnalysisPrecisionEvaluator.cpp
```diff
@@ -0,0 +1,501 @@
...
+{
+  for (auto & node : region.Nodes())
+  {
+    if (auto lambda = dynamic_cast<const rvsdg::LambdaNode *>(&node))
+    {
+      EvaluateFunction(*lambda, aliasAnalysis);
+    }
+    else if (auto structural = dynamic_cast<const rvsdg::StructuralNode *>(&node))
+    {
+      for (size_t n = 0; n < structural->nsubregions(); n++)
```
phate, line 175: for (auto subregion : structural->Subregions())

jlm/llvm/opt/alias-analyses/AliasAnalysisPrecisionEvaluator.cpp
```diff
@@ -0,0 +1,501 @@
...
+      {
+        EvaluateAllFunctions(*structural->subregion(n), aliasAnalysis);
+      }
+    }
+  }
```
phate, line 180: This entire function is a bit wasteful, no? It traverses the entire region tree, but ultimately only cares about lambda nodes. This is a debug pass, so it ultimately does not matter. I leave it to you whether you short-circuit this function and not traverse into gamma, theta, and delta node subregions.

jlm/llvm/opt/alias-analyses/AliasAnalysisPrecisionEvaluator.cpp
```diff
@@ -0,0 +1,501 @@
...
+  if (const auto load = dynamic_cast<const LoadOperation *>(&node.GetOperation()))
+  {
+    const auto size = GetTypeSize(*load->GetLoadedType());
+    CollectPointer(LoadOperation::AddressInput(node).origin(), size, AreLoadsConsideredClobbers());
+  }
+  else if (auto store = dynamic_cast<const StoreOperation *>(&node.GetOperation()))
+  {
+    const auto size = GetTypeSize(store->GetStoredType());
+    CollectPointer(StoreOperation::AddressInput(node).origin(), size, true);
+  }
```
phate, line 291: This code will silently change if we ever get another subclass on the LoadOperation or StoreOperation. I would rather have you spelling it out for all the individual nodes and then having an unreachable here. Not saying the Load/Store-Operation subclasses will ever get more, but we are not done modeling everything from LLVM yet.

jlm/llvm/opt/alias-analyses/AliasAnalysisPrecisionEvaluator.cpp
```diff
@@ -0,0 +1,501 @@
...
+}
+
+void
+AliasAnalysisPrecisionEvaluator::CollectPointersFromStructuralNode(
+    const rvsdg::StructuralNode & node)
+{
+  for (size_t n = 0; n < node.nsubregions(); n++)
```
phate, line 298: for (auto & subregion : Subregions())

jlm/util/Statistics.hpp
```diff
@@ -34,6 +34,7 @@ class Statistics
   {
     FirstEnumValue, // must always be the first enum value, used for iteration
 
+    AliasAnalysisPrecisionEvaluation,
```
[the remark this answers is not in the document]
haved: Good catch

jlm/util/Statistics.cpp
```diff
@@ -18,6 +18,7 @@ static const util::BijectiveMap<Statistics::Id, std::string_view> &
 GetStatisticsIdNames()
 {
   static util::BijectiveMap<Statistics::Id, std::string_view> mapping = {
+    { Statistics::Id::AliasAnalysisPrecisionEvaluation, "AliasAnalysisPrecisionEvaluation" },
```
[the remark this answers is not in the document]
haved: Fixed

jlm/llvm/opt/alias-analyses/AliasAnalysisPrecisionEvaluator.hpp
```diff
@@ -0,0 +1,296 @@
...
+ * Copyright 2025 Håvard Krogstie <krogstie.havard@gmail.com>
+ * See COPYING for terms of redistribution.
+ */
+
+#ifndef JLM_LLVM_OPT_ALIAS_ANALYSES_PRECISIONEVALUATOR_HPP
+#define JLM_LLVM_OPT_ALIAS_ANALYSES_PRECISIONEVALUATOR_HPP
+
+#include "jlm/rvsdg/lambda.hpp"
+#include "jlm/rvsdg/RvsdgModule.hpp"
+#include "jlm/util/GraphWriter.hpp"
```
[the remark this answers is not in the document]
haved: I should look for a way to make CLion do this automatically

jlm/llvm/opt/alias-analyses/AliasAnalysisPrecisionEvaluator.cpp
phate: You are right, I misread the code. Thanks for making it clearer though.

jlm/llvm/opt/alias-analyses/AliasAnalysisPrecisionEvaluator.cpp
haved: I added asserts to yell at us if we ever start adding new subclasses of `LoadOperation` or `StoreOperation`.

haved:
image
Just for fun (and as a sanity check), here is the result of plotting some of the precision metrics. They look just like in the paper, except for now using our own LocalAA instead of LLVM's BasicAA.

jlm/llvm/opt/alias-analyses/AliasAnalysisPrecisionEvaluator.cpp
```diff
@@ -0,0 +1,502 @@
+/*
+ * Copyright 2025 Håvard Krogstie <krogstie.havard@gmail.com>
+ * See COPYING for terms of redistribution.
+ */
+
+#include "jlm/rvsdg/Phi.hpp"
```
phate, line 6: <>

This is possible to configure in Clion. See here: https://youtrack.jetbrains.com/issue/CPP-41725/Option-to-disable-c-c-header-auto-import-is-missing

phate (approved):
Just another question: We currently handle only load and stores in CollectPointersFromSimpleNode(). What about other nodes, such as memcpy, free, or call nodes?

I guess in the traditional evaluation of the papers, they are not really considered?

PR #1212, opened 2025-09-03
Extend LoadNonVolatileOperation::NormalizeIOBarrierAllocaAddress()

phate:
Extends the NormalizeIOBarrierAlloocaAddress normalization to trace the input address of the IOBarrier through gamma and theta nodes.

### DIFF
--- a/jlm/llvm/ir/operators/Load.cpp
+++ b/jlm/llvm/ir/operators/Load.cpp
@@ -495,12 +495,13 @@ LoadNonVolatileOperation::NormalizeIOBarrierAllocaAddress(
   if (!ioBarrierOperation)
     return std::nullopt;
 
-  const auto barredAddress = IOBarrierOperation::BarredInput(*ioBarrierNode).origin();
-  if (!rvsdg::IsOwnerNodeOperation<AllocaOperation>(*barredAddress))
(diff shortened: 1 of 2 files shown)


jlm/llvm/ir/operators/Load.cpp
```diff
@@ -495,12 +495,13 @@ LoadNonVolatileOperation::NormalizeIOBarrierAllocaAddress(
   if (!ioBarrierOperation)
     return std::nullopt;
 
-  const auto barredAddress = IOBarrierOperation::BarredInput(*ioBarrierNode).origin();
-  if (!rvsdg::IsOwnerNodeOperation<AllocaOperation>(*barredAddress))
+  auto & barredAddress = *IOBarrierOperation::BarredInput(*ioBarrierNode).origin();
+  const auto & tracedAddress = rvsdg::TraceOutputIntraProcedurally(barredAddress);
```
haved, line 499: I would like to have an llvm specific version of `TraceOutputIntraProcedurally` that also handles LLVM specific instructions.

Then I would also go one step further and add one that traces through GEPs (at least inbound GEPs), since we do not actually care about the offset into the alloca, just the fact that it is an alloca.

This can be done later, of course

PR #1216, opened 2025-09-04
Add statistics about memory states to the MemoryStateEncoder

haved:
This PR counts the number of memory states, and what type of `MemoryNode` they correspond to, for the following:
 - Region arguments (Intra-procedural only)
 - Loads (including volatile)
 - Stores (including volatile)
 - Call entry merges

It also tracks the largest amount of memory states used by a single such entity.

Finally, it tracks how many of the memory states correspond to `MemoryNode`s that are marked as module escaping.

### DIFF
--- a/jlm/llvm/opt/alias-analyses/MemoryStateEncoder.cpp
+++ b/jlm/llvm/opt/alias-analyses/MemoryStateEncoder.cpp
@@ -15,11 +15,140 @@
 namespace jlm::llvm::aa
 {
 
+/**
+ * \brief Helper struct for counting up MemoryNodes, among some set of entities that use them
+ */
+struct MemoryStateTypeCounter final
+{
+  // The number of entities that have been counted
+  uint64_t NumEntities = 0;
+
+  // Count of total memory states, divided by MemoryNode type
+  uint64_t NumAllocas = 0;
+  uint64_t NumMallocs = 0;
+  uint64_t NumDeltas = 0;
+  uint64_t NumImports = 0;
+  uint64_t NumLambdas = 0;
+  uint64_t NumExternal = 0;
+
+  // Among the MemoryNodes counted above, how many were not escaping
+  uint64_t NumNonEscaped = 0;
+
+  // Remember the single entity with the highest number of memory states
+  uint64_t MaxMemoryStateEntity = 0;
+  // Do the same, but only include non-escaped MemoryNodes
+  uint64_t MaxNonEscapedMemoryStateEntity = 0;
+
+  void
+  CountEntity(
+      uint64_t numAllocas,
+      uint64_t numMallocs,
+      uint64_t numDeltas,
+      uint64_t numImports,
+      uint64_t numLambdas,
+      uint64_t numExternal,
+      uint64_t numNonEscaped)
+  {
+    NumEntities++;
+
+    NumAllocas += numAllocas;
+    NumMallocs += numMallocs;
+    NumDeltas += numDeltas;
+    NumImports += numImports;
+    NumLambdas += numLambdas;
+    NumExternal += numExternal;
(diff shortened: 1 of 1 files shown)


jlm/llvm/opt/alias-analyses/MemoryStateEncoder.cpp
```diff
@@ -15,11 +15,140 @@
...
+      else if (dynamic_cast<const PointsToGraph::DeltaNode *>(memoryNode))
+        numDeltas++;
+      else if (dynamic_cast<const PointsToGraph::ImportNode *>(memoryNode))
+        numImports++;
+      else if (dynamic_cast<const PointsToGraph::LambdaNode *>(memoryNode))
+        numLambdas++;
+      else if (dynamic_cast<const PointsToGraph::ExternalMemoryNode *>(memoryNode))
+        numExternal++;
+      else
+        JLM_UNREACHABLE("Unknown MemoryNode type");
```
phate, line 101: I think there is also an is<> method for memory nodes defined.

jlm/llvm/opt/alias-analyses/MemoryStateEncoder.cpp
```diff
@@ -15,11 +15,140 @@
...
 
+/**
+ * \brief Helper struct for counting up MemoryNodes, among some set of entities that use them
+ */
+struct MemoryStateTypeCounter final
+{
+  // The number of entities that have been counted
+  uint64_t NumEntities = 0;
+
+  // Count of total memory states, divided by MemoryNode type
```
phate, line 26: "total memory states, separated by MemoryNode type"

jlm/llvm/opt/alias-analyses/MemoryStateEncoder.cpp
```diff
@@ -872,6 +1088,10 @@ MemoryStateEncoder::EncodeGammaEntry(rvsdg::GammaNode & gammaNode)
   auto & stateMap = Context_->GetRegionalizedStateMap();
   auto memoryNodes = Context_->GetModRefSummary().GetGammaEntryNodes(gammaNode);
 
+  // Count the memory state arguments once per subregion
+  for (size_t i = 0; i < gammaNode.nsubregions(); i++)
```
phate, line 1092: for (auto & subregion : gammaNode.Subregions())

phate (approved):
Three minor comments.

jlm/llvm/opt/alias-analyses/MemoryStateEncoder.cpp
haved: Hahah, yes, I should have remembered from the last two times you pointed this out :)

jlm/llvm/opt/alias-analyses/MemoryStateEncoder.cpp
haved: I could not find one, so I added one

haved:
I fixed your comments, @phate

I added an `is<>` for PointsToGraph Nodes, as I could not find one.

I also replaced the surrounding uses of `for(size_t i = 0; i < gammaNode.nsubregions(); i++)`, while I was at it

jlm/llvm/opt/alias-analyses/MemoryStateEncoder.cpp
phate: There is already one: https://github.com/phate/jlm/blob/342848731a725be9648cb99fb5370596a325e0bc/jlm/llvm/opt/alias-analyses/PointsToGraph.hpp#L535C1-L542C4

phate (changes requested):
There was already an Is<>() method. See comment.

PR #1219, opened 2025-09-05
Add compatibility changes to better support MacOS

andreaslhjulstad:
Changes some build configuration files to fix compatibility issues on LLVM across OS'es. 

Specifically, this fixes an error where the `ar` command would give an error "no archive members specified", when HLS and MLIR were not enabled. This happened when running `make all` with LLVM@18 installed from HomeBrew and the default MacOS clang compiler from Xcode (see full error message below).

```bash
c++ -c --std=c++17 -Wall -Wpedantic -Wextra -Wno-unused-parameter -Werror -Wfatal-errors -gdwarf-4 -g -O3  -I. -Itests -I/opt/homebrew/Cellar/llvm@18/18.1.8/include  -DSTDC_CONSTANT_MACROS -DSTDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS    -o build/tools/jlc/jlc.la tools/jlc/jlc.cpp
rm -f build/libmlir.a
ar cqv build/libmlir.a 
ar: no archive members specified
usage:  ar -d [-TLsv] archive file ...
        ar -m [-TLsv] archive file ...
        ar -m [-abiTLsv] position archive file ...
        ar -p [-TLsv] archive [file ...]
        ar -q [-cTLsv] archive file ...
        ar -r [-cuTLsv] archive file ...
        ar -r [-abciuTLsv] position archive file ...
        ar -t [-TLsv] archive [file ...]
        ar -x [-ouTLsv] archive [file ...]
make: *** [build/libmlir.a] Error 1
```

Also fixes an error where bash would complain about `CIRCT_LDFLAGS_ARRAY` being unbound when running `./configure.sh` on MacOS(see error message below):

```bash
$ ./configure.sh                                        
./configure.sh: line 176: CIRCT_LDFLAGS_ARRAY[*]: unbound variable
```

### DIFF
--- a/jlm/llvm/backend/IpGraphToLlvmConverter.cpp
+++ b/jlm/llvm/backend/IpGraphToLlvmConverter.cpp
@@ -337,12 +337,12 @@ IpGraphToLlvmConverter::convert_phi(
   auto & llvmContext = Context_->llvm_module().getContext();
   auto & typeConverter = Context_->GetTypeConverter();
 
-  if (rvsdg::is<IOStateType>(phi.type()))
+  if (rvsdg::is<IOStateType>(phi.Type()))
     return nullptr;
-  if (rvsdg::is<MemoryStateType>(phi.type()))
+  if (rvsdg::is<MemoryStateType>(phi.Type()))
     return nullptr;
 
-  auto t = typeConverter.ConvertJlmType(phi.type(), llvmContext);
+  auto t = typeConverter.ConvertJlmType(*phi.Type(), llvmContext);
   return builder.CreatePHI(t, op.narguments());
 }
 
--- a/jlm/llvm/frontend/ControlFlowRestructuring.cpp
+++ b/jlm/llvm/frontend/ControlFlowRestructuring.cpp
@@ -127,7 +127,7 @@ restructure_loop_entry(
 {
   size_t n = 0;
   std::unordered_map<llvm::ControlFlowGraphNode *, size_t> indices;
-  for (auto & node : s.enodes())
+  for (auto & node : s.EntryNodes())
   {
     new_ne->add_outedge(node);
     indices[node] = n++;
@@ -136,7 +136,7 @@ restructure_loop_entry(
   if (ev)
     append_branch(new_ne, ev);
 
-  for (auto & edge : s.eedges())
+  for (auto & edge : s.EntryEdges())
   {
     auto os = edge->sink();
     edge->divert(new_ne);
@@ -166,7 +166,7 @@ restructure_loop_exit(
     This edge is never taken at runtime, but fixes the CFGs structure at compile-time such
     that we can create an RVSDG.
   */
-  if (s.nxedges() == 0)
+  if (s.NumExitEdges() == 0)
   {
     new_nx->add_outedge(exit);
     return;
@@ -174,7 +174,7 @@ restructure_loop_exit(
 
   size_t n = 0;
   std::unordered_map<llvm::ControlFlowGraphNode *, size_t> indices;
-  for (auto & node : s.xnodes())
+  for (auto & node : s.ExitNodes())
   {
     new_nx->add_outedge(node);
     indices[node] = n++;
@@ -183,7 +183,7 @@ restructure_loop_exit(
   if (xv)
     append_branch(new_nx, xv);
 
-  for (auto & edge : s.xedges())
+  for (auto & edge : s.ExitEdges())
   {
     auto os = edge->sink();
     edge->divert(new_nr);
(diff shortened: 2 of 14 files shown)


haved:
@phate this PR changes the order of libraries, but does not cause issues. I finally understand why.

When a library A defines a function a(), which is used by library B, B should come before A in the linker invocation. Otherwise the linker will finish processing A without knowing that a() is used, and forget that a() even exists. Once it starts processing B it will not remember that a() was defined in A, and give undefined reference errors.

However, if A also uses a() itself, it will not be forgotten, so usage of a() in B will also be fine. 

This is what happened in the case of the `MemoryStateJoinOperation`. I tried using it in `libmlir.a`, before it was used anywhere in `libllvm.a`. That is why it failed. Now that the class is used within `libllvm.a`, it is no longer necessary for `libmlir.a` to come before `libllvm.a`

You learn something about linking every day ;)

phate:
@andreaslhjulstad @haved I am bit confused by the library change in this PR. Why do we need to put the libraries all in a sudden in guards? We have builds in the CI that are bare minimum, i.e., do not use the MLIR and HLS dialect, and they build perfectly. So why does this all in a sudden needs to be guarded?

caleridas:
> @phate this PR changes the order of libraries, but does not cause issues. I finally understand why.
> 
> When a library A defines a function a(), which is used by library B, B should come before A in the linker invocation. Otherwise the linker will finish processing A without knowing that a() is used, and forget that a() even exists. Once it starts processing B it will not remember that a() was defined in A, and give undefined reference errors.
> 
> However, if A also uses a() itself, it will not be forgotten, so usage of a() in B will also be fine.
> 
> This is what happened in the case of the `MemoryStateJoinOperation`. I tried using it in `libmlir.a`, before it was used anywhere in `libllvm.a`. That is why it failed. Now that the class is used within `libllvm.a`, it is no longer necessary for `libmlir.a` to come before `libllvm.a`
> 
> You learn something about linking every day ;)
> 
> FYI @caleridas

Yes I am aware of the ordering and it is on principle true for all linkers. I will take a look, I think in general CIRCT ldflags should be handled a bit more like other dependent libs, let me go through the PR...

configure.sh
```diff
@@ -178,7 +178,7 @@ CXXFLAGS=${CXXFLAGS-} ${CXXFLAGS_COMMON} ${CXXFLAGS_TARGET} ${CXXFLAGS_NO_COMMEN
 CPPFLAGS=${CPPFLAGS-} ${CPPFLAGS_COMMON} ${CPPFLAGS_LLVM} ${CPPFLAGS_ASSERTS} ${CPPFLAGS_CIRCT} ${CPPFLAGS_MLIR}
 ENABLE_HLS=${ENABLE_HLS}
 CIRCT_PATH=${CIRCT_PATH}
-CIRCT_LDFLAGS=${CIRCT_LDFLAGS_ARRAY[*]}
+CIRCT_LDFLAGS=${CIRCT_LDFLAGS_ARRAY[*]-}
```
caleridas, line 181: what does the - accomplish for array expansion? I have never seen that (I think an empty array here is just fine)

andreaslhjulstad:
> @andreaslhjulstad @haved I am bit confused by the library change in this PR. Why do we need to put the libraries all in a sudden in guards? We have builds in the CI that are bare minimum, i.e., do not use the MLIR and HLS dialect, and they build perfectly. So why does this all in a sudden needs to be guarded?
> 
> EDIT: Well, duh, it says in the PR message why it is necessary. Hmm, could we rather guard them "in place" to not change the order of the libraries:
> 
> ```
> libtooling_TEST_LIBS = \
>     libtooling \
> ifdef ENABLE_HLS
>     libhls \
> endif
>     ...
> ```

I don't have nearly enough experience with linking and libraries in C++ to say for sure. But the error seems to come from a difference in `GNU ar`and the `BSD ar` (on macOS) behave when no files are provided. In `GNU ar`, an empty archive is created, but in `BSD ar`, it results in an error. So when it tries to create the archive for `libmlir` and `libhls`, and they don't exist, it crashes.

Thus, the simplest fix I could find was to conditionally add the dialects. I see now that it might not be that simple 😅. However, I also think this is an overall more robust approach, since the dialects are conditionally added elsewhere, like e.g. in the Makefile. 

Thanks for your feedback!

configure.sh
caleridas: I thought the "expand to empty string" is implied for empty arrays, admittedly I could be mistaken.

andreaslhjulstad: That's the thing! The script runs fine on Linux, and probably does expand it into an empty string if the array is undefined, but on macOS, it crashes. 

This might have to do with the version of bash. As I can see now that I am running bash 3.2.57 (which is from 2007), since that is the default that macOS ships with.

caleridas: Ah thanks for the context, lgtm (maybe add comment next to it)

PR #1220, opened 2025-09-09
Add scalar evolution pass with method for finding induction variables

andreaslhjulstad:
This PR adds a simple barebones ScalarEvolution class as well as an optimization command to run it. 

Right now, all it does is detect the simplest cases of induction variables in theta-nodes. A (basic) induction variable is a variable which changes with a constant value each iteration.

It uses a fixed-point analysis with a bottom-up approach where it starts with all loop variables in a set, and systematically removes nodes that are not induction variables. At the end, only the induction variables remain in the set.

To run the pass, use 
```bash
$ jlm-opt --ScalarEvolution -o test test.ll
```


jlm/tooling/Command.cpp
```diff
@@ -436,6 +437,8 @@ JlmOptCommand::CreateTransformation(
         CommandLineOptions_.GetRvsdgTreePrinterConfiguration());
   case JlmOptCommandLineOptions::OptimizationId::ThetaGammaInversion:
     return std::make_unique<llvm::LoopUnswitching>();
+  case JlmOptCommandLineOptions::OptimizationId::ScalarEvolution:
+    return std::make_unique<llvm::ScalarEvolution>();
   default:
```
phate, line 442: The optimizations are ordered alphabetically.

tests/c-tests/test-simple-induction-variable.c
```diff
@@ -0,0 +1,11 @@
...
+
+int main()
+{
+  int i = 0;
+  do {
+    i = 1 + i;
+  } while (i < 5);
+
+  return 0;
+}
```
phate, line 11: The test is a bit useless as it is currently not automatically tested. I rather suggest that you add a unit test that tests e.g. your FindInductionVariable() method.

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -0,0 +1,209 @@
...
+    if (const auto simpleNode = dynamic_cast<rvsdg::SimpleNode *>(node))
+    {
+      if (!rvsdg::is<llvm::IntegerBinaryOperation>(simpleNode))
+      {
+        // TODO: Handle nodes that are not binary integer operations
+        continue;
+      }
+      // Check for basic induction variables (BIV)
+
+      assert(simpleNode->ninputs() == 2); // Assert because this should never happen
```
phate, line 142: Comment is useless. It does not provide more information than what the code already states.

jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -0,0 +1,75 @@
...
+      std::vector<rvsdg::ThetaNode::LoopVar> loopVars)
+  {
+    for (auto & loopVar : loopVars)
+    {
+      if (input->origin() == loopVar.pre)
+      {
+        return loopVar;
+      }
+    }
+    return std::nullopt;
```
phate, line 63: I think this method can be implemented more efficiently. Have a look at [MapPreLoopVar()](https://github.com/phate/jlm/blob/5f36b59992b4f248584a35e40d94a6fa68528690/jlm/rvsdg/theta.cpp#L127C1-L135C2). It should be possible to use TryGetRegionParentNode() for this. I do not think there is a need to iterate through all loop variables.

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -0,0 +1,209 @@
...
+  }
+};
+
+void
+ScalarEvolution::Run(
+    rvsdg::RvsdgModule & rvsdgModule,
+    util::StatisticsCollector & statisticsCollector)
+{
+  auto statistics = Statistics::Create(rvsdgModule.SourceFilePath().value());
+  statistics->Start();
```
phate, line 52: InductionVariableMap_ needs to be cleared at the beginning of this function. It should be possible to invoke the Run() method multiple times from the same instance.

phate (changes requested):
Looks good. I left a few minor comments.

PR #1221, opened 2025-09-09
Change traversers and remove trackers

caleridas:
Remove dependence on tracker and node depth for both top-down and bottom-up traversers. Replace it with normal BFS graph traversal instead.

Trackers are not needed anymore anywhere. Remove them as they rely on node depth (which is also supposed to go away).

Traversers are fully deterministic and particularly independent of any addresses; it depends only:
- the order in which nodes are registered as "top" or "bottom" nodes in graph
- the order in which successor or predecessor nodes are attached to outputs and inputs

### DIFF
--- a/jlm/hls/backend/rvsdg2rhls/GammaConversion.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/GammaConversion.cpp
@@ -5,6 +5,8 @@
 
 #include <jlm/hls/backend/rvsdg2rhls/GammaConversion.hpp>
 #include <jlm/hls/ir/hls.hpp>
+#include <jlm/rvsdg/gamma.hpp>
+#include <jlm/rvsdg/RvsdgModule.hpp>
 #include <jlm/rvsdg/theta.hpp>
 #include <jlm/rvsdg/traverser.hpp>
 
@@ -164,8 +166,14 @@ ConvertGammaNodesInRegion(rvsdg::Region & region)
   }
 }
 
+GammaNodeConversion::~GammaNodeConversion() noexcept = default;
+
+GammaNodeConversion::GammaNodeConversion()
+    : Transformation("GammaNodeConversion")
+{}
+
 void
-ConvertGammaNodes(llvm::RvsdgModule & rvsdgModule)
+GammaNodeConversion::Run(rvsdg::RvsdgModule & rvsdgModule, util::StatisticsCollector &)
 {
   ConvertGammaNodesInRegion(rvsdgModule.Rvsdg().GetRootRegion());
 }
--- a/jlm/hls/backend/rvsdg2rhls/GammaConversion.hpp
+++ b/jlm/hls/backend/rvsdg2rhls/GammaConversion.hpp
@@ -6,8 +6,7 @@
 #ifndef JLM_HLS_BACKEND_RVSDG2RHLS_GAMMACONVERSION_HPP
 #define JLM_HLS_BACKEND_RVSDG2RHLS_GAMMACONVERSION_HPP
 
-#include <jlm/llvm/ir/RvsdgModule.hpp>
-#include <jlm/rvsdg/gamma.hpp>
+#include <jlm/rvsdg/Transformation.hpp>
 
 namespace jlm::hls
 {
@@ -17,8 +16,28 @@ namespace jlm::hls
  *
  * @param rvsdgModule The RVSDG module the transformation is performed on.
  */
-void
-ConvertGammaNodes(llvm::RvsdgModule & rvsdgModule);
+class GammaNodeConversion final : public rvsdg::Transformation
+{
+public:
+  ~GammaNodeConversion() noexcept override;
+
+  GammaNodeConversion();
+
+  GammaNodeConversion(const GammaNodeConversion &) = delete;
+
+  GammaNodeConversion &
+  operator=(const GammaNodeConversion &) = delete;
+
+  void
+  Run(rvsdg::RvsdgModule & rvsdgModule, util::StatisticsCollector & statisticsCollector) override;
+
+  static void
+  CreateAndRun(rvsdg::RvsdgModule & rvsdgModule, util::StatisticsCollector & statisticsCollector)
+  {
+    GammaNodeConversion gammaNodeConversion;
+    gammaNodeConversion.Run(rvsdgModule, statisticsCollector);
+  }
+};
 
 }
 
--- a/jlm/hls/backend/rvsdg2rhls/add-buffers.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/add-buffers.cpp
@@ -13,7 +13,7 @@
 namespace jlm::hls
 {
 
-rvsdg::Input *
+static rvsdg::Input *
 FindUserNode(rvsdg::Output * out)
 {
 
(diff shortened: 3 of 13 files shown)


jlm/rvsdg/traverser.hpp
```diff
@@ -92,25 +96,32 @@ class TraversalTracker final
...
+
+  struct State
+  {
+    traversal_nodestate state = traversal_nodestate::ahead;
+    FrontierList::iterator pos = {};
+  };
+
+  Region & region_;
+  std::unordered_map<Node *, State> states_;
+  FrontierList frontier_;
```
phate, line 118: All should start with capital letter.

jlm/rvsdg/traverser.cpp
```diff
@@ -111,9 +117,9 @@ TopDownTraverser::input_change(Input * in, Output *, Output *)
   if (state != traversal_nodestate::ahead)
     return;
 
-  /* make sure node is visited eventually, might now be visited earlier
-  as depth of the node could be lowered */
-  tracker_.set_nodestate(node, traversal_nodestate::frontier);
+  /* node may just have become eligible for visiting, check */
```
phate, line 120: `//`

jlm/rvsdg/traverser.cpp
```diff
@@ -193,6 +199,25 @@ BottomUpTraverser::node_destroy(Node * node)
   for (size_t n = 0; n < node->ninputs(); n++)
   {
     auto producer = TryGetOwnerNode<Node>(*node->input(n)->origin());
+    if (!producer)
+    {
+      continue;
+    }
+    bool successors_visited = true;
+    for (size_t k = 0; k < producer->noutputs(); ++k)
```
phate, line 207: `for (auto & output : producer->Outputs())`

jlm/rvsdg/traverser.cpp
```diff
@@ -223,4 +248,57 @@ BottomUpTraverser::input_change(Input * in, Output * old_origin, Output *)
   tracker_.set_nodestate(node, traversal_nodestate::frontier);
 }
 
+/* traversal tracker implementation */
```
phate, line 251: Remove. It is useless.

haved:
One point regarding determinism: The `TopDownTraverser` uses the `output.Users()` iterator to add successors to the frontier. `Users()` is an `unordered_set`, so if a single output has multiple users, the order in which they are added to the frontier list is not deterministic. One option would be to use the node's region id as a tie-break in these situations.

@caleridas @phate

jlm/rvsdg/traverser.cpp
caleridas: changed

caleridas:
since users are linked list now, traverser is deterministic

I will in a follow-up fix the (minor) complexity issue

caleridas:
I have no clue how this breaks hls, and CI does not spit out any useful error message

@sjalander

sjalander:
insertion_sort.c is one example that fails.

```
mkdir /tmp/jhls-insertion_sort.c-0yg5Px
/usr/lib/llvm-18/bin/clang      -S -emit-llvm -Xclang -disable-O0-optnone -o /tmp/jhls-insertion_sort.c-0yg5Px/insertion_sort-9gNaUn-clang.ll ./src/dynamatic/insertion_sort.c
/usr/lib/llvm-18/bin/llvm-link -S -v -o /tmp/jhls-insertion_sort.c-0yg5Pxmerged.ll /tmp/jhls-insertion_sort.c-0yg5Px/insertion_sort-9gNaUn-clang.ll
/usr/lib/llvm-18/bin/opt -passes=mem2reg -S -o /tmp/jhls-insertion_sort.c-0yg5Pxmerged.m2r.ll /tmp/jhls-insertion_sort.c-0yg5Pxmerged.ll
jlm-hls --extract --hls-function kernel  -o a.out /tmp/jhls-insertion_sort.c-0yg5Pxmerged.m2r.ll
/usr/lib/llvm-18/bin/opt -passes=mem2reg -S -o /tmp/jhls-insertion_sort.c-0yg5Pxfunction.hls.ll a.out.function.ll
jlm-hls -o a.out /tmp/jhls-insertion_sort.c-0yg5Pxfunction.hls.ll
```

The backtrace
```
#0  0x00005555556aa717 in jlm::rvsdg::is<jlm::hls::MemoryRequestOperation> (node=0x5555564ee8c0)
    at ./jlm/rvsdg/node.hpp:838
#1  0x000055555576d37a in jlm::hls::dne (sr=0x5555564d1c40)
    at jlm/hls/backend/rvsdg2rhls/rhls-dne.cpp:440
#2  0x000055555576d58b in jlm::hls::dne (sr=0x55555647d620)
    at jlm/hls/backend/rvsdg2rhls/rhls-dne.cpp:467
#3  0x000055555576d58b in jlm::hls::dne (sr=0x5555564da400)
    at jlm/hls/backend/rvsdg2rhls/rhls-dne.cpp:467
#4  0x0000555555736457 in jlm::hls::decouple_mem_state (region=0x5555564b1990)
    at jlm/hls/backend/rvsdg2rhls/decouple-mem-state.cpp:496
#5  0x000055555573776f in jlm::hls::decouple_mem_state (rm=...)
    at jlm/hls/backend/rvsdg2rhls/decouple-mem-state.cpp:504
#6  0x00005555556ea1dd in jlm::hls::rvsdg2rhls (rhls=..., collector=...)
    at jlm/hls/backend/rvsdg2rhls/rvsdg2rhls.cpp:476
#7  0x00005555555e23bb in main (argc=4, argv=0x7fffffffd918) at tools/jlm-hls/jlm-hls.cpp:90
```

I don't understand why `rvsdg::is<MemoryRequestOperation>(node)` fails.
```
(gdb) f 1
#1  0x000055555576d37a in jlm::hls::dne (sr=0x5555564d1c40)
    at jlm/hls/backend/rvsdg2rhls/rhls-dne.cpp:440
440	        if (rvsdg::is<MemoryRequestOperation>(node))
(gdb) l
435	    changed = false;
436	    for (auto & node : rvsdg::BottomUpTraverser(sr))
437	    {
438	      if (node->IsDead())
439	      {
440	        if (rvsdg::is<MemoryRequestOperation>(node))
441	        {
442	          // TODO: fix this once memory connections are explicit
443	          continue;
444	        }
```

sjalander:
I've not managed to find the bug, but I've been able to reduce it to the following transformations as the culprit:
```
  MemoryStateSeparation::CreateAndRun(rhls, collector);
  ConvertGammaNodes(rhls);
  ConvertThetaNodes(rhls);
  dne(rhls);
```

caleridas:
> @phate @caleridas Thanks for the hint @phate. Those two changes are handled correctly as the traversal is restarted due to the 'break'. It turns out that other modifications also break the traversal. I changed it such that any change restarts the traversal. It's not ideal in terms of efficiency, but given the input code size used for HLS it's not detrimental.

"breaking traversal" -- can you explain more precisely? I "suspect" it is breaking by "looking & replacing in the 'wrong' direction".

however I am curious if is it possible to formulate some conditions on a form of traversal that would be suitable for the kind of the transform, if you can explain that I would be interested in trying to see if I can come up with something that addresses the issue without this "restart until hitting fixpoint" pattern?

caleridas:
> @caleridas I would not sweat about it. The entire RhlsDNE transformation is supposed to disappear eventually. The transformation is also not exactly expressed very cleverly in the first place. It seems to work now. Let's move on, we have bigger fish to fry in my opinion.

I think I found this "retry until no change anymore" pattern also in other places (we talked about normalizers a while ago). I would interpret this as an underlying problem of the form "there are transforms that have a structure that is not satisfied by unidirectional traversal", and if that is the case I think it would indeed be interesting to at least understand it.

phate (approved):
LGTM

PR #1231, opened 2025-09-13
Users linked list

caleridas:
FYI: linked list for users of outputs


jlm/rvsdg/node.hpp
```diff
@@ -243,12 +247,237 @@ class Output
   friend class rvsdg::Region;
 
 public:
-  using UserIterator = util::PtrIterator<Input, std::unordered_set<Input *>::iterator>;
-  using UserConstIterator =
-      util::PtrIterator<const Input, std::unordered_set<Input *>::const_iterator>;
+  class UserIterator
```
phate, line 250: It was not possible to reuse [Input::Iterator](https://github.com/phate/jlm/blob/cf502f1b85c780970c00cf4c3537a7f069402e08/jlm/rvsdg/node.hpp#L83)?

jlm/rvsdg/node.hpp
```diff
@@ -243,12 +247,237 @@ class Output
...
+    operator!=(const UserIterator & a, const UserIterator & b)
+    {
+      return a.ptr_ != b.ptr_;
+    };
+
+  private:
+    Input * ptr_;
+  };
+
+  class UserConstIterator
```
phate, line 306: Same [here](https://github.com/phate/jlm/blob/cf502f1b85c780970c00cf4c3537a7f069402e08/jlm/rvsdg/node.hpp#L149C1-L149C28).

jlm/rvsdg/node.hpp
```diff
@@ -243,12 +247,237 @@ class Output
...
+    empty() const noexcept
+    {
+      return output_->FirstUser_ == nullptr;
+    }
+
+  private:
+    const Output * output_;
+  };
+
+  class UsersConstRange
```
phate, line 428: What about [iterator range](https://github.com/phate/jlm/blob/cf502f1b85c780970c00cf4c3537a7f069402e08/jlm/util/iterator_range.hpp#L19C1-L19C20)?

jlm/rvsdg/node.hpp
```diff
@@ -487,7 +716,9 @@ class Output
   size_t index_;
   std::variant<Node *, Region *> Owner_;
   std::shared_ptr<const rvsdg::Type> Type_;
-  std::unordered_set<jlm::rvsdg::Input *> users_;
+  Input * FirstUser_ = nullptr;
+  Input * LastUser_ = nullptr;
+  std::size_t NumUsers_ = 0;
```
phate, line 721: What about our [intrusive-list data structure](https://github.com/phate/jlm/blob/master/jlm/util/intrusive-list.hpp)?

phate (changes requested):
I generally do not have a problem, but do wonder why you did not use the already present data structures? Are they not applicable for this case, or were you just not aware of them?

jlm/rvsdg/node.hpp
caleridas: changed it

caleridas: sorry, did not change, but Input::Iterator does something else -- as discussed should move somewhere else

jlm/rvsdg/node.hpp
caleridas: as above

phate (approved):
LGTM

PR #1232, opened 2025-09-14
Overhaul loop unswitching

phate:


haved (approved):
Looks great, the only thing I am missing is a docstring on the LoopUnswitching class that explains what the pass actually does. That does not need to be a part of this PR, however

PR #1237, opened 2025-09-17
Rename callable_argtype to CallableArgType

phate:


haved:
@phate 
The T in ArgType is lowercase in the code, but uppercase in the PR title

PR #1242, opened 2025-09-17
Fix loop unswitching check

phate:
The previous implementation of the loop unswitching check relied on that some transformations ran before. This was not the case and therefore loop unswitching did not trigger any longer. This PR renders the check independent of these transformations, ensuring that loop unswitching should trigger again.


phate:
@sjalander It seems that after the UnusedStateRemoval fix, we uncovered another problem. At least looking at the log, it is other benchmarks that are failing now. Any chance you could uncover me the commands again?

PR #1245, opened 2025-09-20
boost::intrusive_list

caleridas:
FYI, what using boost::intrusive_list would look like (i.e. essentially the same).

Not for merge right now, but for illustration.


haved:
The interface looks fine. I think the bigger question is whether or not we want to add a dependency on boost.

It would let us remove our own intrusive list implementation, which is nice.

Boost is an extremely common dependency, so it would not really be a problem to add it. Personally I would be fine with adding it. @phate @caleridas

PR #1256, opened 2025-09-21
HLS: Convert ConstantDistribution pass into a transformation

phate:


jlm/hls/backend/rvsdg2rhls/distribute-constants.hpp
```diff
@@ -3,22 +3,42 @@
...
+
+  ConstantDistribution(const ConstantDistribution &) = delete;
+
+  ConstantDistribution(ConstantDistribution &&) = delete;
+
+  ConstantDistribution &
+  operator=(const ConstantDistribution &) = delete;
+
+  ConstantDistribution &
+  operator=(ConstantDistribution &&) = delete;
```
haved, line 29: Is it a bit overkill to delete move constructors and move assignment? This makes it impossible to put the transformations in a vector, for example

haved (approved):
Looks good, one general comment @phate

jlm/hls/backend/rvsdg2rhls/distribute-constants.hpp
phate: Fixed.

jlm/hls/backend/rvsdg2rhls/distribute-constants.cpp
```diff
@@ -8,10 +8,11 @@
 #include <jlm/hls/ir/hls.hpp>
 #include <jlm/hls/util/view.hpp>
 #include <jlm/rvsdg/gamma.hpp>
+#include <jlm/rvsdg/lambda.hpp>
```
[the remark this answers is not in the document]
phate: Because they were removed from the distribute-constants.hpp

jlm/hls/backend/rvsdg2rhls/distribute-constants.hpp
haved: @phate It looks like you removed the lines that delete the copy constructor/assignment, not the move constructor/assignment. Since this class is completely stateless I guess we don't really need to delete either, but I think the move constructor is the most important to not delete

phate: Meh, sleeping on the wheel. Should be fixed now.

haved (approved):

PR #1267, opened 2025-09-24
HLS: Add jhls transformation sequence

phate:


phate:
@haved This needs eventually some more refactoring, but it allows me for now to get debug graphs out.

phate:
Hmm interesting. Why do I get cycle changes. I should not have cycle changes from unless I screwed up the order/number of passes.

sjalander (approved):

PR #1274, opened 2025-09-26
Fix compilation of ScalarEvolutionTests.cpp

phate:


haved (approved):
Bumping the CI to Ubuntu 24 would also let us get llvm-18-dev without any funky PPA repo stuff as well, so that sounds good @phate

PR #1275, opened 2025-09-27
Add LoadChainSeparation transformation

phate:
Add a transformation that renders chains of load operations independent from each other in the graph. Currently, the transformation only supports the separation within a region and does not go beyond region boundaries. Going beyond region boundaries will be tackled in a future PR.

This transformation will replace the node normalizations that try to perform the same RVSDG transformation as part of the node reduction transformation, albeit way more inefficiently.


haved (approved):
Looks great, 

I tried running NodeReduction after my state encoding, at it was very slow, so this should hopefully be great for speeding it up @phate

PR #1280, opened 2025-09-29
Bump hls-test-suite

sjalander:


haved (approved):

sjalander:
@haved @phate
I guess I should have waited until the checks are completed before asking for a review :)
The CI will not fail if, for some reason, it doesn't produce the right number of cycles.

PR #1282, opened 2025-09-30
Fix Inputs() and Outputs() method of node class

phate:


haved (approved):
Woah, wild that this had not caused problems yet

PR #1287, opened 2025-09-30
Add computeDepth() method to Region class

phate:


phate:
@caleridas 

>That's not very surprising, and implementation looks reasonable. I am just not sure it should be a "member function" of Region: it is not something "intrinsic" to the structure or operation of a region -- rather, it is a kind of analysis pass.

Moved outside of the region class.

jlm/rvsdg/region.cpp
```diff
@@ -515,6 +515,39 @@ Region::ToString(const util::Annotation & annotation, char labelValueSeparator)
   return util::strfmt(annotation.Label(), labelValueSeparator, value);
 }
 
+static size_t
+computeDepth(const Node & node, std::unordered_map<const Node *, size_t> & depthMap)
+{
+  if (depthMap.find(&node) != depthMap.end())
```
caleridas, line 521: more canonical would be to use:

auto i = depthMap.find(&node);
if (i != depthMap.end()) { return i->second; }

to avoid double search

jlm/rvsdg/region.cpp
```diff
@@ -515,6 +515,39 @@ Region::ToString(const util::Annotation & annotation, char labelValueSeparator)
...
+  {
+    return depthMap[&node];
+  }
+
+  size_t depth = 0;
+  for (auto & input : node.Inputs())
```
caleridas, line 527: I was about to say to use std::max_element on iterator range instead -- but it doesn't apply here as this doesn't deal with cached values

jlm/rvsdg/region.cpp
```diff
@@ -515,6 +515,39 @@ Region::ToString(const util::Annotation & annotation, char labelValueSeparator)
   return util::strfmt(annotation.Label(), labelValueSeparator, value);
 }
 
+static size_t
+computeDepth(const Node & node, std::unordered_map<const Node *, size_t> & depthMap)
```
caleridas, line 519: is it really that you want to recursively compute depth? why not TopDownTraverser?

jlm/rvsdg/region.cpp
```diff
@@ -515,6 +515,39 @@ Region::ToString(const util::Annotation & annotation, char labelValueSeparator)
...
+  depthMap[&node] = depth;
+
+  return depth;
+}
+
+std::unordered_map<const Node *, size_t>
+computeDepthMap(const Region & region)
+{
+  std::unordered_map<const Node *, size_t> depthMap;
+  for (auto & node : region.Nodes())
```
caleridas, line 543: see above -- using top-down traverser makes this much more straight

caleridas (approved):
Looks good in principle, but maybe not use recursion but TopDownTraverser?

jlm/rvsdg/region.cpp
phate: Fixed

phate:
@caleridas Changed to top-down traverser with const_cast

jlm/rvsdg/region.cpp
phate: Used top-down traverser

jlm/rvsdg/region.cpp
phate: Used top-down traverser

caleridas (approved):

PR #1294, opened 2025-10-02
TODO

phate:


caleridas:
> @caleridas The const-ness pushes through until the TraversalTracker, leading ultimately to a lot of code duplication. Not sure how to better model this. My two ideas on this:
> 
> 1. Templated traversal tracker base class, with two final subclasses, one where template is `Node*` and another where template is `const Node*`
> 2. TraversalConstIterator: Not much that can really done about it. We would need two iterators.
> 3. I also do not see much of how this could be factored out.
> 
> I am kind of not really happy about this all, but I do not have great ideas about it now. Do you have any?

Just template both traverser and tracker (e.g. as "GenericTopDownTraverser"). Then:

using TopDownTraverser = GenericTopDownTraverser<Node*>;
using ConstTopDownTracerserv = GenericTopDownTraverser<const Node*>;

No subclasses.

haved:
@phate @caleridas I had a go at writing a templated version that does not use region/input change notifications. See #1295.

I think this is the last place in the codebase where `notifiers.hpp` are used, unless someone who uses rvsdg as a library need them.

PR #1295, opened 2025-10-02
Create alternative TopDown and BottomUp Traverser

haved:
Looking at #1294 gave me some ideas, so I had a go at writing the `TopDownTraverser` and `BottomUpTraverser` in a way that takes a `template <bool IsConst>` to create both a const and non-const version.

I went for a slightly different implementation to remove the need for notifications from region changes, and made the rules regarding modifying regions during traversal a bit more strictly defined:
 - It is illegal to delete nodes that have not yet been visited. Deleting the current node is okay.
 - Any node added after traversal begins will NOT be visited, no matter where you add it.

With the old implementation is was technically possible to visit a node before its predecessor had been visited, if an input was diverted after the node was added to the frontier. This implementation adds a check, but the check only runs in the non-const version. The const version does not bother checking.

Because I do not use notifications from the region, it is possible for the frontier to become empty, despite nodes still being left. This will trigger a "re-frontiering", which is a O(n) operation, but it only happens if the region is changed in a very weird way during traversal.

### DIFF
--- a/jlm/rvsdg/traverser.cpp
+++ b/jlm/rvsdg/traverser.cpp
@@ -5,82 +5,113 @@
  */
 
 #include <jlm/rvsdg/graph.hpp>
-#include <jlm/rvsdg/notifiers.hpp>
 #include <jlm/rvsdg/traverser.hpp>
-
-using namespace std::placeholders;
-
-/* top down traverser */
 
 namespace jlm::rvsdg
 {
 
-TopDownTraverser::~TopDownTraverser() noexcept = default;
-
-TopDownTraverser::TopDownTraverser(Region * region)
-    : region_(*region)
+namespace detail
 {
-  for (auto & node : region->TopNodes())
-    tracker_.set_nodestate(&node, traversal_nodestate::frontier);
+template<bool IsConst>
+TopDownTraverserGeneric<IsConst>::~TopDownTraverserGeneric() noexcept = default;
 
-  for (size_t n = 0; n < region->narguments(); n++)
+template<bool IsConst>
+TopDownTraverserGeneric<IsConst>::TopDownTraverserGeneric(RegionType * region)
+{
+  for (auto & node : region->Nodes())
   {
-    auto argument = region->argument(n);
-    for (const auto & user : argument->Users())
-    {
-      if (auto node = TryGetOwnerNode<Node>(user))
-      {
-        if (!predecessors_visited(node))
-          continue;
+    unvisited_.Insert(&node);
 
-        tracker_.set_nodestate(node, traversal_nodestate::frontier);
+    // Build the initial frontier from nodes that have no predecessor nodes
+    bool onFrontier = true;
+    for (auto & input : node.Inputs())
+    {
+      if (TryGetOwnerNode<Node>(*input.origin()))
+      {
+        onFrontier = false;
+        break;
       }
     }
+    if (onFrontier)
+      frontier_.emplace(node.GetNodeId(), &node);
   }
-
-  callbacks_.push_back(on_node_create.connect(std::bind(&TopDownTraverser::node_create, this, _1)));
-  callbacks_.push_back(
-      on_input_change.connect(std::bind(&TopDownTraverser::input_change, this, _1, _2, _3)));
 }
 
+template<bool IsConst>
 bool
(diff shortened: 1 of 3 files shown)


jlm/rvsdg/traverser.hpp
```diff
@@ -24,203 +23,206 @@ class Region;
...
+  /**
+   * Extracts a node from the frontier.
+   * In non-const traversals, it double checks that the node is actually ready to be visited.
+   * @returns the popped node, or nullptr if all nodes have been visited.
+   */
+  [[nodiscard]] NodeType *
+  popFrontier();
+
+  // The set of nodes that are yet to be visited
+  util::HashSet<NodeType *> unvisited_;
```
phate, line 127: Keeping track of the nodes that have not been visited yet is expensive, as it effectively means you have to go through all the nodes in the region twice. Once, to fill this set, and then once the actual traversal. Would it not be possible to inverse it and keep track of the visited nodes, and all other nodes are by default in the unvisited state?

jlm/rvsdg/traverser.cpp
```diff
@@ -5,293 +5,235 @@
...
 
-  tracker_.set_nodestate(node, traversal_nodestate::behind);
-  for (size_t n = 0; n < node->noutputs(); n++)
+  if (unvisited_.IsEmpty())
+    return false;
+
+  // We still have unvisited nodes, despite the frontier being empty
+  // This can happen in rare cases when nodes ahead of the frontier change their inputs.
+  // Go over the unvisited nodes again to create a new frontier
+  for (auto & node : unvisited_.Items())
```
phate, line 67: Well no, this means you can not just keep track of the visited nodes, and need the unvisited nodes.

phate (commented):
@haved I just leave some comments. I will leave the final review to @caleridas 

In case he is okay(-ish) with it, I would like to see some performance measurements compared to the old implementation. I am particularly worried about the re-frontiering part and keeping track of all unvisited nodes. It effectively means that you visit all nodes in a region twice: once for the initialization of the traverser, once for traversal.

The initial constraints as specified in your commit description sound reasonable. We generally do not remove any node, except the one under visitation, before traversal ends anyway. Nodes not visiting that were added throughout traversal is also reasonable. I do not see those as a problem.

Since you already did this, I will hold back on a reimplementation as discussed in #1294 to avoid duplicated work. I am eager to see what @caleridas has to say.

phate:
>Callbacks seem like the best for performance, but maybe they could be added per region instead of having a global variable containing all callbacks? @caleridas

Callbacks should definitely(!) be done per region instead of globally. I was aware of this problem since quite a while, and I suspect that it causes quite a performance hit. However, it was not really relevant so far, so I just left it untouched.

> PS: I think there is a bug in the current traverser that can cause nodes to be visited before their predecessors, which should be fixed.

Could you elaborate on this? Do you have an example where this could happen? We need something more tangible here.

@haved Do I conclude from your last post that your solution is not really competitive in terms of performance, and I should just go ahead over the weekend and generalize the current traversers as laid out in #1294 ?

caleridas:
Callbacks per region: that's almost a no-brainer. While I'd normally suggest boost::signal, that's a bit menory-heavy in my taste and maybe sticking to DIY is better (I think I can cut overhead down to 8 bytes per region per signal).

I don't think global callback is a performance issue, but it is definitely a "cleanliness" issue.

phate:
> I don't think global callback is a performance issue, but it is definitely a "cleanliness" issue.

My gut feeling tells me it is a performance issue. When the region tree is recursively traversed and the innermost regions are handled first, it means we are creating a traverser for each region from the root down to the leaf node. Creating a node and/or changing an input in the leaf means then that the on_node_create() or on_input_change() method is invoked for EVERY traverser that was created even though only the innermost region is responsible for it. My suspicion is that this has a measurable performance impact, but I might also easily misjudge it.

haved:
Yes, @phate, but feel free to use what I wrote here for generics-inspiration :)

haved:
Also, @phate, while I remember it

If inputs have changed such that a node on the frontier no longer has only visited predecessors, we should remove it from the frontier. That is not currently happening.

phate:
@haved This PR can be closed now, no?

haved:
@phate absolutely

PR #1299, opened 2025-10-04
Rework traversers and notifiers

caleridas:
Introduce RegionObserver, use it in traversers

This introduces the concept of a "RegionObserver" which -- when alive -- can observe property changes of a region (node addition / removal and edge changes). Use this instead of global notifier chains for traversers.

Since notifications are now per-region, everything is managed through an object attached to a region. Using a singular RegionObserver object instead of individual callbacks allows to limit the growth of the "Region" object to a single pointer.

This has the following knock-on effects:
- traversers become smaller, and need to manage a bit less (simplified observer setup, no "region" object anymore)
- since notification is per region, this avoids quadratic complexity for nested regions/traversers
- can delete entirety of callback / notifier related code

While at it, the "revisit" feature of BottomUpTraverser was not used anymore, and was in fact broken. So removing this as well.


jlm/rvsdg/region.hpp
```diff
@@ -763,6 +773,34 @@ class Region
...
+  friend class SimpleNode;
+  friend class StructuralNode;
+  friend class Input;
+};
+
+class RegionObserver
+{
+public:
+  ~RegionObserver();
+  RegionObserver(Region * region);
```
haved, line 788: Since we use pointers to these observers and their fields, I think it makes sense to give them deleted copy constructors and move constructors, or alternatively implement a move constructor that updates the pointers

haved (commented):
Looks nice, @caleridas , I have one comment regarding the linked lists

jlm/rvsdg/traverser.cpp
```diff
@@ -5,7 +5,6 @@
  */
 
 #include <jlm/rvsdg/graph.hpp>
-#include <jlm/rvsdg/notifiers.hpp>
 #include <jlm/rvsdg/traverser.hpp>
 
 using namespace std::placeholders;
```
haved, line 10: This one is not needed any more, no?

jlm/rvsdg/region.hpp
haved: The region itself can update the head observer's pprev pointer in its move constructor as well

caleridas: I rather made it non-copyable (and non-movable) -- this should really semantically be a "pinned" class.

haved (dismissed):
Looks good, @caleridas !

I see some issues with the traversers in how they handle changed inputs and such, but that can be a different PR, as those problems have been there for a long time already

jlm/rvsdg/region.hpp
```diff
@@ -763,6 +778,40 @@ class Region
...
+  operator=(const RegionObserver &) = delete;
+
+  virtual void
+  NodeCreate(Node * node);
+
+  virtual void
+  NodeDestroy(Node * node);
+
+  virtual void
+  InputChange(Input * input, Output * old_origin, Output * new_origin);
```
phate, line 808: Any particular reason why they are not abstract methods?

jlm/rvsdg/traverser.cpp
```diff
@@ -5,20 +5,38 @@
...
 #include <jlm/rvsdg/traverser.hpp>
 
-using namespace std::placeholders;
-
 /* top down traverser */
 
 namespace jlm::rvsdg
 {
 
+TopDownTraverser::Observer::Observer(Region * region, TopDownTraverser * traverser)
```
phate, line 15: References

jlm/rvsdg/traverser.cpp
```diff
@@ -138,11 +146,33 @@ HasSuccessors(const Node & node)
   return false;
 }
 
+BottomUpTraverser::Observer::Observer(Region * region, BottomUpTraverser * traverser)
```
phate, line 149: References

phate (dismissed):
Mostly style from my side. I like the idea.

jlm/rvsdg/region.hpp
caleridas: Defaulting them as "do nothing" adds a tiny bit of (dubious) convenience. Changing this to a clean interface.

jlm/rvsdg/region.hpp
```diff
@@ -763,6 +778,40 @@ class Region
...
+
+private:
+  RegionObserver ** pprev_;
```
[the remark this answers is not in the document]
caleridas: The extra "p" is an attempt to encode that it doesn't point to the previous object, but to a pointer inside the previous object (which may also be the list root. No strong opinion on name, didn't really think much further than this knee-jerk name.

jlm/rvsdg/region.hpp
```diff
@@ -763,6 +778,40 @@ class Region
...
   region_nodes_list Nodes_;
+  RegionObserver * observers_ = nullptr;
+
+  friend class RegionObserver;
+  friend class SimpleNode;
+  friend class StructuralNode;
+  friend class Input;
+};
+
+class RegionObserver
```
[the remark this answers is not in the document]
caleridas: done

jlm/rvsdg/region.hpp
caleridas: made it pure interface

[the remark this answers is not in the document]
caleridas: change names

jlm/rvsdg/region.hpp
```diff
@@ -754,6 +760,15 @@ class Region
...
 
+  void
+  notifyNodeCreate(Node * node);
+
+  void
+  notifyNodeDestroy(Node * node);
+
+  void
+  notifyInputChange(Input * input, Output * old_origin, Output * new_origin);
+
```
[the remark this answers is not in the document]
caleridas: yes I am aware that I simply preserved the original interface of the callbacks... I feel like I am spending about 25% time on each PR wrangling with the impedance mismatch on passing references and pointers between structs and call arguments, e.g.:

```
traversal_nodestate
TraversalTracker::get_nodestate(Node * node)
{
  auto i = states_.find(node);
```

I turn call arg into a reference, turn it back into pointer for the struct, then when taking it out again turn it back into reference.

Probably for Input/Output it would by far be best to simply have "handle" types that encapsulate a pointer and pass these around with value semantics to sidestep this whole thing (which would also allow more easily changing underlying repr).

PR #1304, opened 2025-10-07
comment out memoryStateDecoupling

phate:


phate:
@sjalander @haved Is the memory state decoupling pass a necessity to get anything through HLS? I thought it is an optimization, but not a necessity? Any other pass that follows or precedes it has a requirement that this pass comes before or after?

PR #1307, opened 2025-10-07
Add size to intrusive list, and move region's top/bottom/node callbacks to private

### LINKED ISSUE #1305: RemoveBottomNode and RemoveTopNode are O(n)
They check the size of the list, both before and after removing, but they are linked lists, so this is O(n).

When a delta node is an array of, e.g., 190000 constants, this means each one of those nodes gets first added, and later removed, from the bottom list, which is super slow.

The count is only ever used to assert that the list we removed from is the correct one, so it should not really run in release builds.
It is so slow that I also think we might want to remove it from debug builds.

haved:
Uses `friend class` to keep the region invariant maintaining callbacks private.

Adds a `size_t size_` to the intrusive linked list.

Sadly the splice operation on the linked list can no longer be O(1) when we need to maintain sizes. @caleridas

Makes the anchor non-copyable and non-moveable.

Fixes #1305

### DIFF
--- a/jlm/hls/backend/rvsdg2rhls/rvsdg2rhls.cpp
+++ b/jlm/hls/backend/rvsdg2rhls/rvsdg2rhls.cpp
@@ -468,6 +468,9 @@ createTransformationSequence(rvsdg::DotWriter & dotWriter, const bool dumpRvsdgD
   auto bufferInsertion = std::make_shared<BufferInsertion>();
   auto rhlsVerification = std::make_shared<RhlsVerification>();
 
+  // Use this transformation to dump HLS dot graphs at specific points in the sequence
+  [[maybe_unused]] auto dumpDot = std::make_shared<DumpDotTransformation>();
+
   std::vector<std::shared_ptr<rvsdg::Transformation>> sequence({
       loopUnswitching,
       deadNodeElimination,
--- a/jlm/hls/util/view.cpp
+++ b/jlm/hls/util/view.cpp
@@ -501,4 +501,19 @@ DotToSvg(const std::string & file_name)
     exit(EXIT_FAILURE);
 }
 
+DumpDotTransformation::~DumpDotTransformation() noexcept = default;
+
+DumpDotTransformation::DumpDotTransformation()
+    : Transformation("DumpDotTransformation")
+{}
+
+void
+DumpDotTransformation::Run(
+    rvsdg::RvsdgModule & rvsdgModule,
+    util::StatisticsCollector & statisticsCollector)
+{
+  const auto file = statisticsCollector.CreateOutputFile("rvsdg-graph.dot", true);
+  DumpDot(&rvsdgModule.Rvsdg().GetRootRegion(), file.path().to_str());
+}
+
 } // namespace jlm::hls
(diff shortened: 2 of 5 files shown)


jlm/rvsdg/node.hpp
```diff
@@ -277,13 +277,13 @@ class Output
...
   {
-    return NumUsers_;
+    return Users_.size();
   }
 
   /**
    * Determines whether the output is dead.
    *
-   * An output is considered dead when it has no users.
+   * An output is considered dead when it has no users.x
```
phate, line 286: superfluous x

jlm/util/intrusive-list.hpp
```diff
@@ -443,6 +452,9 @@ class IntrusiveList
       return;
     }
 
+    // Maintaining the sizes of the linked lists requires range splicing to be O(n)
+    size_t count = std::distance(begin, end);
```
phate, line 456: Stupid question: Is slicing anywhere actually used? In other words, do we pay a performance penalty here?

jlm/rvsdg/node.hpp
```diff
@@ -292,7 +292,7 @@ class Output
   [[nodiscard]] bool
   IsDead() const noexcept
   {
-    return NumUsers_ == 0;
+    return Users_.size() == 0;
```
phate, line 295: As the intrusive list keeps now track of the size of the list, it should be possible to remove the NumUsers_ attribute from the Output class.

phate (approved):
I have nothing spectacular, but I would also like @caleridas to approve before you merge.

caleridas:
> Uses `friend class` to keep the region invariant maintaining callbacks private.
> 
> Adds a `size_t size_` to the intrusive linked list.
> 
> Sadly the splice operation on the linked list can no longer be O(1) when we need to maintain sizes. @caleridas
> 
> Makes the anchor non-copyable and non-moveable.
> 
> Fixes #1305

Preferentially don't add as member to intrusive list, but to the upper structures that want O(1) size (i.e. "Region"). That's canonical as far as list structures go (see also boost intrusive).

That also answers the questions regarding splice.

jlm/rvsdg/node.hpp
caleridas: As per other comment, I would suggest not counting in the list ADT but outside

jlm/rvsdg/region.hpp
```diff
@@ -509,133 +512,146 @@ class Region
...
+    subregions into the target region. Substitutions
+    will be performed as specified, and the substitution
+    map will be updated as nodes are copied.
+  */
+  void
+  copy(Region * target, SubstitutionMap & smap, bool copy_arguments, bool copy_results) const;
+
+  /**
+   * Removes all dead nodes from the region, including nodes that become dead during pruning.
+   * @param recursive it true, any subregions are also pruned, as well as their subregions etc.
```
phate, line 567: iF true, any...

phate (dismissed):

jlm/rvsdg/region.hpp
haved: fixed

PR #1334, opened 2025-10-17
Refactor constant distribution pass

phate:


tests/jlm/hls/backend/rvsdg2rhls/DistributeConstantsTests.cpp
```diff
@@ -69,7 +68,6 @@ GammaSubregionUsage()
 
   // Assert
   assert(lambdaNode->subregion()->numNodes() == 2);
-  assert(constantNode.output(0)->IsDead());
```
phate: The old implementation also did a poor man's DNE. The new implementation does not try this, but rather relies on DNE to clean up afterwards.

haved:
@phate seems like the hls test suite is angry again

phate:
@haved The reworked pass does its job now. However, it seems that we also distribute the constants slightly different compared to before (a few more benchmarks changed number of cycles). I am wondering whether it is worth investigating or whether we can just live with it.

sjalander:
@phate @haved 
Since it's the decoupled benchmarks that are impacted, I'd like to test with a longer memory latency to see how they get affected.

sjalander:
@phate 
Maybe usually only provide doxygen on the public API but isn't there a benefit for internal use to also have it for the private methods.

phate:
@haved @sjalander Could I get a review, please?

PR #1335, opened 2025-10-17
Make TopDown and BottomUp traversers generic for constness

haved:
This is done by creating two generic Observers. One of them is an actual observer that forwards all events to the traverser, while the other is a dummy class that does nothing except take a `const Region &`.

It still uses the same `TraversalTracker` as before, just makes it generic over the constness of the `Node` type.

It would also be possible to not make the `TraversalTracker` generic, and instead do `const_cast` every time we need a non-const Node.

Lastly it adds a check that only visited nodes are removed in the bottom up traverser. Otherwise it can cause bugs.

The new const traversers are called `TopDownConstTraverser` and `BottomUpConstTraverser`.

FYI @caleridas

### DIFF
--- a/jlm/rvsdg/traverser.cpp
+++ b/jlm/rvsdg/traverser.cpp
@@ -31,7 +31,7 @@ template<typename Traverser>
 ForwardingObserver<Traverser>::~ForwardingObserver() noexcept = default;
 
 template<typename Traverser>
-ForwardingObserver<Traverser>::ForwardingObserver(Region & region, Traverser & traverser)
+ForwardingObserver<Traverser>::ForwardingObserver(const Region & region, Traverser & traverser)
     : RegionObserver(region),
       traverser_(traverser)
 {}
@@ -71,18 +71,13 @@ ForwardingObserver<Traverser>::onInputDestroy(Input * input)
   traverser_.onInputDestroy(input);
 }
 
-template<typename Traverser>
-DummyObserver<Traverser>::DummyObserver(
-    [[maybe_unused]] const Region & region,
-    [[maybe_unused]] Traverser & traverser)
-{}
-
 template<bool IsConst>
 TopDownTraverserGeneric<IsConst>::~TopDownTraverserGeneric() noexcept = default;
 
 template<bool IsConst>
 TopDownTraverserGeneric<IsConst>::TopDownTraverserGeneric(RegionType * region)
-    : observer_(*region, *this)
+    : observer_(*region, *this),
+      nodeIdCutoff_(region->getNextNodeId())
 {
   for (auto & node : region->TopNodes())
   {
@@ -103,7 +98,7 @@ TopDownTraverserGeneric<IsConst>::TopDownTraverserGeneric(RegionType * region)
 
 template<bool IsConst>
 bool
-TopDownTraverserGeneric<IsConst>::isOutputActivated(const Output & output)
+TopDownTraverserGeneric<IsConst>::isOutputActivated(const Output & output) const
 {
   if (auto pred = TryGetOwnerNode<Node>(output))
   {
@@ -135,29 +130,40 @@ template<bool IsConst>
 typename TopDownTraverserGeneric<IsConst>::NodeType *
 TopDownTraverserGeneric<IsConst>::next()
 {
-  const auto node = tracker_.peek();
-  if (!node)
-    return nullptr;
+  while (true)
+  {
+    const auto node = tracker_.peek();
+    if (!node)
+      return nullptr;
 
-  markAsVisited(*node);
+    markAsVisited(*node);
 
-  return node;
+    // Skip nodes that were created after traversal started
+    if (node->GetNodeId() >= nodeIdCutoff_)
+      continue;
+
+    return node;
+  }
 }
 
 template<bool IsConst>
 void
 TopDownTraverserGeneric<IsConst>::onNodeCreate(NodeType * node)
 {
-  for (const auto & input : node->Inputs())
+  if (node->ninputs() == 0)
   {
-    if (isOutputActivated(*input.origin()))
+    tracker_.checkNodeActivation(node, node->ninputs());
+  }
+  else
+  {
+    for (const auto & input : node->Inputs())
     {
-      tracker_.incActivationCount(node, node->ninputs());
+      if (isOutputActivated(*input.origin()))
+      {
+        tracker_.incActivationCount(node, node->ninputs());
+      }
     }
   }
-
-  if (node->ninputs() == 0)
-    tracker_.checkNodeActivation(node, node->ninputs());
 
   // If node would end up on frontier (because all predecessors
   // have been visited), mark it as visited instead (we do not
@@ -232,10 +238,13 @@ TopDownTraverserGeneric<IsConst>::onInputDestroy(Input * input)
 
   if (isOutputActivated(*input->origin()))
   {
-    tracker_.decActivationCount(node, node->ninputs() - 1);
+    // Removing an activated input can never cause de-activation of the node,
+    // so we can use 0 as the deactivation threshold
+    tracker_.decActivationCount(node, 0);
   }
   else
   {
+    // The callback is called before the input is actually removed, so use the new threshold
     tracker_.checkNodeActivation(node, node->ninputs() - 1);
   }
 }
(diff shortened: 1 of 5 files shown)


jlm/rvsdg/traverser.hpp
```diff
@@ -24,208 +24,201 @@ class Region;
...
 
-template<typename T>
+enum class TraversalNodestate
+{
+  ahead = -1,
+  frontier = 0,
+  behind = +1
+};
+
+/* support class to track traversal states of nodes */
```
phate: A doxygen comment would be nice.

jlm/rvsdg/traverser.hpp
```diff
@@ -24,208 +24,201 @@ class Region;
...
-  typedef value_type * pointer;
-  typedef value_type & reference;
+  using iterator_category = std::input_iterator_tag;
+  using value_type = NodeType *;
+  using difference_type = ssize_t;
+  using pointer = value_type *;
+  using reference = value_type &;
 
-  constexpr explicit TraverserIterator(T * traverser = nullptr, Node * node = nullptr) noexcept
+  constexpr TraverserIterator(Traverser * traverser, NodeType * node) noexcept
```
phate: Both parameters could be references, no?

jlm/rvsdg/traverser.hpp
```diff
@@ -236,87 +229,104 @@ class TopDownTraverser final
...
+ *
+ * The top down traverser visits all nodes in a region, starting at the nodes that have no inputs
+ * besides graph arguments, i.e. from the topmost nodes to the nodes at the bottom.
+ * Nodes created during traversal are not visited.
+ *
+ * The main usage of the top down traverser is for analyzing graphs, or replacing subgraphs in the
+ * already visited part of a region.
+ *
+ * An alternative to traversing all nodes using next() is the utilization of begin() and end().
+ */
```
phate: I believe this documentation should be on the TopDownTraverserGeneric.

jlm/rvsdg/traverser.cpp
```diff
@@ -193,258 +222,275 @@ TopDownTraverser::onInputChange(Input * in, Output * old_output, Output * new_ou
...
-  auto node = TryGetOwnerNode<Node>(*input);
+  const auto node = TryGetOwnerNode<Node>(*input);
   if (!node)
     return;
 
-  if (isOutputActivated(input->origin()))
+  if (isOutputActivated(*input->origin()))
   {
-    tracker_.decActivationCount(node, 0);
+    tracker_.decActivationCount(node, node->ninputs() - 1);
```
phate: This line changed compare to the original code.

jlm/rvsdg/traverser.cpp
```diff
@@ -193,258 +222,275 @@ TopDownTraverser::onInputChange(Input * in, Output * old_output, Output * new_ou
...
   {
-    if (isInputActivated(input))
-    {
-      tracker_.decActivationCount(pred, 0);
-    }
-    else
-    {
-      tracker_.checkNodeActivation(pred, pred->numSuccessors() - 1);
-    }
+    tracker_.decActivationCount(node, node->numSuccessors() - 1);
```
phate: This line changed compared to the previous code.

jlm/rvsdg/traverser.hpp
```diff
@@ -236,87 +229,104 @@ class TopDownTraverser final
...
-    void
-    onInputDestroy(Input * input) override;
-
-  private:
-    BottomUpTraverser & traverser_;
-  };
-
   bool
-  isInputActivated(const Input * input) const;
+  isInputActivated(const Input & output);
```
caleridas: const Input & output ??? And why remove const qual from function?

caleridas:
> This is done by creating two generic Observers. One of them is an actual observer that forwards all events to the traverser, while the other is a dummy class that does nothing except take a `const Region &`.
> 
> It still uses the same `TraversalTracker` as before, just makes it generic over the constness of the `Node` type.
> 
> It would also be possible to not make the `TraversalTracker` generic, and instead do `const_cast` every time we need a non-const Node.
> 
> Lastly it adds a check that only visited nodes are removed in the bottom up traverser. Otherwise it can cause bugs.
> 
> The new const traversers are called `TopDownConstTraverser` and `BottomUpConstTraverser`.
> 
> FYI @caleridas

Thinking about this, I actually disagree about the dummy observer -- just make it a regular observer. The fact that the traverser itself is const does not mean that it is allowed to assume immutability of the graph through other means. It just means that the traverser itself cannot provide to a caller the means of mutation, hence only return const nodes.

If you want immutability of the underlying graph, you would need some way to express "exclusivity" of access to the graph by traverser which is quite the opposite of const (namely exclusive ownership).

haved:
@caleridas I also thought about that. We could provide "interior mutability" to allow observers to be created on `const Region &`.

I originally decided against it, as I found it very unlikely that we ever actually make modifications to a graph while it is being traversed, unless the changes are made by the algorithm using the traverser itself. In general I also find it very unlikely that we ever actually have more than one traverser on a given region at a time, but allowing multiple concurrent observers using a linked list is so cheap that there is no reason not to.

I would be open to adding const traversers as proper observers. It seems highly unlikely that any changes will be made to the region, so there should be almost no overhead due to notifying the observer.

As a final note, I think the restrictions we are implementing / proposing on the region under traversal (prohibit removing nodes that have not been visited in bottom up, prohibit directing inputs to nodes that have not been visited in top down) make it even less likely that we ever have multiple traversers on the same region.

We could even give the const traverser a region observer that throws an error if the region ever changes.

What do you think @phate?

caleridas:
> @caleridas I also thought about that. We could provide "interior mutability" to allow observers to be created on `const Region &`.
> 
> I originally decided against it, as I found it very unlikely that we ever actually make modifications to a graph while it is being traversed, unless the changes are made by the algorithm using the traverser itself. In general I also find it very unlikely that we ever actually have more than one traverser on a given region at a time, but allowing multiple concurrent observers using a linked list is so cheap that there is no reason not to.
> 
> I would be open to adding const traversers as proper observers. It seems highly unlikely that any changes will be made to the region, so there should be almost no overhead due to notifying the observer.
> 
> As a final note, I think the restrictions we are implementing / proposing on the region under traversal (prohibit removing nodes that have not been visited in bottom up, prohibit directing inputs to nodes that have not been visited in top down) make it even less likely that we ever have multiple traversers on the same region.
> 
> We could even give the const traverser a region observer that throws an error if the region ever changes.
> 
> What do you think @phate?

I think marking the observer support accessible to const Region is very sensible (it amounts to marking the list head as "mutable" which is reasonable imho). Pedantically, the callbacks then provide access to non-const Node* which could be tightened, but I think it is entirely fine to leave that open and just ensure that traversers themselves are const-correct.

jlm/rvsdg/traverser.hpp
haved: done

jlm/rvsdg/traverser.hpp
```diff
@@ -236,87 +229,104 @@ class TopDownTraverser final
...
+using TopDownConstTraverser = detail::TopDownTraverserGeneric<true>;
+
+/** \brief BottomUp Traverser
+ *
+ * The bottom up traverser visits all nodes in a region, starting at the nodes that have no users
+ * besides graph results, i.e. from the bottommost nodes to the nodes at the top.
+ * Nodes created during traversal are never visited.
+ *
+ * An alternative to traversing all nodes using next() is the utilization of begin() and end().
+ */
```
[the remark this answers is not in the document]
haved: done

jlm/rvsdg/traverser.cpp
haved: This change makes no difference, so I changed this back to 0 and added a comment to explain.

The bugfix I was thinking about was in `onInputChange`, where it would do nothing if the input was not activated. However, moving a not activated input might remove the last non-activated user of a node, which should place the node on the frontier. Likewise it may cause the new origin node to no longer be on the frontier

haved:
I did what @caleridas suggested and made it possible to create observers on `const Region`s. It does break const correctness internally, which could be tightened later, along with all the other places where a `const Input` gives a non-const `Output *` etc.

Since we for now can not enforce that inputs are not re-routed to unvisited parts in the TopDown (without breaking several passes), I instead added the band-aid solution of checking the Node:Id of nodes before returning them from `next()`.

jlm/rvsdg/traverser.cpp
haved: Yes  good point, it should have been separate

jlm/rvsdg/traverser.hpp
```diff
@@ -236,87 +229,102 @@ class TopDownTraverser final
...
+  ObserverType observer_;
+  // Any node with Id >= this cutoff was created after traversal began and will be skipped
+  Node::Id nodeIdCutoff_;
 };
 
-class BottomUpTraverser final
+/** \brief BottomUp Traverser
+ *
+ * The bottom up traverser visits all nodes in a region, starting at the nodes that have no users
+ * besides graph results, i.e. from the bottommost nodes to the nodes at the top.
```
phate: typo: bottommost

PR #1341, opened 2025-10-21
Implement scalar evolution logic

andreaslhjulstad:
This PR extends the pre-existing ScalarEvolution (SCEV) analysis pass with a working implementation for computing SCEV expressions. The pass now creates an add-recurrence for each induction variable, which shows how the variable evolves in the loop. In the future, this (possibly recursive) add recurrence will be folded into a chain-recurrence to make it easier to apply it for optimizations.

Limitations:
As of now, we can only compute the SCEV expressions for nodes that depend on ADD nodes, with plans to extend this to SUB nodes later. Also, we can only compute expressions for loop variables that have a constant input-value. The next step is to look into how we can support non-constant loop-invariants (e.g. variables that are returned from a function).


haved:
@andreaslhjulstad can you merge your branch with master?
It should be a case of
```sh
# while on the scalar-evolution branch
git fetch origin master:master
git merge master
git push
```

You might get merge conflicts due to your previous commits being squashed into one commit when your last PR was merged.

PR #1345, opened 2025-10-22
Reimplement pullin_bottom() method

phate:
This reimplements the pullin_bottom() method, making it independent from the dynamically computed depth value of a node.


haved:
The code looks good, @phate, but we get linker errors. I imagine it is due to `computeDepthMap` being in `librvsdg.a`, but never used there, so the linker forgets it exists.

I thought we had fixed everything to do with linker ordering, though..

haved:
Ah, no, sorry. That was not it. We declare `computeDepthMap` both as a static method and as a standalone function. Only the latter actually has a definition. @phate

Issue #1333, opened 2025-10-16
Invariant Value Redirection could be made aware of predicates


haved:
In a simple program like the following:

```c
int popcount(int v) {
    int pcnt = 0;
    do {
        pcnt++;
        v = v & (v-1);
    } while(v > 0);
    return pcnt;
}
```

After compiling to LLVM IR, doing mem2reg, and converting it to RVSDG, we get a theta with a gamma at the bottom, like so:

<img width="1177" height="899" alt="Image" src="https://github.com/user-attachments/assets/287f2fe2-cf07-426c-bbdc-8361e19da71b" />

The reason we get this is that the top of the do-while gets phi instructions in the LLVM IR, which are then turned into variable assignments in the jlm frontend. Even though the original C code was a `do` `while` loop, which should map nicely onto a theta node, we still get this gamma node.

Observe, however, that this gamma node could be removed.

First of all, the predicate for the theta is produced by the gamma, where gamma region 0 gives the theta predicate 0, and gamma region 1 gives the theta predicate 1. Thus, we could just re-use the gamma predicate (or alternatively create our own match node for it, in case the regions and predicates need to be swapped)

Next, observe that loop variable 0 is dead after the theta. This means that in the final iteration of the theta, it does not matter what value we give to the post-result for this variable. We only care about its value when the theta is still looping. Since the post-result gets its value from the gamma, and we know that the gamma and the theta shares predicate, we only need to consider region 1 (graph5). In that region the result is just a copy of entry variable 1, which comes from the `IAdd`. If invariant value redirection was smart enough to discard region 0, it would see this, and route the output of the `IAdd` directly into the loop variable's post result.

Next consider loop variable 1, which is dead within the loop. It is only used as an output from the theta node. Since it originates from a gamma with a shared predicate, we know that only region 0 is relevant. In there, we see that the value is invariant, and comes from entry variable 3. IVR can again re-route around the gamma.

Because this theta was originally a nice do-while loop, every single loop variable is either invariant, dead within the theta, or dead after the theta. Smart IVR can thus make the entire gamma dead.

The same ideas will also apply if we have two gammas, where the second gamma takes entry variables from the exit variables of the first gamma, and the predicate is shared.

@caleridas @phate

PR #1350, opened 2025-10-25
Add theta-gamma redirection to InvariantValueRedirection pass

phate:
Extends the invariant value redirection pass with theta-gamma redirection. Theta-gamma redirection tries to determine whether the predicate of a theta node correlates with the predicate of a gamma node in the theta node's subregion, and then redirects the loop variables that fulfill certain conditions.

Addresses parts of #1333

### DIFF
--- a/tests/jlm/llvm/opt/InvariantValueRedirectionTests.cpp
+++ b/tests/jlm/llvm/opt/InvariantValueRedirectionTests.cpp
@@ -544,8 +544,6 @@ testThetaGammaRedirection()
 
   GraphExport::Create(*lambdaOutput, "test");
 
-  std::cout << view(&rvsdg.GetRootRegion()) << std::flush;
-
   // Act
   RunInvariantValueRedirection(*rvsdgModule);
 


phate:
@haved I fully see your point but I would like to refrain from adding it in this PR for the following reasons:
1. Redirecting the predicate is not really a transformation that should happen in the invariant value redirection, as it is not really an invariant value (minor concern, but I am afraid of getting a hodgepodge of transformation).
2. I have concerns with our back-end and whether it will be able to handle "shared predicates". I do not know whether these concerns are justified (I seem to remember that there was something), but I would rather not find out in this PR as it basically unblocks @andreaslhjulstad as it is. Once this PR is in, he can take happily take C code and get the respective loop variables that are not going through a gamma out.
3. This idea can be extended in various directions. For example, handling gamma-gamma correlating predicates or even copying nodes out of the gamma regions (which does not belong in IVR), gamma node that has another gamma node in one its subregions that also determines the predicate etc.

I think that there is quite a bit more here, but I would like to start to crawl first before we run. Thus, it it is acceptable for you, then I would like to get this PR in first and rather add other PRs later on top.

PR #1352, opened 2025-10-27
New common node elimination

haved:
Replaces Common Node Elimination pass with a slightly different approach.

Each congruence set has a "leader" output. Once an output has become a leader it will never become a follower. When marking simple nodes, we just as before look at the leftmost input and compare ourselves to all other simple nodes whose leftmost input is congruent with the origin. If no congruent leader is found, the node becomes its own leader. (more precisely: all its outputs are their own leader). A leader will always come earlier in the top down traversal order than all its followers.

When marking thetas we speculatively use the origins of loop variable inputs to partition loop variables, and mark the theta subregion. Once the subregion has been marked, it might turn out that loop variable results were not supposed to be congruent after all. In which case the loop variables are further partitioned, and the subregion is marked again.

A fun trick here is that the second time a node is visited, it already knows that it found a leader last time, so it quickly checks if it is still congruent with its old leader. If it is, then it does not need to look at any other nodes. Only if some partitioning has happened since last time to cause the node to no longer be congruent with its old leader, will it start looking for other nodes that it may be congruent with.

Runtime comparison in shown in the below plot:

image

In both cases we do all tricks in RegionAwareModRef, and then time the CNE afterwards.
The axis show billions on nanoseconds, i.e., seconds. The slowest file thus goes from ~400 seconds to 12 seconds.

I do want to look into what makes xdisp.c take 12 seconds, btw..

I also want to see what happens to the graphs that took 12 000 seconds with the old CNE.

### DIFF
--- a/jlm/llvm/backend/IpGraphToLlvmConverter.cpp
+++ b/jlm/llvm/backend/IpGraphToLlvmConverter.cpp
@@ -194,7 +194,7 @@ IpGraphToLlvmConverter::convert_ctlconstant(
     const std::vector<const Variable *> &,
     ::llvm::IRBuilder<> & builder)
 {
-  JLM_ASSERT(is_ctlconstant_op(op));
+  JLM_ASSERT(is<rvsdg::ControlConstantOperation>(op));
   auto & cop = *static_cast<const rvsdg::ControlConstantOperation *>(&op);
 
   size_t nbits = cop.value().nalternatives() == 2 ? 1 : 32;
--- a/jlm/rvsdg/control.hpp
+++ b/jlm/rvsdg/control.hpp
@@ -62,12 +62,6 @@ private:
   size_t nalternatives_;
 };
 
-static inline bool
-is_ctltype(const jlm::rvsdg::Type & type) noexcept
-{
-  return dynamic_cast<const ControlType *>(&type) != nullptr;
-}
-
 class ControlValueRepresentation
 {
 public:
@@ -128,19 +122,6 @@ typedef DomainConstOperation<
     ControlValueRepresentationFormatValue,
     ControlValueRepresentationTypeOfValue>
     ControlConstantOperation;
-
-static inline bool
-is_ctlconstant_op(const Operation & op) noexcept
-{
-  return dynamic_cast<const ControlConstantOperation *>(&op) != nullptr;
-}
-
-static inline const ControlConstantOperation &
-to_ctlconstant_op(const Operation & op) noexcept
-{
-  JLM_ASSERT(is_ctlconstant_op(op));
-  return *static_cast<const ControlConstantOperation *>(&op);
-}
 
 /**
  * Match operator
--- a/jlm/rvsdg/gamma.cpp
+++ b/jlm/rvsdg/gamma.cpp
@@ -20,7 +20,7 @@ static bool
 is_predicate_reducible(const GammaNode * gamma)
 {
   auto constant = rvsdg::TryGetOwnerNode<SimpleNode>(*gamma->predicate()->origin());
-  return constant && is_ctlconstant_op(constant->GetOperation());
+  return constant && is<ControlConstantOperation>(constant->GetOperation());
 }
 
 static void
@@ -94,7 +94,7 @@ is_control_constant_reducible(GammaNode * gamma)
   std::unordered_set<jlm::rvsdg::Output *> outputs;
   for (const auto & exitvar : gamma->GetExitVars())
   {
-    if (!is_ctltype(*exitvar.output->Type()))
+    if (!is<ControlType>(*exitvar.output->Type()))
       continue;
 
     size_t n = 0;
@@ -138,8 +138,9 @@ perform_control_constant_reduction(std::unordered_set<jlm::rvsdg::Output *> & ou
     for (size_t n = 0; n < xv.branchResult.size(); n++)
     {
       auto origin = xv.branchResult[n]->origin();
-      auto & value =
-          to_ctlconstant_op(AssertGetOwnerNode<SimpleNode>(*origin).GetOperation()).value();
+      auto [ctlConstantNode, ctlConstantOperation] =
+          TryGetSimpleNodeAndOptionalOp<ControlConstantOperation>(*origin);
+      auto & value = ctlConstantOperation->value();
       nalternatives = value.nalternatives();
       if (map.find(n) != map.end())
         new_mapping[map[n]] = value.alternative();


jlm/llvm/DotWriter.hpp
```diff
@@ -33,6 +33,15 @@ class LlvmDotWriter : public rvsdg::DotWriter
...
   AnnotateEdge(const rvsdg::Input & rvsdgInput, util::graph::Edge & edge) override;
+
+public:
+  /**
+   * Utility function for dumping the given region in dot format
+   * @param region the region to dump
+   * @param out the ostream to write the output to
+   */
+  static void
+  Dump(rvsdg::Region & region, std::ostream & out);
```
phate: There is no definition for this function in this PR.

jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -0,0 +1,1044 @@
...
+      {
+        markLambda(lambda, context);
+      },
+      [&](const rvsdg::PhiNode & phi)
+      {
+        markPhi(phi, context);
+      },
+      [&](const rvsdg::DeltaNode & delta)
+      {
+        markDelta(delta, context);
```
phate: The markDelta function can be inlined. It is literally a single line of code and comment.

phate (commented):
First round. I need more time to finish the review.

jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -0,0 +1,1017 @@
...
+public:
+  [[nodiscard]] CongruenceSets &
+  getCongruenceSets()
+  {
+    return congruenceSets_;
+  }
+
+private:
+  CongruenceSets congruenceSets_;
+};
```
phate: This indirection feels a bit useless with only a single member. Could the CongruenceSets class not just be renamed into context and we skip this indirection?

jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    {
+      leaders.push_back(&node);
+      return;
+    }
+
+    // Check if we are still congruent with the leader
+    if (checkNodesCongruent(*existingLeaderNode, node, context))
+    {
+      return;
+    }
```
phate: Why can it ever happen that we were once congruent with another node, but not any longer?

According to my understanding, this should never happen

jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -0,0 +1,1017 @@
...
+
+  // No existing leader top node found, create new congruence sets for each of node's outputs
+  markNodeAsLeader(node, context);
+  leaders.push_back(&node);
+}
+
+/**
+ * Places the given \p node in a congruence set, based on its operation and inputs.
+ * If the node already has a congruence set, it gets double-checked against the leader.
+ * The node must not be a top node. @see markTopNode.
```
phate: markTopNode does not exist. I think you meant markSimpleTopNode.

I suspect the reason why doxygen CI does not pick up on it is because that the docoumentation in cpp files is just removed/ignored.

jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    else if (auto structural = dynamic_cast<rvsdg::StructuralNode *>(node))
+    {
+      divertStructuralNode(*structural, context);
+    }
+    else
+    {
+      throw std::logic_error("Unknown node type");
+    }
+  }
+}
```
phate: MatchTypeOrFail

jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    // Does not include the leader.
+    util::HashSet<const rvsdg::Output *> followers;
+  };
+
+  // Type used for indexing congruence sets
+  using Index = size_t;
+  static constexpr auto NoCongruenceSet = std::numeric_limits<size_t>::max();
+
+  /**
+   * @return the number of congruence sets in the data structure. All sets have a lower index.
```
phate: I do not understand the "All sets have a lower index" part

jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    // A leader always comes earlier than its followers in the traverser.
+    const rvsdg::Output * leader;
+    // The set of followers. Can both grow and shrink during the marking phase.
+    // Does not include the leader.
+    util::HashSet<const rvsdg::Output *> followers;
+  };
+
+  // Type used for indexing congruence sets
+  using Index = size_t;
+  static constexpr auto NoCongruenceSet = std::numeric_limits<size_t>::max();
```
phate: "NoCongruenceSetIndex" ?

jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -0,0 +1,1017 @@
...
+ * the congruence set leader.
+ */
+class CongruenceSets
+{
+public:
+  /**
+   * Represents a set of T where all Ts in the set are considered identical.
+   * All the Ts must be in the same region.
+   * One T is the leader, and will eventually become the single source of truth.
+   */
```
phate: The documentation is horribly oudated.

jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -0,0 +1,1017 @@
...
+  struct CongruenceSet
+  {
+    // The set leader. Never changes.
+    // Once an output has become a leader, it will never be a follower.
+    // A leader always comes earlier than its followers in the traverser.
```
phate: I do not understand what "A leader always comes earlier than its followers in the traverser" means? Why is this piece of information here (whatever it means)?

jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -0,0 +1,1017 @@
...
+  getFollowers(Index index) const
+  {
+    JLM_ASSERT(index < sets_.size());
+    return sets_[index].followers;
+  }
+
+private:
+  // The list of congruence sets
+  std::vector<CongruenceSet> sets_;
+  // A mapping from T to the congruence set it belongs to, either as leader or follower
```
phate: Documentation is outdated

jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -0,0 +1,1017 @@
...
+static void
+markGraphImports(const rvsdg::Region & region, CommonNodeElimination::Context & context)
+{
+  auto & congruenceSets = context.getCongruenceSets();
+
+  for (auto argument : region.Arguments())
+  {
+    congruenceSets.getOrCreateSetForLeader(*argument);
+  }
+}
```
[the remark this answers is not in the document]
phate: It is a good "fist start" (not sure what was there in the first place), but maybe a FIXME to highlight this would a good idea if you do not want to fix it.

jlm/llvm/opt/CommonNodeElimination.cpp
haved: done

jlm/llvm/opt/CommonNodeElimination.cpp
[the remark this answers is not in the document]
haved: I added a FIXME. At least marking them as distinct is always sound

jlm/llvm/opt/CommonNodeElimination.cpp
haved: yup

jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -0,0 +1,1017 @@
...
+
+/**
+ * Uses the provided list of partition indices to partition the arguments of the given region.
+ * Each argument has a corresponding value in the \p partitions list.
+ * The exact partition value is not important, but only arguments with identical
+ * partition value are allowed to remain in the same congruence set.
+ * If a pair of arguments are already in different congruence sets, they will remain separate.
+ *
+ * @param region the region whose arguments should be partitioned.
+ * @param partitions integers used to partition arguments. Must have length equal to narguments.
```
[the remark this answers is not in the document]
haved: fixed

jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -0,0 +1,1017 @@
...
+ * Check if all results of the given exit variable
+ * @param exitVar
+ * @param context
+ * @return
+ */
+static std::optional<CongruenceSets::Index>
+tryGetGammaEntryVarCongruenceSet(
+    rvsdg::GammaNode::ExitVar & exitVar,
+    CommonNodeElimination::Context & context)
+{
```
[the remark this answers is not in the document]
haved: I renamed it to Exit. I agree the name was a bit weird. Added documentation

jlm/llvm/opt/CommonNodeElimination.cpp
haved: done

jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -0,0 +1,1017 @@
...
+    // Once an output has become a leader, it will never be a follower.
+    // A leader always comes earlier than its followers in the traverser.
+    const rvsdg::Output * leader;
+    // The set of followers. Can both grow and shrink during the marking phase.
+    // Does not include the leader.
+    util::HashSet<const rvsdg::Output *> followers;
+  };
+
+  // Type used for indexing congruence sets
+  using Index = size_t;
```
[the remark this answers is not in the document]
haved: done

jlm/llvm/opt/CommonNodeElimination.cpp
haved: done

jlm/llvm/opt/CommonNodeElimination.cpp
haved: fixed

PR #1357, opened 2025-10-30
Fail CI on changed cycle count

sjalander:
jlm is now determenistic, so this PR enables the CI to fail if the simulated cycle count change for the HLS test suite. A new argument is provided for the run-hls-test.sh script that updates the golden cycles, which is stored in .github/golden for easiness of updating them when a PR affects the cycle count.

### DIFF
--- a/tests/TestRvsdgs.hpp
+++ b/tests/TestRvsdgs.hpp
@@ -166,6 +166,7 @@ public:
  * \code{.c}
  *   void f()
  *   {
+ *     // Represented using allocas
  *     uint32_t a, b;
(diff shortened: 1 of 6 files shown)


sjalander:
> > Also, considering that the currently the cycle count is different from the gold cycle count, why does the CI not fail in this PR?
> 
> Any idea? Or do I misunderstand the intention of this PR?

You haven't approved the PR that will enable this feature yet. This PR is not ready to be reviewed.

phate (approved):
I think the PR description needs to be updated, otherwise this looks good.

PR #1362, opened 2025-11-04
Implement chain recurrence folding

andreaslhjulstad:
This PR implements logic for creating chain recurrences from SCEV trees. It also refactors much of the implementation by removing the first fixed-point analysis and the method to replace placeholders.

While SCEV trees can be nested, chain recurrences are one level only, and allows us to more easily reason about the step in an induction variable, something that will be useful when this is applied in an optimization. 

The reason for removing the fixed-point analysis is this: The fixed-point analysis would traverse the graph and check, for all induction variable candidates, if any illegal operations was a part of the tree, and if so, remove them from the candidate set, and we would only create SCEV trees for the valid candidates. However, we would still have to check for dependencies between these candidates, and they could still be invalid as induction variables, so we would have to filter them twice. I think a safer approach is to create SCEV trees for all loop variables, and just use SCEVUnknown for illegal operations. This means that for any output, we can still look up it's SCEV, which gives us extra information to reason about, that otherwise wouldn't be there. It is also more efficient.


phate (commented):
This looks all good to me, but I will leave the final decision to @haved

phate:
@haved Let's try to get this in as well.

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -422,35 +289,247 @@ ScalarEvolution::CreateChainRecurrences(
...
+    }
+    else
+    {
+      if (ChainRecurrenceMap_.find(scevPlaceholder->GetPrePointer()) != ChainRecurrenceMap_.end())
+      {
+        // We have a dependency of another IV
+        // Get it's saved value. This is safe to do due to the topological ordering
+        auto storedRec = ChainRecurrenceMap_.at(scevPlaceholder->GetPrePointer())->Clone();
+
+        return std::unique_ptr<SCEVChainRecurrence>(
```
haved: This return statement here is a bit surprising, as all other branches use the `chrec` to create the return value.

jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -188,23 +219,54 @@ class SCEVChrecExpr final : public SCEV
...
   }
 
+  SCEV *
+  GetStartValue() const
+  {
+    return Operands_[0].get();
+  }
+
+  void
+  SetStartValue(const std::unique_ptr<SCEV> & initScev)
```
haved: The name `SetStartValue` is not getting across that we are pushing a start value to the front.

PR #1374, opened 2025-11-06
Add fix-point computation to PredicateCorrelation pass

phate:
It could happen that we have multiple theta-gamma predicate correlations chained together. We would like to catch them all, and therefore introduce a fix-point computation to do so. I believe that this is a temporary solution until we introduced gamma-gamma predicate correlation, but this needs to be reevaluated at a later point in time.


phate:
@haved I merged master into the branch and fixed the merge conflicts.

haved (approved):

PR #1385, opened 2025-11-13
Remove Steensgaard

haved:
Due to issues that we do not know how to resolve, and the desire to change the PointsToGraph, the Steensgaard implementation is removed.


phate (changes requested):
@haved You pushed the string stuff now as well......

haved:
@phate there, should be good now

PR #1387, opened 2025-11-14
Scalar evolution statistics

andreaslhjulstad:
This is a first version of the scalar evolution statistics implementation.

I used a companion Context class, and added 6 measurements with corresponding labels. I am a bit unsure if this is the best way to go about it, so feedback is greatly appreciated.

This PR also extends the implementation to support for- and while-loops by utilizing the `tryGetConstantSignedInteger` method in `trace.hpp`, as a more robust way to get the start values for a loop variable. Thanks to @haved for the suggestion.


jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -15,6 +16,81 @@
...
+  Context &
+  operator=(const Context &) = delete;
+
+  Context &
+  operator=(Context &&) = delete;
+
+  void
+  InsertChrec(const rvsdg::ThetaNode & thetaNode, std::unique_ptr<SCEVChainRecurrence> & chrec)
+  {
+    ChrecMap_[&thetaNode].insert(std::unique_ptr<SCEVChainRecurrence>(chrec.release()));
```
haved: Instead of creating a new unique pointer from a released pointer, you can use std::move

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -54,28 +139,45 @@ ScalarEvolution::Run(
...
+        for (const auto loopVar : thetaNode->GetLoopVars())
+        {
+          Context_.get()->AddLoopVar(*loopVar.pre);
+        }
+
+        auto chrecMap = PerformSCEVAnalysis(*thetaNode);
+        for (auto & [output, chrec] : chrecMap)
+        {
+          // Add computed (non-trivially unknown) chrecs to context for statistics
+          const bool isTrivialUnknown =
```
haved: Could this logic be moved to a method on the class called `isTrivialUnknown` or maybe just `isUnknown`, with a documentation string that explains what it means for a Chain Recurrence to be unknown?

If a chain recurrence contains any step that is unknown, does it not make sense to regard the whole thing as unknown?

haved (approved):
Looks very good, I just have some comments about unique pointers @andreaslhjulstad

andreaslhjulstad:
> @andreaslhjulstad Do you want us to merge this PR?

Yeah, that would be good

PR #1391, opened 2025-11-21
Make invariant value redirection handle different sets of memory node IDs in calls and lambdas

haved:
Also renames some functions that may return `nullptr` to get a `tryXXX`, to prevent confusion.

### DIFF
--- a/jlm/rvsdg/DotWriter.cpp
+++ b/jlm/rvsdg/DotWriter.cpp
@@ -13,6 +13,28 @@ namespace jlm::rvsdg
 
 DotWriter::~DotWriter() noexcept = default;
 
+void
+DotWriter::AnnotateTypeGraphNode(const Type & type, util::graph::Node & node)
+{}
+
+void
+DotWriter::AnnotateGraphNode(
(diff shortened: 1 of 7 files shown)


jlm/llvm/ir/operators/MemoryStateOperations.cpp
```diff
@@ -333,6 +333,28 @@ LambdaEntryMemoryStateSplitOperation::copy() const
...
+    return nullptr;
+  }
+
+  if (!operation->memoryNodeIdToIndexMap_.HasKey(memoryNodeId))
+  {
+    return nullptr;
+  }
+
+  const auto index = operation->memoryNodeIdToIndexMap_.LookupKey(memoryNodeId);
+  JLM_ASSERT(index < node.noutputs());
```
phate: Should be unnecessary, as the method `output()` method below should have that check already.

haved: done

phate (approved):
LGTM

PR #1393, opened 2025-11-23
Fix compilation for gcc 13.3.0

phate:
This is the 2nd time I have to fix this. I think it is time to slowly update our CI to a newer Ubuntu version.


phate:
This is the 2nd or 3rd time I have to fix this. I think it is time to slowly update our CI to a newer Ubuntu version such that we catch these problems.

haved (approved):
Yes I had to use clang instead of gcc because of this, didn't know removing `const` would solve the false warning

PR #1396, opened 2025-11-24
Add utilities for finding/executing external programs

phate:


phate:
@caleridas @haved I am not the greatest system programmer, but I need some tools to system independently get a dot viewer. Please take it apart.

jlm/util/Program.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include <jlm/util/common.hpp>
+#include <jlm/util/Program.hpp>
+#include <jlm/util/strfmt.hpp>
+
+namespace jlm::util
+{
+
+std::filesystem::path
+tryFindExecutablePath(const std::string_view & programName)
```
haved: Const reference to string view should just be a string view by value

jlm/util/Program.cpp
```diff
@@ -0,0 +1,64 @@
...
+  JLM_ASSERT(!programPath.empty());
+  JLM_ASSERT(std::filesystem::is_regular_file(std::filesystem::status(programPath)));
+
+  std::string command = programPath.string();
+  for (auto & argument : programArguments)
+  {
+    command += strfmt(" ", argument);
+  }
+
+  return system(command.c_str());
```
haved: Using `system` is not recommended, as we need to do argument escaping ourselves.

If a user has placed a file in a folder with a space in its name, we suddenly can't open it, for example

haved (commented):
I have some comments, I will see if I can find a bit about best practices for this type of stuff in C++ @phate

jlm/util/Program.cpp
phate: Fixed.

jlm/util/Program.cpp
phate: Second try, avoiding system()

haved (dismissed):

jlm/util/Program.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+  return std::filesystem::path(result);
+}
+
+int
+executeProgramAndWait(
+    const std::filesystem::path & programPath,
+    const std::vector<std::string> & programArguments)
+{
+  const pid_t pid = fork();
```
caleridas: Please use posix_spawn instead of fork/exec

jlm/util/Program.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include <jlm/util/Program.hpp>
+#include <jlm/util/strfmt.hpp>
+
+#include <sys/wait.h>
+#include <unistd.h>
+
+namespace jlm::util
+{
+
+std::filesystem::path
```
caleridas: If you want to use path, use "bash -c" wrapper instead of searching yourself

caleridas (changes requested):
Comments online, there are more canonical ways to launch this

jlm/util/Program.cpp
phate: I was a bit uncertain here what exactly you mean, but I hope I understood you right.

caleridas (approved):
looks good, should do the job

PR #1409, opened 2025-12-01
Improved performance of the NodePullIn optimization

sjalander:
Avoids the use of unordered_set for counting nodes.


sjalander:
@haved 
I made a mistake regarding how unordered_set() works. Checking if it's the same node that's visited, then the runtime of the optimization goes through the roof again. I'm closing this PR as other improvements are needed.

PR #1412, opened 2025-12-03
[AndersenRegionAware] Fix the Mod/Ref set of function calls that might be longjumped out of

### LINKED ISSUE #1400: setjmp/longjmp can cause miscompilation with AAAndersenRegionAware
Seen in the LLVM test suite test `SingleSource/UnitTests/2008-07-13-InlineSetjmp.c`

A minimal reproduction is:
```c
#include <setjmp.h>      
#include <stdio.h>

jmp_buf g;
static int h() {
    longjmp(g,1);
}

static void f(int* x) {
    if (setjmp(g)) // setjmp returns 0 the first time it returns
        return;

    *x = 10;

    // The problem is that this call can not modify a, so is not sequentialized with the above store.
    h();
}

int main() {
    int a = 1;
    f(&a);

    // This is expected to print 10
    printf("%d\n", a);
    return 0;
}
```

The store `*x = 10` has a Mod/Ref set consisting of just `a`, as it it supposed to.

The Mod/Ref set of the `h()` function call does not include `a`, however as `h` only makes an external function call, and external functions are not supposed to be able to modify `a`. We know that `a` is non-reentrant, so any store to `a` that happens within `h()` has to be a new instance of `a`.

The above is all true, but the existence of the `longjmp` causes `h()` to behave in a weird way. The call to `longjmp()` could even be in a different module, so the call to `h()` could really be any external function, that needs to be sequentialized after the store to `a`.

I think we can work around this to keep all modules without explicit `setjmp` calls unaffected.

I guess I should also note that while `setjmp` is rare, this same problem would also be relevant for `try/catch` in C++ if we ever got there.

haved:
Fixes #1400

### DIFF
--- a/jlm/llvm/opt/alias-analyses/RegionAwareModRefSummarizer.cpp
+++ b/jlm/llvm/opt/alias-analyses/RegionAwareModRefSummarizer.cpp
@@ -66,6 +66,7 @@ class RegionAwareModRefSummarizer::Statistics final : public util::Statistics
   static constexpr auto NumSimpleAllocas_ = "#SimpleAllocas";
   static constexpr auto NumNonReentrantAllocas_ = "#NonReentrantAllocas";
   static constexpr auto NumCallGraphSccs_ = "#CallGraphSccs";
+  static constexpr auto NumFunctionsInActiveSetjmp_ = "#FunctionsInActiveSetjmp";
   static constexpr auto NumCallGraphSccsCanCallExternal_ = "#CallGraphSccsCanCallExternal";
 
   static constexpr auto CallGraphTimer_ = "CallGraphTimer";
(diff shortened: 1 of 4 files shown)


phate (approved):

haved:
@phate you are absolutely right, this code is also missing a fix that I discovered because of my test..

jlm/llvm/opt/alias-analyses/RegionAwareModRefSummarizer.cpp
```diff
@@ -546,6 +595,28 @@ CollectLambdaNodes(const rvsdg::RvsdgModule & rvsdgModule)
   return result;
 }
 
+/**
+ * Helper function for checking if the given call is a call to the special setjmp function.
+ * @param callNode the node in question
+ * @return true if the call is a direct call to setjmp, false otherwise
+ */
+static bool
+IsSetjmpCall(const rvsdg::SimpleNode & callNode)
```
phate: Small i at the beginning

jlm/llvm/opt/alias-analyses/RegionAwareModRefSummarizer.hpp
```diff
@@ -109,6 +109,15 @@ class RegionAwareModRefSummarizer final : public ModRefSummarizer
...
   CreateCallGraph(const rvsdg::RvsdgModule & rvsdgModule);
 
+  /**
+   * For each SCC in the call graph, determines if the SCC either contains all external functions,
+   * or can call into external functions, either directly or indirectly.
+   *
+   * @return the number of SCCs that can call external functions
+   */
+  size_t
+  FindSccsThatCanCallExternal();
```
phate: small f

phate (approved):

phate (approved):

PR #1414, opened 2025-12-05
Improved performance of the NodePullIn optimization

sjalander:
Avoids the use of unordered_set for counting nodes.


phate:
I had an attempt at  refactoring this in #1415

PR #1416, opened 2025-12-07
Remove incorrect JLM_ASSERT in call/lambda entry/exit reductions

phate:


haved:
@phate I agree, this should be (has to be) done by the inlining pass

PR #1421, opened 2025-12-10
Disable hoisting of nodes with states

phate:
The check only looked at the outputs of a node, but that was not sufficient any longer and lead to hoisting of load nodes. The reason for this is:

1. We can have nodes that do not have states, but still might have side-effects, such as integer division. They are held in place by IOStateBarrier nodes, which only have states as input.
2. We can have load nodes without state edges. They are also held in place by IOStateBarrier nodes.

In addition, the pruning of nodes is introduced.


phate:
@haved The PR is ready for review now.

PR #1426, opened 2025-12-15
Make traceOutput more powerful and use it in the call classifier

### LINKED ISSUE #1423: Almost redundant tracing implementation in call.cpp
To facilitate tracing from a `CallOperation` to the lambda it is calling, the call classifier has its own machinery for tracing an Output though regions up to where it comes from. 

This is very similar to the general `traceOutput` function we now have in `node.hpp` (rsvdg namespace) and `trace.hpp` (llvm namespace). There are however some important differences.

The call one does not trace though a phi fixedpoint variable argument, because it wants to distinguish between recursive and non-recursive direct calls. This is totally fine, and I think the other, general, trace functions could do the same, as the semantics of the fixed point variables are a bit special.

A more fundamental difference is that the trace function in `call.cpp` makes a much greater effort in determining if a gamma output or theta output is invariant. While the `traceOutput` family does a simple check to see if the region result(s) is directly attached to a region argument(s), the one in `call.cpp` actually goes into the subregion to check if it can trace its way to a region argument.

First of all I think the code for doing this in gamma nodes is wrong.

Second, I'm wondering how much effort it makes sense to do when tracing. If we have done Invariant Value Redirection recently, there should not really be any reason to trace into subregions.

We could also imagine being even more extreme in our attempts at tracing, and start to do a sort of common output coalescing. This would only be able to give more information if common node elimination has not been run recently.

I would like to avoid having two tracing functions, but at the same time I think it is nice for the general tracing function to not have to build a hash map of the world as it is tracing. I think it might be possible to get a good compromise where we make the current `traceOutput` able to trace through anything that the current `call.cpp` trace function can, but without needing to maintain any state during tracing. In that case I think we should remove the custom tracing function from `call.cpp` and use the general one everywhere. 

@phate

haved:
The `CallOperation`'s classifier contained output tracing machinery that did mostly the same as the regular `traceOutput` and friends, but was a bit more capable, as it would traverse into subregions to try extra hard to see if outputs are invariant.

This PR makes the general `traceOutput` as powerful, in a way that does not require state to be stored. The call classifier is changed to make use of it instead of having its own duplicated logic.

I do think the solution we currently have to handle IOBarrier from LLVM is a bit bad, as it requires creating a separate `traceOutput` for when we are inside the LLVM dialect. It would make more sense if operations had a general way of reporting that some outputs are invariant, but that is left for later.

Fixes #1423

### DIFF
--- a/tests/jlm/rvsdg/TraceTests.cpp
+++ b/tests/jlm/rvsdg/TraceTests.cpp
@@ -8,8 +8,8 @@
 #include <test-types.hpp>
 
 #include <jlm/rvsdg/gamma.hpp>
-#include <jlm/rvsdg/theta.hpp>
 #include <jlm/rvsdg/simple-node.hpp>
+#include <jlm/rvsdg/theta.hpp>
 #include <jlm/rvsdg/Trace.hpp>
 #include <jlm/rvsdg/view.hpp>
 


jlm/rvsdg/Trace.hpp
```diff
@@ -0,0 +1,138 @@
...
+class OutputTracer
+{
+public:
+  virtual ~OutputTracer();
+
+  /**
+   * Creates an OutputTracer with the given configuration
+   * @param isDeep if true, tracing goes into subregions, trying to trace through structural nodes.
+   *        Otherwise, outputs of structural nodes use a simple invariant check.
+   * @param isInterprocedural allow tracing out of loops
```
phate: Did you mean "allow tracing out of lambdas" here? Otherwise, this makes no sense to me as interprocedural does not mean "out of loops".

jlm/rvsdg/Trace.hpp
```diff
@@ -0,0 +1,138 @@
...
+   */
+  [[nodiscard]] Output *
+  tryTraceThroughGamma(GammaNode & gammaNode, Output & output);
+
+  /**
+   * Attempts to trace the output of a theta node through the node.
+   * This is only possible if the output is invariant within the theta node.
+   * @param thetaNode the gamma node to trace through
+   * @param output an output of the given gamma node
+   * @return the origin of the output value on the input side of the gamma, or nullptr.
```
phate: input side of the theta

jlm/rvsdg/Trace.hpp
```diff
@@ -0,0 +1,138 @@
...
+   */
+  [[nodiscard]] Output *
+  tryTraceThroughTheta(ThetaNode & thetaNode, Output & output);
```
phate: Same with "@pre" here

jlm/rvsdg/Trace.hpp
```diff
@@ -0,0 +1,138 @@
...
+   * @param output the output to trace from
+   * @param mayLeaveRegion if false, tracing stops if it reaches a region argument
+   * @return the result of tracing from the given output, if possible. Otherwise, \p output.
+   */
+  [[nodiscard]] virtual Output &
+  traceStep(Output & output, bool mayLeaveRegion);
+
+  // When true, tracing enters subregions of structural nodes to check if the value is invariant.
+  // When false, values are only considered invariant if they are directly connected to arguments.
+  bool isDeep_;
```
phate: Suggestion: "traceInStructuralNodes"?

jlm/rvsdg/Trace.cpp
```diff
@@ -0,0 +1,233 @@
...
+    if (const auto ctxVar = lambda->MapBinderContextVar(output))
+      return *ctxVar->input->origin();
+
+    return output;
+  }
+
+  // Handle delta context variables
+  if (const auto delta = rvsdg::TryGetRegionParentNode<DeltaNode>(output))
+  {
+    // If the argument is a contex variable, continue normalizing
```
phate: continue tracing

haved:
I tended to your comments and added a test with nested structural nodes, that also tests having `testInStructuralNodes` set to `false`. @phate

One of your comments I did not implement, as using `ThetaLoopVarIsInvariant` is not powerful enough to get the nested behavior. We get a kind of circular dependency between deep invariance checking and tracing, which I'm not entirely sure how to handle.

PR #1427, opened 2025-12-16
Reimplement RegionArgument Removal

phate:
The performance of the sweep phase of DNE is currently quadratic in the number of arguments/results/inputs/outputs. The reason for this is that we remove arguments/results/inputs/outputs one at a time and need to readjust indices for all other remaining arguments/results/inputs/outputs while doing so. I would like to move away from the individual removal of arguments/results/inputs/outputs towards mass removal.

This is the first step in this direction by reimplementing `Region::RemoveArgumentsWhere()` to not rely on the individual removal methods. I will adjust the other `Remove*Where()` methods in future PRs, and then utilize them in the removal functions of the structural nodes.


haved (dismissed):

caleridas (dismissed):
Cannot say much, but I am not certain where / if this is even called anymore, it does not appear to be called at all: Since most (all?) things should go the more specific removal functions (e.g. GammaNode::RemoveExitVars) I am not fully certain where exactly it would end up being used.

caleridas:
> > Cannot say much, but I am not certain where / if this is even called anymore, it does not appear to be called at all: Since most (all?) things should go the more specific removal functions (e.g. GammaNode::RemoveExitVars) I am not fully certain where exactly it would end up being used.
> 
> @caleridas Yes, the more specific removal functions all call `RemoveArgument()` under the hood, which leads to the quadratic behavior. This is exactly what I want to replace with this method.

I see -- but they all materialize the set of things removed already, so you might as well change / specialize signature slightly to get a "vector of things to be removed". Might want to sort again for safety to be sure, but might as well also rely on pre-sorting.

phate:
@caleridas @haved I changed the interface again, as I realized that this was not really working as planned when I tried to use it. There is some code duplication there right now, but I would like to ask you to ignore it. I will straighten this out later. Hell, I am not even convinced that we actually need the `RemoveArguments()` method in the first place. The `PruneArguments()` might actually be sufficient. The only place where we need a "RemoveWithIndices" function is the inputs and results as they are the entry points in the removal (arguments/outputs cannot be removed before they are dead, which only happens if the inputs/results have been removed, so the simple prune functions should be sufficient I believe).

PR #1429, opened 2025-12-18
Replace Node::RemoveInputsWhere() with RemoveInputs()

phate:


jlm/rvsdg/node.cpp
```diff
@@ -325,6 +325,43 @@ Node::removeInput(size_t index, bool notifyRegion)
...
+    {
+      if (notifyRegion)
+        region()->notifyInputDestroy(input.get());
+      input.reset();
+      numRemovedInputs++;
+    }
+    else
+    {
+      input->index_ = numLiveInputs;
+      inputs_[numLiveInputs++] = std::move(input);
```
haved: Should there be a if(numLiveInputs < n) to avoid moving onto itself, and updating the index with the same index?

PR #1432, opened 2025-12-19
Inline pass fix

haved:
Fixes the inlining pass in the following ways:
 - inlining is not performed if the callee has
   - Varargs
   - Calls to `setjmp`
   - Calls to `alloca` with runtime known sizes
   - Calls to `alloca` that are not on the top level

When inlining is performed, it performs the following extra steps:
 - Matches up `CallEntryMerge`/`LambdaEntrySplit` and `LambdaExitMerge`/`CallExitSplit` to route memory states directly, or completely around the function body if the memory state id is not used inside the callee.
 - All alloca operations in the callee are hoised to the top level of the caller

In order to do tracing correctly even in the case of functions in phi nodes, I had to add an additional configuration option to the `OutputTracer` that controls if tracing is allowed to enter into phi nodes. We often want to go into phi nodes to find the actual lambda we are referring to, but this risks tracing descending into sibling regions, and not just going up through the region hierarchy.

Adds tests of the alloca handling in inlining.

### DIFF
--- a/scripts/run-llvm-test-suite.sh
+++ b/scripts/run-llvm-test-suite.sh
@@ -3,7 +3,7 @@ set -eu
 
 # URL to the benchmark git repository and the commit to be used
 GIT_REPOSITORY=https://github.com/phate/llvm-test-suite.git
-GIT_COMMIT=f34f7d42a4759b8338afa11419ee6987b2e7dfee
+GIT_COMMIT=15a95578d37837bb121b7443b6bfd2d5e45a78f7
 
 # Get the absolute path to this script and set default JLM paths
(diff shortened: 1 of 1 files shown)


sjalander:
@haved 
Ideally, this one should include an updated version of the llvm-test-suite with the inline transformation enabled for the CI.

phate:
> @haved 
> Ideally, this one should include an updated version of the llvm-test-suite with the inline transformation enabled for the CI.

That might not work as we have not tested it with all of love test suite. The problems arose only from a test with the single source test. Let's not for now.

PR #1451, opened 2025-12-25
Replace types for testing

phate:
This PR does the following:
1. Implements a configurable TestType in the jlm::rvsdg namespace
2. Replaces all uses of jlm::tests::ValueType and jlm::tests::StateType with the new jlm::rvsdg::TestType
3. Removes test-types.hpp and test-types.cpp, which contained the jlm::tests::ValueType and jlm::tests::StateType

The goal is to get the libjlmtests library completely dissolved as it is in the way to convert all unit tests to google tests. It contains a main() definition, which does not go well with google tests.

### DIFF
--- a/jlm/llvm/opt/ScalarEvolution.cpp
+++ b/jlm/llvm/opt/ScalarEvolution.cpp
@@ -127,7 +127,8 @@ public:
 };
 
 ScalarEvolution::ScalarEvolution()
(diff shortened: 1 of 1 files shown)


haved (approved):
Looks good, but clang format is not happy @phate

haved (dismissed):

PR #1453, opened 2025-12-25
Use unique_ptr for operation argument of ThreeAddressCode class

phate:


haved (dismissed):
Looks good, one comment. And HLS is not happy @phate

phate:
@haved @sjalander Updated golden cycles and addressed comment.

PR #1455, opened 2025-12-27
Make ControlType a value type

phate:


haved (approved):
I think this change makes a lot of sense,  but I don't know if there are any unforseen consequences

PR #1467, opened 2025-12-31
alias analysis: use MatchType to match against operation classes

caleridas:
Prefer to use MatchType* variants instead of matching against operation classes using an if/is cascade.


phate:
@haved I leave that one to you

PR #1473, opened 2026-01-04
Remove outdated libjlmtest

### LINKED ISSUE #1403: Convert all unit tests to gtest
We currently have our own home grown and limited unit test infrastructure. We should convert to google tests to streamline our infrastructure.

phate:
All unit tests have been converted to google tests. We do not have any use for the libjlmtest any longer.

Closes #1403


phate:
@caleridas Can we remove any makefile rules regarding the old unit test infrastructure?

For example, those here: https://github.com/phate/jlm/blob/0312f486efb8f9efc6583909d245eb9580b1f881/Makefile.macros#L9C1-L18C73 ?

PR #1489, opened 2026-01-08
Adds support for literal information in the struct type

sjalander:
LLVM has two types of structs, literal and identified. This information was not explicitly kept in the RVSDG struct type, which has been added.

### DIFF
--- a/jlm/llvm/ir/types.hpp
+++ b/jlm/llvm/ir/types.hpp
@@ -196,7 +196,7 @@ public:
         isLiteral_(isLiteral)
   {}
 
-  StructType(const StructType &) = default;
+  StructType(const StructType &) = delete;
 
   StructType(StructType &&) = delete;
 


jlm/llvm/ir/types.hpp
```diff
@@ -254,21 +268,22 @@ class StructType final : public rvsdg::Type
...
-    return std::make_shared<StructType>(isPacked, declaration);
+    return std::make_shared<StructType>(declaration, isPacked, isLiteral);
   }
 
 private:
-  bool IsPacked_;
   std::string Name_;
   const Declaration & Declaration_;
+  bool IsPacked_;
+  bool IsLiteral_;
```
phate: `isLiteral_`

phate:
@haved You looked into this recently. I would like your review on it as well, please.

sjalander:
@phate @haved 
See the Detailed Description
https://llvm.org/doxygen/classllvm_1_1StructType.html

PR #1494, opened 2026-01-11
Introduce IDs for regions

phate:


phate:
@caleridas It is invaluable to be able to stop with a conditional breakpoint in the debugger at the correct node/region where there is a problem. I would not be able to know how otherwise to debug bigger graphs. The plan is also to have those IDs in the viewer such that we can visually debug the problems (otherwise the offending region is impossible to find).

PR #1502, opened 2026-01-16
Extend scalar evolution with support for multiplication and subtraction

andreaslhjulstad:
This PR extends the existing scalar evolution analysis implementation by adding support for handling subtraction and multiplication operations. Previously, statements that included these operations were treated as unknown. 

More specifically, it adds cases for treating subtraction and multiplication operations. For multiplication, (two) new `SCEV` types are added, a `SCEVMulExpr` and a `SCEVNAryMulExpr`, mirroring the `SCEVAddExpr` and `SCEVNAryAddExpr` for addition operations, a new method for folding multiplication operations, `ApplyMultFolding()`, is also added. Subtraction however, does not have its own SCEV type(s), but is rather treated as an addition with the right hand-side of the expression being negated, and so it does not require as much logic to support. A new method called `GetNegativeSCEV()` is added for this purpose.

In order to support different operations, we also need to keep track of which operation with which a variable depends on another (or itself). As a result, the dependency logic (necessary for invalidating certain cases and topological sorting) has been extended to include the operation as well as the number of times a variable depends on another.

Currently, the implementation treats variables that have direct dependencies with multiplication (i.e. a = a * 2) as invalid, whilst other cases of multiplication that don't result in geometric sequences are OK, (e.g. a = a + b * 2). This is mainly because these are not able to be modeled as affine recurrences and because the main focus of LLVM's SCEV is linear and polynomial sequences, and these are the ones that are most relevant for optimizations like loop strength reduction and induction variable substitution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,11 @@ The MLIR RVSDG dialect and the MLIR backend can be setup with the following comm
 An introduction to the RVSDG and the optimizations supported by jlm can be found in the
 following articles:
 
+H. R. Krogstie, H. Bahmann, M. Själander, and N. Reissmann
+*"PIP: Making Andersen's Points-to Analysis Sound and Practical for Incomplete C Programs"*
+Proceedings of the ACM/IEEE International Conference on Code Generation and Optimization (CGO), Feb. 2026.
+https://www.arxiv.org/pdf/2512.07299
+
 D. Metz, N. Reissmann, and M. Själander
 *"R-HLS: An IR for Dynamic High-Level Synthesis and Memory Disambiguation based on Regions and State Edges"*
 Proceedings of the ACM/IEEE International Conference on Computer-Aided Design (ICCAD), Oct. 2024.
--- a/jlm/llvm/opt/CommonNodeElimination.cpp
+++ b/jlm/llvm/opt/CommonNodeElimination.cpp
@@ -642,29 +642,37 @@ markSubregionsFromInputs(
 
   for (auto & subregion : node.Subregions())
   {
-    // create a partitioning of the region arguments
-    std::vector<size_t> partitions(subregion.narguments());
-    // Arguments that do not belong to any input are given partition keys higher than any real index
-    size_t nextUniquePartitionKey = context.numCongruenceSets();
-
-    for (const auto argument : subregion.Arguments())
+    if (subregion.narguments() == 0)
     {
-      if (const auto input = argument->input())
-      {
-        // If the argument corresponds to an input, use the partition key of the input
-        partitions[argument->index()] = context.getSetFor(*input->origin());
-      }
-      else
-      {
-        // Otherwise make sure the argument is not partitioned with any other argument
-        partitions[argument->index()] = nextUniquePartitionKey++;
-      }
-    }
-
-    if (partitionArguments(subregion, partitions, context))
-    {
-      anyChanges = true;
       markRegion(subregion, context);
+    }
+    else
+    {
+      // create a partitioning of the region arguments
+      std::vector<size_t> partitions(subregion.narguments());
+      // Arguments that do not belong to any input are given partition keys higher than any real
+      // index
+      size_t nextUniquePartitionKey = context.numCongruenceSets();
+
+      for (const auto argument : subregion.Arguments())
+      {
+        if (const auto input = argument->input())
+        {
+          // If the argument corresponds to an input, use the partition key of the input
+          partitions[argument->index()] = context.getSetFor(*input->origin());
+        }
+        else
+        {
+          // Otherwise make sure the argument is not partitioned with any other argument
+          partitions[argument->index()] = nextUniquePartitionKey++;
+        }
+      }
+
+      if (partitionArguments(subregion, partitions, context))
+      {
+        anyChanges = true;
+        markRegion(subregion, context);
+      }
     }
   }
 
--- a/jlm/llvm/opt/LoopUnswitching.cpp
+++ b/jlm/llvm/opt/LoopUnswitching.cpp
@@ -6,6 +6,7 @@
 #include <jlm/llvm/ir/operators.hpp>
 #include <jlm/llvm/ir/RvsdgModule.hpp>
 #include <jlm/llvm/opt/LoopUnswitching.hpp>
+#include <jlm/llvm/opt/PredicateCorrelation.hpp>
 #include <jlm/llvm/opt/pull.hpp>
 #include <jlm/rvsdg/gamma.hpp>
 #include <jlm/rvsdg/theta.hpp>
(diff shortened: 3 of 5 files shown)


jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -162,6 +162,13 @@ ScalarEvolution::IsUnknown(const SCEVChainRecurrence & chrec)
   return false;
 }
 
+bool
+ScalarEvolution::IsState(const rvsdg::Output & output)
```
phate: Could you not just use `output->Type()->Kind() == Kind::State` instead of this function?

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -178,14 +185,18 @@ ScalarEvolution::AnalyzeRegion(const rvsdg::Region & region)
...
-          Context_.get()->AddLoopVar(*loopVar.pre);
+          if (!IsState(*loopVar.pre)) // Only add loop variables that are not states
+            Context_.get()->AddLoopVar(*loopVar.pre);
         }
 
         auto chrecMap = PerformSCEVAnalysis(*thetaNode);
         for (auto & [output, chrec] : chrecMap)
         {
-          if (!IsUnknown(*chrec))
+          if (!(IsUnknown(*chrec) || IsState(*output)))
```
phate: My brain does not deal with this well. I would rather express this as `!IsUnknown() && !IsState()`

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -200,6 +211,34 @@ ScalarEvolution::TryGetSCEVForOutput(const rvsdg::Output & output)
...
+  // -(c)
+  if (const auto c = dynamic_cast<const SCEVConstant *>(&scev))
+  {
+    const auto value = c->GetValue();
+    return std::make_unique<SCEVConstant>(-value);
+  }
+  // -(-x) -> x
+  if (const auto mul = dynamic_cast<const SCEVMulExpr *>(&scev))
+  {
+    if (const auto c = dynamic_cast<const SCEVConstant *>(mul->GetLeftOperand());
```
phate: What about the rhs operand? The expression could also be x * -1?

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -229,9 +268,33 @@ ScalarEvolution::GetOrCreateSCEVForOutput(const rvsdg::Output & output)
...
+      const auto lhs = simpleNode->input(0)->origin();
+      const auto rhs = simpleNode->input(1)->origin();
+
+      const auto lhsScev = GetOrCreateSCEVForOutput(*lhs);
+      const auto rhsScev = GetOrCreateSCEVForOutput(*rhs);
+
+      result = std::make_unique<SCEVMulExpr>(lhsScev->Clone(), rhsScev->Clone());
     }
   }
   // TODO: Handle more cases
```
phate: What cases are missing? Could this TODO be removed?

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -503,21 +596,113 @@ ScalarEvolution::CreateChainRecurrence(
...
+  }
+  if (const auto scevMulExpr = dynamic_cast<const SCEVMulExpr *>(&scevTree))
+  {
+    /* We have the following folding rules from the CR algebra:
+     * G * {e,+,f}         =>       {G * e,+,G * f}
+     * {e,+,f} * {g,+,h}   =>       {e * g,+,e * h + f * g + f * h,+,2*f*h}
+     *
+     * AFAIK these are the only rules that we are able to support as there is no general rule for
+     * folding two addrecs of arbitrary length with the multiplication operation
+     */
```
phate: Please use `//`

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -532,89 +717,136 @@ isNonZeroConstant(const SCEVConstant * c)
...
-   * number of operands), and add this to the chrec. Folding two of these n-ary add expressions will
-   * result in another n-ary add expression, which consists of all the operands in both the left and
-   * the right expression.
+   * For constants and unknowns this is trivial, however it becomes a bit complicated when we
+   * factor in SCEVInit nodes. These nodes represent the initial value of an IV in the case where
+   * the exact value is unknown at compile time. E.g. function argument or result from a
+   * call-instruction. In the cases where we have to fold one or more of these init-nodes, we create
+   * an n-ary add expression (add expression with an arbitrary number of operands), and add this to
+   * the chrec. Folding two of these n-ary add expressions will result in another n-ary add
+   * expression, which consists of all the operands in both the left and the right expression.
```
phate: `//`

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -532,89 +717,136 @@ isNonZeroConstant(const SCEVConstant * c)
...
+  const auto lhsUnknown = dynamic_cast<const SCEVUnknown *>(lhsOperand);
+  const auto rhsUnknown = dynamic_cast<const SCEVUnknown *>(rhsOperand);
+  const auto lhsInit = dynamic_cast<const SCEVInit *>(lhsOperand);
+  const auto rhsInit = dynamic_cast<const SCEVInit *>(rhsOperand);
+  const auto lhsNAryAddExpr = dynamic_cast<const SCEVNAryAddExpr *>(lhsOperand);
+  const auto rhsNAryAddExpr = dynamic_cast<const SCEVNAryAddExpr *>(rhsOperand);
+  const auto lhsNAryMulExpr = dynamic_cast<const SCEVNAryMulExpr *>(lhsOperand);
+  const auto rhsNAryMulExpr = dynamic_cast<const SCEVNAryMulExpr *>(rhsOperand);
+  const auto lhsConstant = dynamic_cast<const SCEVConstant *>(lhsOperand);
+  const auto rhsConstant = dynamic_cast<const SCEVConstant *>(rhsOperand);
```
phate: Push these casts to the respective blocks where they are utilized first. There is no need to perform all the casts first, just to realize that they are actually never used.

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -628,11 +860,135 @@ ScalarEvolution::ApplyFolding(SCEV * lhsOperand, SCEV * rhsOperand)
...
+  const auto lhsUnknown = dynamic_cast<const SCEVUnknown *>(lhsOperand);
+  const auto rhsUnknown = dynamic_cast<const SCEVUnknown *>(rhsOperand);
+  const auto lhsInit = dynamic_cast<const SCEVInit *>(lhsOperand);
+  const auto rhsInit = dynamic_cast<const SCEVInit *>(rhsOperand);
+  const auto lhsNAryMulExpr = dynamic_cast<const SCEVNAryMulExpr *>(lhsOperand);
+  const auto rhsNAryMulExpr = dynamic_cast<const SCEVNAryMulExpr *>(rhsOperand);
+  const auto lhsNAryAddExpr = dynamic_cast<const SCEVNAryAddExpr *>(lhsOperand);
+  const auto rhsNAryAddExpr = dynamic_cast<const SCEVNAryAddExpr *>(rhsOperand);
+  const auto lhsConstant = dynamic_cast<const SCEVConstant *>(lhsOperand);
+  const auto rhsConstant = dynamic_cast<const SCEVConstant *>(rhsOperand);
```
phate: Same here, push them down.

jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -361,11 +451,31 @@ class ScalarEvolution final : public jlm::rvsdg::Transformation
   class Statistics;
 
 public:
+  enum class DependencyOp
+  {
+    Add,
+    Mult,
+    Direct,
```
phate: What means Direct? I think some documentation would be useful here.

jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -361,11 +451,31 @@ class ScalarEvolution final : public jlm::rvsdg::Transformation
...
+  };
+
+  struct DependencyInfo
+  {
+    int count;
```
phate: count of what?

phate (commented):
Mostly minor comments. Looks good otherwise.

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -532,89 +687,134 @@ isNonZeroConstant(const SCEVConstant * c)
...
+  {
+    // Multiply expression with nonzero constant - create add expression
+    const auto * mulExpr = lhsNAryMulExpr ? lhsNAryMulExpr : rhsNAryMulExpr;
+    const auto * constant = lhsConstant ? lhsConstant : rhsConstant;
+    return std::make_unique<SCEVNAryAddExpr>(mulExpr->Clone(), constant->Clone());
+  }
+
+  if (lhsNAryMulExpr || rhsNAryMulExpr)
+  {
+    // Single multiply expression (other is zero)
```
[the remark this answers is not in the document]
andreaslhjulstad: That is bad wording on my part. This check is only true if there is only one operand. In that case, there is nothing to fold, so we just return a clone of the element. Changed the comment now

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -246,38 +309,39 @@ ScalarEvolution::GetOrCreateSCEVForOutput(const rvsdg::Output & output)
   return result;
 }
 
-std::unordered_map<const rvsdg::Output *, int>
-ScalarEvolution::FindDependenciesForSCEV(const SCEV & currentSCEV, const rvsdg::Output & currentIV)
+void
+ScalarEvolution::FindDependenciesForSCEV(
+    const SCEV & scev,
+    DependencyMap & dependencies,
+    const DependencyOp op = DependencyOp::None)
```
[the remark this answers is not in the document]
andreaslhjulstad: It is really only used in the method below, in order to filter out geometric sequences that we don't support (i.e. `a = a * 2`). The reason it is passed down as an argument in `FindDependenciesForSCEV()` is because the `Add` operation is less strict than the `Mul` operation, so it takes precedence. E.g. `a = a + 2 * b` is a valid (polynomial) sequence, so the `Mul` operation should not overload the `Add` dependency between a and b.

```c++
bool
ScalarEvolution::IsValidInductionVariable(
    const rvsdg::Output & variable,
    IVDependencyGraph & dependencyGraph)
{
  // First check that variable has only one self-reference,
  if (dependencyGraph[&variable][&variable].count != 1)
    return false;

  // Check that it has no reference via a mult-operation
  // (results in a geometric sequence - which we treat as an invalid induction variable)
  auto deps = dependencyGraph[&variable];
  for (auto [output, dependencyInfo] : deps)
  {
    if (dependencyInfo.operation == DependencyOp::Mul)
    {
      return false;
    }
  }

  // Then check for cycles through other variables
  std::unordered_set<const rvsdg::Output *> visited{};
  std::unordered_set<const rvsdg::Output *> recursionStack{};
  return !HasCycleThroughOthers(variable, variable, dependencyGraph, visited, recursionStack);
}
```

PR #1511, opened 2026-01-19
LoopUnswitching: remove superfluous node sinking into gamma 

phate:
This preparation step was unnecessary as the transformation never kicked in without this being fulfilled anyway.


phate:
This was verified [here](https://github.com/phate/jlm/pull/1511/changes/53c0894275575649f746c769e7a5f0e20aef6f87). I would appreciate that #1512 gets merged first though.

haved (approved):

PR #1513, opened 2026-01-19
LoopUnswitching: factor out exit region handling

phate:


phate:
@haved I resolved the conflict.

haved (approved):

PR #1516, opened 2026-01-21
Store value forwarding

haved:
This first attempt uses the LocalAliasAnalysis during tracing to find candiates for forwarding. It is only able to handle cases where the store is in the same region as the load, or in a parent region of the load.

Tests are missing

### DIFF
--- a/jlm/llvm/opt/StoreValueForwarding.cpp
+++ b/jlm/llvm/opt/StoreValueForwarding.cpp
@@ -86,11 +86,10 @@ StoreValueForwarding::StoreValueForwarding()
 void
 StoreValueForwarding::traverseInterProceduralRegion(rvsdg::Region & region)
 {
-  rvsdg::TopDownTraverser traverser(&region);
-  for (auto node : traverser)
+  for (auto & node : region.Nodes())
   {
     rvsdg::MatchTypeOrFail(
-        *node,
+        node,
         [&](rvsdg::PhiNode & phiNode)
         {
           traverseInterProceduralRegion(*phiNode.subregion());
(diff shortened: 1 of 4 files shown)


jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+StoreValueForwarding::StoreValueForwarding()
+    : Transformation("StoreValueForwarding")
+{}
+
+void
+StoreValueForwarding::traverseInterProceduralRegion(rvsdg::Region & region)
+{
+  rvsdg::TopDownTraverser traverser(&region);
+  for (auto node : traverser)
```
phate: Do we actually need a traverser here? We can just iterate through the nodes, no?

jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -0,0 +1,318 @@
...
+void
+StoreValueForwarding::processSimpleNode(rvsdg::SimpleNode & node)
+{
+  if (is<LoadNonVolatileOperation>(&node))
+  {
+    processLoadNode(node);
+  }
+
+  // For other node types, we don't need to do anything for store value forwarding
+}
```
phate: This function is not really worth, is it? I suggest to inline it.

jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    // If we already have a different commonStoreNode, forwarding is not possible
+    if (commonStoreNode)
+      return false;
+
+    // Only allow forwarding values with identical types
+    auto storedType = StoreOperation::StoredValueType(foundStoreNode);
+    if (*storedType != *loadedType)
+      return false;
```
phate: Push the computation of the loadedType down here. There is no need to compute it up when it might actually not be used.

PR #1521, opened 2026-01-28
ScalarEvolution: Implement support for nested loops

andreaslhjulstad:
This PR adds two new methods: `CombineChrecsAcrossLoops()` and `TryReplaceInitForSCEV()` in the ScalarEvolution class which enable computing fully-fledged recurrences across different loops (both nested and side-by-side). Also stores all chrecs and SCEVs in the companion context class to enable access to chrecs and SCEVs from all loops, it also simplifies the logic quite a bit.

The main idea for computing recurrences across loops is to compute recurrences for all the loops separately, and then stitch them together afterwards. This is done by recursively traversing the recurrences and trying to replace the `SCEVInit` nodes in the recurrence with a recurrence that has (hopefully) been computed in an analysis of a different loop. Both of the new methods run in a fixed-point manner in order to support arbitrary levels of nested loops.

### DIFF
--- a/jlm/llvm/opt/ScalarEvolution.cpp
+++ b/jlm/llvm/opt/ScalarEvolution.cpp
@@ -949,8 +949,6 @@ ScalarEvolution::ApplyAddFolding(const SCEV * lhsOperand, const SCEV * rhsOperan
 std::unique_ptr<SCEV>
 ScalarEvolution::ApplyMulFolding(const SCEV * lhsOperand, const SCEV * rhsOperand)
 {
-  // Apply folding rules for multiplication to combine two operands into one
-  //
   // We have the following folding rules from the CR algebra:
   // G * {e,+,f}         =>       {G * e,+,G * f}
   // {e,+,f} * {g,+,h}   =>       {e * g,+,e * h + f * g + f * h,+,2*f*h}


jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -229,40 +259,54 @@ class SCEVMulExpr final : public SCEVBinaryExpr
...
+  AddOperand(std::unique_ptr<SCEV> scev)
   {
-    return Loop_;
+    Operands_.push_back(std::move(scev));
   }
 
-  SCEV *
-  GetStartValue() const
+  void
+  SwapOperand(const size_t index, const std::unique_ptr<SCEV> & operand)
```
phate: This is more of a `replaceOperand()`, no?

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -141,37 +161,25 @@ ScalarEvolution::Run(
...
-    {
-      return true;
-    }
-  }
-  return false;
-}
-
 void
-ScalarEvolution::AnalyzeRegion(const rvsdg::Region & region)
+ScalarEvolution::AnalyzeRegion(const rvsdg::Region & region, Context & ctx)
```
phate: I fail to understand why you leave those methods static even though they need a Context. Why do you not take the static modifier away and then they do not require the context as argument, but can directly access it in the class as they are a member method.

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -550,149 +603,111 @@ ScalarEvolution::CreateChainRecurrence(
...
-    const auto rhsChrec = CreateChainRecurrence(IV, *scevMulExpr->GetRightOperand(), thetaNode);
+    return clone_as<SCEVChainRecurrence>(*ApplyMulFolding(lhsStep.get(), rhsStep.get()));
+  }
+  return chrec;
+}
 
-    const auto lhsSize = lhsChrec->GetOperands().size();
-    const auto rhsSize = rhsChrec->GetOperands().size();
+bool
+isNonZeroConstant(const SCEVConstant * c)
```
phate: Would it not be better to implement as a method on the class?

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -36,59 +35,95 @@ class ScalarEvolution::Context final
...
+    for (auto & [out, chrec] : ChrecMap_)
     {
-      for (auto & chrec : chrecs.Items())
-      {
-        // Count chrecs with specific order
-        if (static_cast<int>(chrec->GetOperands().size()) == n + 1)
-          count++;
-      }
+      // Count chrecs with specific order
+      if (static_cast<int>(chrec->GetOperands().size()) == n + 1 && !IsUnknown(*chrec))
```
phate: Why is this static_cast<> here necessary?

I believe(!) that the static_cast<> can be thrown out if you convert the the parameter `n` to a `size_t` instead of an `int`. It makes no sense to have an `int` there anyway, as a negative order makes no sense, right?

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -133,6 +168,18 @@ ScalarEvolution::ScalarEvolution()
 
 ScalarEvolution::~ScalarEvolution() noexcept = default;
 
+std::unique_ptr<ScalarEvolution::Context>
+ScalarEvolution::CreateContext()
```
phate: Is this method really necessary? The context is only used internally, and this function just contains a simple function call and nothing else.

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -181,79 +216,208 @@ ScalarEvolution::AnalyzeRegion(const rvsdg::Region & region)
...
   {
-    return std::make_unique<SCEVAddExpr>(
-        GetNegativeSCEV(*add->GetLeftOperand()),
-        GetNegativeSCEV(*add->GetRightOperand()));
+    if (std::find(order.begin(), order.end(), loopVar.pre) == order.end())
+      // This is not a valid IV, so it hasn't been added yet
+      allVars.push_back(loopVar.pre);
+  }
+
+  for (size_t i = 0; i < order.size(); ++i)
```
phate: `for (auto ....)`

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -550,149 +627,117 @@ ScalarEvolution::CreateChainRecurrence(
...
-        chrec->AddOperand(lhsOperand->Clone());
-      }
-    }
-    // Handle G * {e,+,f,...} where G is loop invariant
-    else if (lhsSize == 1)
-    {
-      // G * {e,+,f,...} = {G * e,+,G * f,...}
-      auto lhsOperand = lhsChrec->GetOperand(0);
+bool
+ScalarEvolution::IsNonZeroConstant(const SCEVConstant * c)
```
phate: I meant on the SCEVConstant class, not the ScalarEvolution class.

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -181,79 +216,208 @@ ScalarEvolution::AnalyzeRegion(const rvsdg::Region & region)
...
+    }
+  }
+  if (const auto nArySCEV = dynamic_cast<const SCEVNAryExpr *>(&scev))
+  {
+    // An n-ary scev is any scev with an arbitrary number of operands: chain recurrence, n-ary add
+    // and n-ary mult. We want to recursively check all it's operands for Init nodes
+    auto clone = CloneAs<SCEVNAryExpr>(*nArySCEV);
+    const auto operands = nArySCEV->GetOperands();
+    bool changed = false;
+    for (size_t i = 0; i < operands.size(); ++i)
```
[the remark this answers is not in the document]
andreaslhjulstad: I personally think iterating on index here is the best solution as we need to keep track of the index to know which operand to replace. If I were to change to a range-based loop, we would need to keep track of the index manually, which seems like unnecessary work.

PR #1526, opened 2026-02-10
LoopUnswitching: Reimplement transformation

### LINKED ISSUE #895: Fix bug in ThetaGammaInversion
The bug was reported by @davidmetz in #884 :

I actually managed to find an example where TGI produces wrong behavior.
The examples I previously had just duplicated the loop into both gamma branches, but this is worse.
In this kernel a store happens in a loop - except in the last iteration.
The current TGI wrongly pulls out the gamma, since it shares the predicate with the theta.
```
void kernel(int *a) {
    int i = 0;
    do {
        i++;
        if (i != 10)
            a[i] = i;
    } while (i != 10);
}

int main(int argc, char **argv) {
    int a[11] = {0};
    kernel(a);
    assert(a[0] == 0);
    for (int i = 1; i < 10; ++i) {
        assert(a[i] == i);
    }
    assert(a[10] == 0);
    return 0;
}
```
before TGI:
image
after TGI:
image

phate:
This PR reimplements the loop unswitching transformation. It performs the following changes compared to the previous implementation:
1. Fixes a bug where the wrong origins for variables was picked up. In other words, the old transformation was broken.
2. Avoids the duplication of the exit subregion (the old transformation duplicates it)
3. Adds documentation to avoid confusion in the future.

Close #895

### DIFF
--- a/jlm/llvm/opt/ScalarEvolution.cpp
+++ b/jlm/llvm/opt/ScalarEvolution.cpp
@@ -3,15 +3,17 @@
  * See COPYING for terms of redistribution.
  */
 
+#include <jlm/llvm/ir/operators/GetElementPtr.hpp>
 #include <jlm/llvm/ir/operators/IntegerOperations.hpp>
+#include <jlm/llvm/ir/operators/IOBarrier.hpp>
+#include <jlm/llvm/ir/operators/Load.hpp>
+#include <jlm/llvm/ir/operators/sext.hpp>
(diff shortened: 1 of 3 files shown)


haved (dismissed):

haved:
@phate looks good, but HLS is unhappy

phate:
@haved I adjusted the golden cycles to appease HLS

PR #1528, opened 2026-02-18
Fix compilation of scalar evolution unit tests

phate:
My compiler refuses to compile when there is a `const` in front of the node reference.


haved:
@phate I'm not sure why but GitHub's UI says this PR has no additions and no deletions. Is the fix already in master?

PR #1532, opened 2026-02-18
ScalarEvolution: Implement loop trip count analysis

andreaslhjulstad:
This PR adds a first version of trip count analysis to the `ScalarEvolution` class. It allows for computing the number of times a theta node executes by using the computed chain recurrences from the SCEV analysis.

Currently, the trip count analysis is ran as part of the SCEV "transformation" pass and relies on the variables stored in the `Context` of the `ScalarEvolution` class, but this is a temporary solution. The method, `GetPredictedTripCount()` takes a theta node as input and returns a `TripCount` struct, which has a `Kind` attribute which can be either `Finite`, `Infinite` or `CouldNotCompute`. In the case where it is finite, the count can be retrieved using `GetCount()` on the `TripCount` object. An `Infinite` trip count symbolizes a loop which never terminates. 

The trip count is equal to the "backedge taken count" + 1, because each loop executes at least once. The backedge taken count can be reliably computed for recurrences that are **_affine_** (has a start value and a constant step) and **_quadratic_** (has two step increments). For chrecs with a size greater than this and for constant values, the implementation returns "could not compute". It also handles "trivial" cases where the backedge is not taken or the loop runs forever.

### DIFF
--- a/jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
+++ b/jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
@@ -324,14 +324,8 @@ RvsdgToIpGraphConverter::ConvertGammaNode(const rvsdg::GammaNode & gammaNode)
 }
 
 bool
-RvsdgToIpGraphConverter::RequiresSsaPhiOperation(
-    const rvsdg::ThetaNode::LoopVar & loopVar,
-    const Variable & v)
+RvsdgToIpGraphConverter::RequiresSsaPhiOperation(const rvsdg::ThetaNode::LoopVar & loopVar)
 {
-  // FIXME: solely decide on the input instead of using the variable
-  if (is<GlobalVariable>(&v))
-    return false;
-
   if (ThetaLoopVarIsInvariant(loopVar))
     return false;
 
@@ -357,7 +351,7 @@ RvsdgToIpGraphConverter::ConvertThetaNode(const rvsdg::ThetaNode & thetaNode)
   for (const auto & loopVar : thetaNode.GetLoopVars())
   {
     auto variable = Context_->GetVariable(loopVar.input->origin());
-    if (RequiresSsaPhiOperation(loopVar, *variable))
+    if (RequiresSsaPhiOperation(loopVar))
     {
       auto phi = entryBlock->append_last(SsaPhiOperation::create({}, loopVar.pre->Type()));
       phis.push_back(phi);
@@ -372,21 +366,17 @@ RvsdgToIpGraphConverter::ConvertThetaNode(const rvsdg::ThetaNode & thetaNode)
   size_t phiIndex = 0;
   for (const auto & loopVar : thetaNode.GetLoopVars())
   {
-    auto entryVariable = Context_->GetVariable(loopVar.input->origin());
-    if (RequiresSsaPhiOperation(loopVar, *entryVariable))
+    auto resultVariable = Context_->GetVariable(loopVar.post->origin());
+    Context_->InsertVariable(loopVar.output, resultVariable);
+    if (RequiresSsaPhiOperation(loopVar))
     {
-      auto resultVariable = Context_->GetVariable(loopVar.post->origin());
+      auto entryVariable = Context_->GetVariable(loopVar.input->origin());
       const auto phi = phis[phiIndex++];
       phi->replace(
           SsaPhiOperation(
               { preEntryBlock, Context_->GetLastProcessedBasicBlock() },
               resultVariable->Type()),
           { entryVariable, resultVariable });
-      Context_->InsertVariable(loopVar.output, resultVariable);
-    }
-    else
-    {
-      Context_->InsertVariable(loopVar.output, Context_->GetVariable(loopVar.post->origin()));
     }
   }
   JLM_ASSERT(phiIndex == phis.size());
--- a/jlm/llvm/backend/RvsdgToIpGraphConverter.hpp
+++ b/jlm/llvm/backend/RvsdgToIpGraphConverter.hpp
@@ -102,7 +102,7 @@ private:
   CreateInitialization(const rvsdg::DeltaNode & deltaNode);
 
   static bool
-  RequiresSsaPhiOperation(const rvsdg::ThetaNode::LoopVar & loopVar, const Variable & v);
+  RequiresSsaPhiOperation(const rvsdg::ThetaNode::LoopVar & loopVar);
 
   std::unique_ptr<Context> Context_;
 };


andreaslhjulstad:
> @andreaslhjulstad you say that early exits are not supported, what does this mean in practice? RVSSG doesn't support breaking or returning inside loops, so I would think the trip count you provide is actually always the precise number of iterations?

Oh I did not know that. In that case, it will be the precise number yes. Updated the description!

jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -401,6 +401,40 @@ class SCEVChainRecurrence final : public SCEVNAryExpr
...
+    if (Operands_.size() < 2)
+    {
+      return nullptr;
+    }
+    if (Operands_.size() == 2)
+    {
+      return SCEV::CloneAs<SCEVConstant>(*Operands_[1]);
+    }
+    auto newRec = SCEVChainRecurrence::Create(*Loop_);
+    for (size_t i = 1; i < Operands_.size(); i++)
```
phate: `for(auto & operand : Operands_)`

jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -401,6 +401,40 @@ class SCEVChainRecurrence final : public SCEVNAryExpr
...
+    return chrec.GetOperands().size() == 2;
+  }
+
+  bool static IsQuadratic(const SCEVChainRecurrence & chrec)
+  {
+    return chrec.GetOperands().size() == 3;
+  }
+
+  std::unique_ptr<SCEV>
+  GetStep() const
```
phate: Personally, I find the interface confusing. It returns a unique pointer, but that can be a nullptr. I would rather prefer this to be wrapped in an `std::optional<>`, or at least provide some documentation.

jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -566,6 +600,62 @@ class ScalarEvolution final : public jlm::rvsdg::Transformation
...
+  {
+    enum class Kind
+    {
+      Finite,
+      Infinite,
+      CouldNotCompute
+    };
+
+    static TripCount
+    Finite(const size_t n)
```
phate: Can you prefix them with `create`, i.e., `createFinite()`, `createInfinite()`, etc.

Moreover, what is the difference between infinite and could-not-compute?

jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -566,6 +600,62 @@ class ScalarEvolution final : public jlm::rvsdg::Transformation
...
+    bool
+    IsCouldNotCompute() const
+    {
+      return kind == Kind::CouldNotCompute;
+    }
+
+    size_t
+    GetCount() const
+    {
+      assert(IsFinite() && "TripCount is not finite");
```
phate: JLM_ASSERT()

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -156,6 +176,22 @@ class ScalarEvolution::Statistics final : public util::Statistics
...
+    bool first = true;
+    for (auto & [thetaNode, tripCount] : context.GetTripCountMap())
+    {
+      if (!first)
+        tripCounts += ',';
+      first = false;
+
+      const std::string count = tripCount.IsFinite()   ? std::to_string(tripCount.GetCount())
+                              : tripCount.IsInfinite() ? "Infinite"
+                                                       : "CouldNotCompute";
```
phate: Please program a bit more defensively. You only have three options now, but might have more in the future. Thus, make the else case a `std::logic_error`

Also you might want its own function for the string-count-to-string conversion.

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -223,11 +265,463 @@ ScalarEvolution::AnalyzeRegion(const rvsdg::Region & region)
...
+      rvsdg::TryGetSimpleNodeAndOptionalOp<rvsdg::MatchOperation>(*pred->origin());
+  if (!matchOperation)
+    throw std::logic_error("Predicate is not connected to a match node!");
+
+  JLM_ASSERT(node->ninputs() == 1); // Match node only has 1 input
+
+  const auto origin = node->input(0)->origin();
+  const auto comparisonNode = rvsdg::TryGetOwnerNode<rvsdg::SimpleNode>(*origin);
+  if (!comparisonNode)
+    throw std::logic_error("Match node is not connected to a simple node!");
```
phate: same here

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -223,11 +265,463 @@ ScalarEvolution::AnalyzeRegion(const rvsdg::Region & region)
...
+  if (!(dynamic_cast<const IntegerSltOperation *>(comparisonOperation)
+        || dynamic_cast<const IntegerSleOperation *>(comparisonOperation)
+        || dynamic_cast<const IntegerUltOperation *>(comparisonOperation)
+        || dynamic_cast<const IntegerUleOperation *>(comparisonOperation)
+        || dynamic_cast<const IntegerSgtOperation *>(comparisonOperation)
+        || dynamic_cast<const IntegerSgeOperation *>(comparisonOperation)
+        || dynamic_cast<const IntegerUgtOperation *>(comparisonOperation)
+        || dynamic_cast<const IntegerUgeOperation *>(comparisonOperation)
+        || dynamic_cast<const IntegerNeOperation *>(comparisonOperation)
+        || dynamic_cast<const IntegerEqOperation *>(comparisonOperation)))
```
phate: You might want to use the `is<>()` method here

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -223,11 +265,463 @@ ScalarEvolution::AnalyzeRegion(const rvsdg::Region & region)
...
+      const auto backedgeTakenCount =
+          ComputeBackedgeTakenCountForChrec(*chrec, bound, comparisonOperation);
+      if (backedgeTakenCount.has_value())
+      {
+        return TripCount::Finite(*backedgeTakenCount + 1);
+      }
+    }
+  }
+  if (dynamic_cast<const IntegerSgeOperation *>(comparisonOperation)
+      || dynamic_cast<const IntegerUgeOperation *>(comparisonOperation))
```
phate: `is<>()`

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -223,11 +265,463 @@ ScalarEvolution::AnalyzeRegion(const rvsdg::Region & region)
...
+            ComputeBackedgeTakenCountForChrec(*chrec, bound, comparisonOperation);
+        if (start >= bound && (bound - start) % step == 0)
+          return TripCount::Finite(*backedgeTakenCount + 1);
+      }
+    }
+    if (start == bound)
+      return TripCount::Finite(1);
+  }
+
+  if (dynamic_cast<const IntegerEqOperation *>(comparisonOperation))
```
phate: `is<>()`

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -223,11 +265,463 @@ ScalarEvolution::AnalyzeRegion(const rvsdg::Region & region)
...
+      if (!StepAlwaysZero(*stepSCEV))
+        return TripCount::Finite(2); // Backedge taken once
+    }
+    else
+      return TripCount::Finite(1);
+  }
+
+  if (SCEVChainRecurrence::IsQuadratic(*chrec))
+  {
+    // For quadratic recurrences, the value could evolve in an unpredictable way. In these cases,
```
phate: `an unpredictable way` is a bit of strong formulation here.

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -223,11 +265,463 @@ ScalarEvolution::AnalyzeRegion(const rvsdg::Region & region)
...
+  // f(i) - k changes sign, where f(i) is the value of the recurrence at iteration i and k is the
+  // bound value
+
+  const auto start = dynamic_cast<const SCEVConstant *>(chrec.GetStartValue())->GetValue();
+  const auto stepSCEV = chrec.GetStep();
+
+  bool isEqualsComparison = dynamic_cast<const IntegerSleOperation *>(comparisonOperation)
+                         || dynamic_cast<const IntegerUleOperation *>(comparisonOperation)
+                         || dynamic_cast<const IntegerSgeOperation *>(comparisonOperation)
+                         || dynamic_cast<const IntegerUgeOperation *>(comparisonOperation);
```
phate: `is<>()`

phate (changes requested):

PR #1533, opened 2026-02-18
InvariantValueRedirection: redirect load memory states

phate:
If there are no users of the loaded value of a load, the nodes' memory states can be redirect to their respective inputs' users.


phate:
@haved I am unsure whether this is already sufficient to cover all cases or whether we still need something in DNE. Let's start with this and see how it goes.

PR #1540, opened 2026-02-25
DNE: Eliminate dead lode nodes

phate:
This PR enables dead node elimination to recognize load nodes where only the loaded value is dead (and not the memory state edges) and remove them.


haved (approved):

haved:
@phate I was able to review

PR #1542, opened 2026-02-26
ScalarEvolution: Extend multiplication support

andreaslhjulstad:
This PR extends and refactors the multiplication folding of two chain recurrences in the `ScalarEvolution` implementation. Previously, we only handled multiplication of two chain recurrences when their size was >= 2. In other cases we would just return Unknown. Now, we have a general method which supports multiplication of two chrecs of arbitrary size, similar to the LLVM implementation. It is effectively an implementation of the algorithm `CRProd` detailed by Bachmann et al. in "Chains of recurrences—a method to expedite the evaluation of closed-form functions" (1994) (https://doi.org/10.1145/190347.190423).

It also does some refactoring by moving a lot of the logic out from the `ApplyMulFolding()` function and into a function `ComputeProductOfChrecs()`, which is used for handling all cases.

### DIFF
--- a/jlm/llvm/opt/ScalarEvolution.cpp
+++ b/jlm/llvm/opt/ScalarEvolution.cpp
@@ -1908,12 +1908,13 @@ ScalarEvolution::CanCreateChainRecurrence(
       return false;
     }
 
-  // Check that it has no reference via a mult-operation
-  for (auto [out, dependencyInfo] : deps)
+  if (rvsdg::TryGetRegionParentNode<rvsdg::ThetaNode>(output))
   {
-    if (dependencyInfo.operation == DependencyOp::Mul)
+    // If this is the output of a loop variable, check that it has no reference via a mult-operation
+    for (auto [out, dependencyInfo] : deps)
     {
-      return false;
+      if (dependencyInfo.operation == DependencyOp::Mul)
+        return false;
     }
   }
 


jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -377,6 +377,12 @@ class SCEVNAryExpr : public SCEV
     return Operands_.at(index).get();
   }
 
+  size_t
+  Size() const
```
phate: I find `Size()` as name for the number of operands extremely misleading. What about `numOperands()`?

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -1548,6 +1548,135 @@ ScalarEvolution::ApplyAddFolding(const SCEV * lhsOperand, const SCEV * rhsOperan
...
+
+  if (rhsSize > lhsSize)
+    std::swap(lhsChrec, rhsChrec);
+
+  if (rhsSize == 1)
+  {
+    const auto element = rhsChrec->GetOperand(0);
+
+    return SCEV::CloneAs<SCEVChainRecurrence>(*ApplyMulFolding(lhsChrec, element));
+  }
```
phate: The `rhsSize == 0` case was already checked above.

PR #1543, opened 2026-02-26
ScalarEvolution: Support for Store GEPs and other small improvements

andreaslhjulstad:
This PR improves the ScalarEvolution implementation by allowing us to handle Store operations in the same way that Loads are handled. We simply traverse through it to it's parent node and use SCEVUnknown to represent the operation itself. Since the only output from a Store operation is a state edge, we can no longer naively filter out loop variables that are states when creating SCEVs. Therefore we leave state edges in, but filter out the ones that do not change in the loop at all, by comparing the `post` value with the `pre` value.

It also simplifies the way of creating a chain recurrence by adding a constructor that takes a parameter pack of operands. Which means that the operands don't have to be manually added using AddOperand() when creating a new chrec. Finally, it refactors the `GetOrCreateStepForSCEV()` method by removing the check for an existing step (was never true), and changing to return early instead of returning at the end.

### DIFF
--- a/jlm/llvm/opt/DeadNodeElimination.cpp
+++ b/jlm/llvm/opt/DeadNodeElimination.cpp
@@ -36,46 +36,36 @@ namespace jlm::llvm
 class DeadNodeElimination::Context final
 {
 public:
-  void
+  /**
+   * Marks the given \p output as alive.
+   * @return true if the output was previously not marked, false otherwise.
+   */
+  bool
   markAlive(const rvsdg::Output & output)
   {
-    if (const auto simpleNode = rvsdg::TryGetOwnerNode<rvsdg::SimpleNode>(output))
-    {
-      SimpleNodes_.insert(simpleNode);
-      return;
-    }
+    return Outputs_.insert(&output);
+  }
 
-    Outputs_.insert(&output);
+  /**
+   * Marks the given \p simpleNode as alive.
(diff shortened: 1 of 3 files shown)


jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -828,7 +856,11 @@ ScalarEvolution::PerformSCEVAnalysis(const rvsdg::ThetaNode & thetaNode)
   std::vector<rvsdg::ThetaNode::LoopVar> nonStateLoopVars;
   for (const auto loopVar : thetaNode.GetLoopVars())
   {
-    if (loopVar.pre->Type()->Kind() != rvsdg::TypeKind::State)
+    // In some cases (e.g. with store operations), we still want to create a SCEV tree for the loop
+    // variable even though it is a state variable. However, we still want to filter out state
+    // variables that are purely for scaffolding as they are uninteresting for the analysis.
+    if (!(loopVar.pre->Type()->Kind() == rvsdg::TypeKind::State
+          && loopVar.post->origin() == loopVar.pre)) // Not used in any operations within the loop
```
phate: Instead of `loopVar.post->origin() == loopVar.pre`, use `ThetaLoopVarIsInvariant()`, please.

andreaslhjulstad: Didn't know we had that. Fixed!

PR #1545, opened 2026-03-03
StoreValueForwarding: Add routing out of gammas and thetas

haved:
For each load node, tracing is performed, storing metadata in a `LoadTracingInfo` struct.
The most important piece of metadata created during tracing is:
 > what is the last value stored to the relevant memory right before the given node, or right before the exit of the given region.

If the last value stored depends on which branch execution came from, the last store becomes the branch "join point" instead. A join point is either the exit of a gamma, the entry of a theta, or the exit of a theta. (The exit of a theta is arguably not a join, but we need tracing to stop there in order to create loop variables for routing values out of the theta body).

For each load, two stages are performed. The first stage only does tracing. Only if tracing was successful (all branches lead to a MustAlias store that can be forwarded), is the second stage performed. In the second stage, outputs are routed from store(s) to the load users.

To make the code simple to reason about, loop back-edges are never followed during tracing, neither in stage 1 nor 2. When tracing back in time reaches a loop variable pre, the recursive tracing function only continues through to the loop variable input. The loop variable post is added to a queue/set to be handled later. This ensures that all loop posts are handled eventually, without risking recursive functions visiting the same node multiple times at once, at different depths of the call stack, which was the source of a ton of confusion.


jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -135,154 +144,444 @@ StoreValueForwarding::traverseIntraProceduralRegion(rvsdg::Region & region)
...
 }
 
-void
-StoreValueForwarding::processLoadNode(rvsdg::SimpleNode & loadNode)
+// Enum containing the possible relationships between a load operation and a store node
+enum class StoreNodeInfo
 {
-  context_->numTotalLoads++;
+  ValueForwarding,  // The store can be forwarded to the load node
+  ClobberNoForward, // The store may clobber the loaded value, but can not be forwarded
```
phate: I fail to understand what this value stands for.

phate: I understand what this means now after reading the code. Maybe you could elaborate here where this scenario could arise (alias analysis etc.)

phate (changes requested):

PR #1546, opened 2026-03-03
Operation rework prep

caleridas:
Prepare to rework the operations classes: separate "SimpleOperation" and 
"StructuralOperation" into completely disjoint inheritance hierarchies. To make this
work, fix up all users that want to inspect operations so they forcibly obtain a 
SimpleNode first.

This is part of a sequence to clean up all structural nodes and provide better ways of
customizing them / applying attributes. It also ensures that there is one (and only one!)
path to check for specific node semantics:
- to check for "high-level node type", match it against SimpleNode, LambdaNode, GammaNode, ...
- to discriminate SimpleNode semantics further, check against specific operation classes

Whereever appropriate, change code to use MatchType and its various varieties.

### DIFF
--- a/tests/jlm/mlir/TestIntegerOperationsJlmToMlirToJlm.cpp
+++ b/tests/jlm/mlir/TestIntegerOperationsJlmToMlirToJlm.cpp
@@ -95,21 +95,24 @@ TestIntegerBinaryOperation()
       bool foundBinaryOp = false;
       for (auto & node : region->Nodes())
       {
-        auto convertedBinaryOp = dynamic_cast<const JlmOperation *>(&node.GetOperation());
-        if (convertedBinaryOp)
+        if (auto simplenode = dynamic_cast<const jlm::rvsdg::SimpleNode *>(&node))
         {
-          EXPECT_EQ(convertedBinaryOp->nresults(), 1);
-          EXPECT_EQ(convertedBinaryOp->narguments(), 2);
-          auto inputBitType1 = jlm::util::assertedCast<const jlm::rvsdg::BitType>(
-              convertedBinaryOp->argument(0).get());
-          EXPECT_EQ(inputBitType1->nbits(), nbits);
-          auto inputBitType2 = jlm::util::assertedCast<const jlm::rvsdg::BitType>(
-              convertedBinaryOp->argument(1).get());
-          EXPECT_EQ(inputBitType2->nbits(), nbits);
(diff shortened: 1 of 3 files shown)


jlm/hls/opt/cne.cpp
```diff
@@ -273,8 +273,11 @@ congruent(Output * o1, Output * o2, VisitorSet & vs, Context & ctx)
     }
   }
 
-  if (jlm::rvsdg::is<SimpleOperation>(n1) && jlm::rvsdg::is<SimpleOperation>(n2)
-      && n1->GetOperation() == n2->GetOperation() && n1->ninputs() == n2->ninputs()
+  auto s1 = dynamic_cast<const rvsdg::SimpleNode *>(n1);
+  auto s2 = dynamic_cast<const rvsdg::SimpleNode *>(n2);
+  if (s1 && s2 && jlm::rvsdg::is<SimpleOperation>(s1->GetOperation())
+      && jlm::rvsdg::is<SimpleOperation>(s2->GetOperation())
```
phate: You can remove the `jlm::rvsdg::is<SimpleOperation>(s1->GetOperation())` and `jlm::rvsdg::is<SimpleOperation>(s2->GetOperation())` as you already checked before that s1 and s2 are simple nodes.

jlm/llvm/ir/operators/MemCpy.hpp
```diff
@@ -68,7 +68,7 @@ class MemCpyOperation : public rvsdg::SimpleOperation
   [[nodiscard]] static rvsdg::Input &
   destinationInput(const rvsdg::Node & node) noexcept
   {
-    JLM_ASSERT(is<MemCpyOperation>(&node));
+    JLM_ASSERT(is<MemCpyOperation>(static_cast<const rvsdg::SimpleNode &>(node).GetOperation()));
```
phate: assertedCast in this file

jlm/llvm/ir/operators/operators.hpp
```diff
@@ -2533,7 +2533,7 @@ class FreeOperation final : public rvsdg::SimpleOperation
   [[nodiscard]] static rvsdg::Input &
   addressInput(const rvsdg::Node & node) noexcept
   {
-    JLM_ASSERT(is<FreeOperation>(&node));
+    JLM_ASSERT(is<FreeOperation>(static_cast<const rvsdg::SimpleNode &>(node).GetOperation()));
```
phate: assertedCast

PR #1547, opened 2026-03-04
Add toDot() and view() methods to ControlFlowGraph class

phate:


haved (dismissed):
Looks good, the only thing I'm missing is a docstring explaining that view() actually launches an external program, and is intended for use in debuggers. The fact that it is return type `void` kind of gives that away, however, so not a big deal

PR #1559, opened 2026-03-09
LoopStrengthReduction: Add first version of LSR implementation

andreaslhjulstad:
This PR adds a first implementation of the loop strength reduction pass. It traverses the theta node bottom up and finds candidate operations. These are statements that are linear combinations of loop variables and constants that involve multiplication in some way. 

A candidate operation must satisfy three conditions:
1. The statement is a linear combinations of loop variables and constants
2. Its RVSDG subtree contains at least one multiplication node (IntegerMulOperation) ensuring we only reduce operations that actually benefit from strength reduction.
3. Its SCEV evaluates to a chain recurrence of the form {a,+,b}, meaning the value can be expressed as a start value plus a constant step per iteration. This is what makes it a valid induction variable candidate.

We add the `Output` of each candidate operation node to a set. After the traversal is finished, we replace the operations with a simple addition using the following steps:

For a variable **_j_** defined as a linear combination of loop variables and constants with chain recurrence **{a,+,b}** where **_b_** can also be 0 (invariant):
1. Add a new loop variable **_s_** with `input` value **_a_**
4. Divert all the users of **_j_** to use the `pre` value of **_s_**

If **_b_** is _not_ zero:

5. Create a new constant operation with value **_b_**
6. Insert an add-node which takes as inputs the `pre` value of **_s_** and the constant **_b_**
7. Set the `post` value of **_s_** to the output of the new add-node
 
Dangling nodes are cleaned up at the end using `prune()` on each region we process.

Loop strength reduction achieves three things:

1. It converts multiplication operations, that may be costly on some systems, into cheaper additions. 
2. In the case of **_b_** being 0, which means that the value doesn't change within the loop, we essentially do loop-invariant code motion by replacing the operation with a constant in the "pre-header" of the loop. 
3. It makes it possible for us to perform induction variable elimination, where we can remove loop variables that (after strength reduction) are only used in loop tests by rewriting the predicate to use the newly added induction variable. 

As an added bonus, LSR also makes it easier to perform copy propagation and dead code elimination, allowing for further optimizations down the line.


jlm/llvm/opt/LoopStrengthReduction.cpp
```diff
@@ -0,0 +1,346 @@
...
+#include <jlm/llvm/opt/ScalarEvolution.hpp>
+#include <jlm/rvsdg/RvsdgModule.hpp>
+#include <jlm/util/Statistics.hpp>
+
+namespace jlm::llvm
+{
+class LoopStrengthReduction::Context final
+{
+public:
+  ~Context() = default;
```
phate: `noexcept`

phate (changes requested):
The only reason why I did not approve is because of the intermingling of testing whether the transformation is possible and performing the transformation. All other comments are minor stuff.

PR #1594, opened 2026-04-09
LSR: Generalize reduction of arithmetic operations

andreaslhjulstad:
This PR extends the existing LSR transformation pass by adding functionality for reducing arithmetic operations that are represented by chain recurrences consisting of elements that are chain recurrences, `SCEVNAryExpr`'s and `SCEVInit` values, and not just constants. In short, the new change generalizes the transformation so that any operation that contains a multiplication, can be represented as an affine recurrence, and is dependent on an induction variable, can be reduced. 

The way this was implemented was by generalizing the logic of hoisting constants out of a loop by creating a corresponding constant in the loop's preheader to work for arbitrary SCEV expressions. Also generalizes this to work over multiple layers of nested loops by adding methods to route a value down from an outer loop to an inner loop as well as tracing a value from an inner loop to an outer loop.

In order to achieve this, some changes also had to be made to the Scalar Evolution analysis. Most notable is the fact that `SCEVChainRecurrence`s now store the output of the value they represent with an `Output_` field in the class. This was done in order to retrieve this information in the LSR analysis later to check if the recurrence exists as an induction variable or if we need to create a new one for it, which can happen in cases where we have folded some constant into the chrec of an induction variable, either with addition or multiplication.

### DIFF
--- a/jlm/llvm/opt/LoopStrengthReduction.cpp
+++ b/jlm/llvm/opt/LoopStrengthReduction.cpp
@@ -299,8 +299,24 @@ LoopStrengthReduction::FindLoopPath(const rvsdg::ThetaNode & from, const rvsdg::
   return reversed;
 }
 
+bool
+LoopStrengthReduction::IsAncestorRegion(
+    const rvsdg::Region & candidateAncestor,
+    const rvsdg::Region & region)
+{
+  if (region.IsRootRegion())
+    return false;
+
+  const auto parentRegion = region.node()->region();
+
+  if (&candidateAncestor == parentRegion)
+    return true;
+
+  return IsAncestorRegion(candidateAncestor, *parentRegion);
+}
+
 std::optional<rvsdg::Output *>
-LoopStrengthReduction::RouteValueThroughLoops(
+LoopStrengthReduction::TryRouteValueThroughLoops(
     rvsdg::Output & origin,
     const rvsdg::ThetaNode & from,
     const rvsdg::ThetaNode & to)
@@ -364,8 +380,47 @@ LoopStrengthReduction::HoistChrec(
   }
 
   const auto routed = RouteValueThroughLoops(*chrecOutput, *chrec.GetLoop(), thetaNode);
+  if (IsAncestorRegion(*thetaNode.subregion(), *targetLoop->subregion()))
+  {
+    const auto traced = TryTraceValueUpwards(*chrecOutput, *thetaNode.subregion());
+    if (!traced.has_value())
+      return std::nullopt;
+
+    return *traced;
+  }
+
+  const auto routed = TryRouteValueThroughLoops(*chrecOutput, *targetLoop, thetaNode);
   if (routed.has_value())
     return *routed;
+
+  return std::nullopt;
+}
+
+std::optional<rvsdg::Output *>
+LoopStrengthReduction::TryTraceValueUpwards(rvsdg::Output & origin, rvsdg::Region & targetRegion)
+{
+  if (origin.region() == &targetRegion)
+  {
+    return &origin;
+  }
+
+  if (const auto theta = rvsdg::TryGetRegionParentNode<rvsdg::ThetaNode>(origin))
+  {
+    const auto loopVar = theta->MapPreLoopVar(origin);
+    auto traced = TryTraceValueUpwards(*loopVar.input->origin(), targetRegion);
+    if (traced.has_value())
+      return *traced;
+  }
+  else if (const auto gamma = rvsdg::TryGetRegionParentNode<rvsdg::GammaNode>(origin))
+  {
+    const auto roleVar = gamma->MapBranchArgument(origin);
+    if (const auto entryVar = std::get_if<rvsdg::GammaNode::EntryVar>(&roleVar))
+    {
+      auto traced = TryTraceValueUpwards(*entryVar->input->origin(), targetRegion);
+      if (traced.has_value())
+        return *traced;
+    }
+  }
 
   return std::nullopt;
 }
@@ -387,7 +442,7 @@ LoopStrengthReduction::HoistSCEVExpresssion(
   if (const auto init = dynamic_cast<const SCEVInit *>(&scev))
   {
     const auto prePointer = init->GetPrePointer();
-    const auto targetLoop = init->GetLoop();
+    const auto targetLoop = rvsdg::TryGetRegionParentNode<rvsdg::ThetaNode>(*prePointer);
     if (targetLoop == nullptr)
       return std::nullopt;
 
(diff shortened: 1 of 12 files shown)


jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -63,11 +63,11 @@ class SCEVUnknown final : public SCEV
 class SCEVInit final : public SCEV
 {
 public:
-  explicit SCEVInit(const rvsdg::Output & pre)
+  explicit SCEVInit(rvsdg::Output & pre)
       : PrePointer_{ &pre }
   {}
 
-  const rvsdg::Output *
+  rvsdg::Output *
```
phate: Return a reference here

phate: There can be a `[[nodiscard]]` before the return type

jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -63,11 +63,11 @@ class SCEVUnknown final : public SCEV
...
   GetPrePointer() const
```
phate: It can be annotated with `noexcept`

jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -88,23 +88,23 @@ class SCEVInit final : public SCEV
...
 class SCEVPlaceholder final : public SCEV
 {
 public:
-  explicit SCEVPlaceholder(const rvsdg::Output & pre)
+  explicit SCEVPlaceholder(rvsdg::Output & pre)
       : PrePointer_{ &pre }
   {}
 
-  const rvsdg::Output *
+  rvsdg::Output *
```
phate: Same here

jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -346,37 +346,73 @@ class SCEVNAryExpr : public SCEV
...
   }
 
+  /**
+   * Checks the operands of the given \p chrec to see if any of them are unknown.
+   *
+   * @param chrec the chain recurrence to be checked
+   * @return true if the recurrence contains an unknown, false otherwise
+   */
+  bool static IsUnknown(const SCEVNAryExpr & chrec)
+  {
```
phate: I think the entire function could be simplified if it would take a SCEVExpr as argument.

jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -346,37 +346,73 @@ class SCEVNAryExpr : public SCEV
...
+      rvsdg::Output & output,
+      Args &&... operands)
       : SCEVNAryExpr(std::forward<Args>(operands)...),
-        Loop_{ &theta }
+        Loop_{ &theta },
+        Output_{ &output }
   {}
 
-  const rvsdg::ThetaNode *
+  rvsdg::ThetaNode *
```
phate: reference

jlm/llvm/opt/ScalarEvolution.hpp
```diff
@@ -346,37 +346,73 @@ class SCEVNAryExpr : public SCEV
...
   GetLoop() const
   {
     return Loop_;
   }
 
+  rvsdg::Output *
```
phate: reference

jlm/llvm/opt/ScalarEvolution.cpp
```diff
@@ -284,17 +284,17 @@ ScalarEvolution::Run(
 };
 
 void
-ScalarEvolution::AnalyzeRegion(const rvsdg::Region & region)
+ScalarEvolution::AnalyzeRegion(rvsdg::Region & region)
 {
-  for (const auto & node : region.Nodes())
+  for (rvsdg::Node & node : region.Nodes())
```
phate: Why not auto?

jlm/llvm/opt/LoopStrengthReduction.cpp
```diff
@@ -262,6 +265,277 @@ LoopStrengthReduction::ReplaceCandidateOperation(
...
+      const auto loopVar = intermediateTheta->AddLoopVar(current);
+      current = loopVar.pre;
+    }
+    else if (const auto intermediateGamma = dynamic_cast<rvsdg::GammaNode *>(intermediateNode))
+    {
+      const auto [input, branchArgument] = intermediateGamma->AddEntryVar(current);
+      current = branchArgument[1];
+    }
+  }
+  return current;
```
phate: I think you can scrap the FindLoopPath() and TryRouteValueThroughLoops() function. Please have a look at [RouteToRegion()](https://github.com/phate/jlm/blob/426a302e2ce1ebaa76290f8596fcfea79a77896c/jlm/rvsdg/node.hpp#L540-L541). It should do what you are looking for.

jlm/llvm/opt/LoopStrengthReduction.cpp
```diff
@@ -262,6 +265,277 @@ LoopStrengthReduction::ReplaceCandidateOperation(
...
+    if (const auto entryVar = std::get_if<rvsdg::GammaNode::EntryVar>(&roleVar))
+    {
+      auto traced = TryTraceValueUpwards(*entryVar->input->origin(), targetRegion);
+      if (traced.has_value())
+        return *traced;
+    }
+  }
+
+  return std::nullopt;
+}
```
phate: Please have a look at rvsdg/Trace.hpp. It should contain a tracer that does this job already for you.

phate (commented):
I think quite a few things can be simplified as we already have support for them. See my comments, please.

phate (commented):
@andreaslhjulstad Could we split this PR in multiple PRs? There are quite a few "unrelated" changes that could be moved into their own PRs:

1. IsAncestor() could be moved to its own PR. Some tests would be nice on this one
2. The entire tracing story could be moved into their own PR.
3. The rest of the changes LSR changes

This would help us in terms of reviews and make it also a bit faster in turn around time as some bits are less controversial than others.

jlm/llvm/opt/LoopStrengthReduction.cpp
```diff
@@ -272,63 +437,105 @@ LoopStrengthReduction::ReplaceArithmeticOperation(
...
     // Chrec has the form {a,+,b} which is a basic induction variable
-    const auto & stepPtr = chrec->GetStep();
+    const auto & stepPtr = chrec.GetStep();
 
     JLM_ASSERT(stepPtr.has_value());
 
     const auto & stepSCEV = *stepPtr;
     const auto & stepConstant = dynamic_cast<const SCEVConstant *>(stepSCEV.get());
+    const auto & stepInit = dynamic_cast<const SCEVInit *>(stepSCEV.get());
+    const auto & stepNAryExpr = dynamic_cast<const SCEVNAryExpr *>(stepSCEV.get());
```
phate: They are only used in the if conditions, no? If so, inline them.

phate (dismissed):

phate (approved):

PR #1596, opened 2026-04-12
Add json output for graph dumps

haved:
Adds the option of outputting json from the `GraphWriter` API. The json should take up much less space than the existing dot output.

The [viewer](https://dot-tree-viz.vercel.app/) has been updated to support converting graphs from json into GraphViz, and does so on an ad-hoc basis.

With this implementation, there is some dupliction between the dot output in C++, and the json->dot conversion in TypeScript in the viewer. At some point in the future it would be nice to only have a single path, but that can be left for later.

Also, the ASCII output from the `GraphWriter` is never really used, so it could also probably be removed at some point.


jlm/util/GraphWriter.cpp
```diff
@@ -124,26 +128,39 @@ PrintStringAsHtmlAttributeName(std::ostream & out, std::string_view string)
 }
 
 /**
- * Returns a C string with the given amount of indentation.
- * The string is valid until this function is called again.
- * @param indent the number of indentation levels
- * @return a string of spaces corresponding to the indentation level.
+ * Prints the given \p string as a JSON string to the \p out stream.
+ * The result is always quoted. qoutes, backslashes and newlines get escaped.
```
phate: Typo: qoutes

phate (commented):
Looks all good to me. Great work! Any chance for some tests on top of it?

haved:
@phate I added tests of the Json output to the existing tests

phate (approved):

PR #1598, opened 2026-04-13
Add INSTALL variable to Makefile

mortkever:
Add a default value for INSTALL variable. No check is done whether INSTALL is set so this seems like a reasonable default. This should help the layman, while not bothering those who have it set to something custom.


Makefile
```diff
@@ -5,6 +5,8 @@ BUILD_OUT_PREFIX = build/
 
 default_target: all
 
+INSTALL ?= install
```
phate, line 8: @caleridas I have the feeling that this rather should be set in the Makefile.config than here.

caleridas: well the entire INSTALL infrastructure is a bit incomplete at present (pending my old change), so until I resolve that completely it might be reasonable to default it in the makefile

phate: All right, let's go with that then.

phate (approved):

PR #1600, opened 2026-04-14
Add toJson() methods to Region class

phate:


phate:
@haved This certainly ain't pretty, but it should do the job for now. The only thing left is then to switch to the json output in the RVSDG tree printer pass.

There is some refactoring necessary after this adventure.

haved (approved):

PR #1606, opened 2026-04-15
Trace directly from gamma branch argument to input

haved:
this change is purely an optimization, no functional difference.

It also adds vector size reserving to some of the other functions that actually return vectors.

### DIFF
--- a/jlm/rvsdg/gamma.cpp
+++ b/jlm/rvsdg/gamma.cpp
@@ -344,7 +344,9 @@ const rvsdg::Input &
 GammaNode::mapBranchArgumentToInput(const rvsdg::Output & output) const
 {
   JLM_ASSERT(rvsdg::TryGetRegionParentNode<GammaNode>(output) == this);
-  return *input(output.index());
+  auto & in = *input(output.index());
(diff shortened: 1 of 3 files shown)


jlm/rvsdg/gamma.cpp
```diff
@@ -338,6 +340,13 @@ GammaNode::MapBranchArgument(const rvsdg::Output & output) const
   }
 }
 
+const rvsdg::Input &
+GammaNode::mapBranchArgumentToInput(const rvsdg::Output & output) const
+{
+  JLM_ASSERT(rvsdg::TryGetRegionParentNode<GammaNode>(output) == this);
+  return *input(output.index());
```
phate: Please, could you change the last line to something like:

```
auto & i = input(output.index);
JLM_ASSERT(*i->Type() == *output.Type())
return i
```

I want some more rails here as grabbing the wrong input could lead to nasty bugs.

phate (changes requested):
Sorry for being pedantic here.

haved:
@phate I added your suggestion

phate (approved):

PR #1609, opened 2026-04-18
Add caching support to rvsdg::OutputTracer

phate:


jlm/llvm/ir/Trace.cpp
```diff
@@ -51,7 +53,8 @@ OutputTracer::traceStep(rvsdg::Output & output, bool mayLeaveRegion)
 rvsdg::Output &
 traceOutput(rvsdg::Output & output)
 {
-  OutputTracer tracer;
+  constexpr bool enableCaching = true;
```
haved: Why is caching enabled here, which is the helper function for doing one-off tracing of single outputs?

I agree it can help with structural nodes, which otherwise could be traced multiple times due to loops. It might very well be the case that caching should be enabled by default. 

But all other places you create tracers in this PR, caching is turned off. Why is this location the only one where caching is on?

jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -371,7 +371,8 @@ class LoadTracingInfo
   getLastStoreBeforeInputInternal(rvsdg::Input & input)
   {
     // Create a tracer, allow it to go through loads, as we only care about stores
-    llvm::OutputTracer tracer;
+    constexpr bool enableCaching = false;
```
haved, line 374: This is the tracer used in SVF, so why is caching false here? I thought this was our performance problem. Is it not?

jlm/rvsdg/Trace.cpp
```diff
@@ -237,18 +249,45 @@ OutputTracer::traceStep(Output & output, bool mayLeaveRegion)
...
+  OutputTracer tracer(enableCaching);
   tracer.setInterprocedural(false);
   return tracer.trace(output);
 }
 
 Output &
 traceOutput(Output & output)
 {
-  OutputTracer tracer;
+  constexpr bool enableCaching = false;
```
haved, line 289: This is the equivalent helper for doing one-off tracing in the RVSDG version of the tracer. Why is caching disabled here, and not in the LLVM version?

haved (commented):
The caching itself looks very nice, a good middle ground between caching everything and too little.

I am a bit confused about where you enable it, though. See comments @phate

jlm/llvm/ir/Trace.cpp
phate: That was a mistake. Set to false now.

phate:
@haved I set the enableCaching everywhere to false now. By default, caching should NOT be enabled and this PR should not change the behavior of the tracer for its use sides in any way. Sorry for the mistake! Please have a look again.

haved (approved):

PR #1611, opened 2026-04-19
IfConversion: Fix match operation handling

phate:
The transformation did not gracefully handle match operations with multiple alternatives mapping to the same subregion.


phate:
@haved This one needs to go in first before I put up the switch conversion improvement.

haved (approved):

PR #1612, opened 2026-04-19
LlvmModuleConversion: improve translation of switch statements

phate:
A switch statement was translated by creating an outgoing edge for every case of the switch. However, this did not take into account that cases can have duplicated target basic blocks. This PR changes the translation to only create a single outgoing edge for each target basic block instead of for each case. This should improve the resulting RVSDG structure as it avoids CFG restructuring for these switch statements as well as improving the runtime behavior of subsequent transformations.

### DIFF
--- a/jlm/llvm/Makefile.sub
+++ b/jlm/llvm/Makefile.sub
@@ -176,6 +176,7 @@ run-libllvm-tests_SOURCES = \
     jlm/llvm/frontend/ExportTests.cpp \
     jlm/llvm/frontend/FNegTests.cpp \
     jlm/llvm/frontend/FunctionCallTests.cpp \
+    jlm/llvm/frontend/LlvmModuleConversionTests.cpp \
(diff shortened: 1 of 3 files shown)


phate:
@haved Some examples from insn-attrb how it changed the RVSDG structure:

Before this patch:
1. insn_min_length function: 2 gammas in lambda subregion, one with 943 subregions, the other one with 31 subregions
2. internal_dfa_insn_code: 2 gammasin lambda subregion, one with 1686 subregions, the other one with 172 subregions
3. insn_default_latency: 2 gammas in lambda subregion, one with 1686 subregions, the other one with 165 subregions

After this patch:
1. insn_min_length function: 1 gamma in lambda subregion with 59 subregions
2. internal_dfa_insn_code: 1 gamma in lambda subregion with 284 subregions
3. insn_default_latency: 1 gamma in lambda subregion with 266 subregions

haved (approved):

PR #1617, opened 2026-04-20
Add method to check if a region is ancestor of another in region.cpp

andreaslhjulstad:


jlm/rvsdg/region.hpp
```diff
@@ -670,6 +670,17 @@ class Region
...
 
+  /**
+   * Determines if \p ancestor is a direct or indirect parent region of \p region
+   * in the RVSDG.
+   *
+   * @param ancestor The region to test as a potential ancestor.
+   * @param region The region to check the ancestry of.
+   * @return true if \p ancestor appears somewhere above \p region in the RVSDG, false otherwise.
+   */
+  [[nodiscard]] static bool
```
phate, line 681: The name should be `isAncestor()`

jlm/rvsdg/region.hpp
```diff
@@ -670,6 +670,17 @@ class Region
...
+  IsAncestor(const rvsdg::Region & ancestor, const rvsdg::Region & region) noexcept;
```
phate: My brain would have liked to turn the parameters around. First the region you want to test the ancestry from, then the ancestor region. The function is called `isAncestor()`, so I would expect the region to test the ancestry first.

phate (dismissed):

phate (approved):

PR #1618, opened 2026-04-20
Add optional withinRegion parameter to trace methods

andreaslhjulstad:
Replaces `mayLeaveRegion` to allow tracing to stop at a certain region.

### DIFF
--- a/jlm/rvsdg/Trace.cpp
+++ b/jlm/rvsdg/Trace.cpp
@@ -291,7 +291,8 @@ traceOutputIntraProcedurally(Output & output)
 Output &
 traceOutput(Output & output, rvsdg::Region * targetRegion)
 {
-  OutputTracer tracer;
+  constexpr bool enableCaching = false;
+  OutputTracer tracer(enableCaching);
   return tracer.trace(output, targetRegion);
 }
 
--- a/jlm/rvsdg/Trace.cpp
+++ b/jlm/rvsdg/Trace.cpp
@@ -21,17 +21,11 @@ OutputTracer::OutputTracer(const bool enableCaching) noexcept
 Output &
 OutputTracer::trace(Output & output)
 {
-  return trace(output, true, nullptr);
+  return trace(output, nullptr);
 }
 
 Output &
-OutputTracer::trace(Output & output, rvsdg::Region * targetRegion)
-{
-  return trace(output, true, targetRegion);
-}
-
-Output &
-OutputTracer::trace(Output & output, bool mayLeaveRegion, rvsdg::Region * targetRegion)
+OutputTracer::trace(Output & output, rvsdg::Region * withinRegion)
 {
   Output * head = &output;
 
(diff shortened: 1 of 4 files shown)


jlm/rvsdg/Trace.hpp
```diff
@@ -115,6 +115,13 @@ class OutputTracer
   [[nodiscard]] Output &
   trace(Output & output);
 
+  /**
+   * Traces from the given \p output to find the source of the output's value.
+   * The optional parameter \p withinRegion prevents values from being traced out of the region.
```
phate: Please clarify explicitly what happens when withinRegion is a nullptr.

jlm/rvsdg/Trace.hpp
```diff
@@ -248,6 +247,9 @@ traceOutputIntraProcedurally(const Output & output)
 Output &
 traceOutput(Output & output);
 
+Output &
+traceOutput(Output & output, rvsdg::Region * targetRegion);
+
```
phate: Could you not just overload the already present traceOutput() method with a default parameter:

```
Output &
traceOutput(Output & output, rvsdg::Region * targetRegion = nullptr)
```

Also, why is it called targetRegion here and withinRegion above?

jlm/rvsdg/Trace.cpp
```diff
@@ -140,11 +142,11 @@ OutputTracer::tryTraceThroughTheta(ThetaNode & thetaNode, Output & output)
 }
 
 Output &
-OutputTracer::traceStep(Output & output, bool mayLeaveRegion)
+OutputTracer::traceStep(Output & output, rvsdg::Region * withinRegion)
 {
-  if (!mayLeaveRegion && TryGetRegionParentNode<Node>(output))
+  if (withinRegion && withinRegion == TryGetOwnerRegion(output))
```
phate, line 147: Are you sure that TryGetOwnerRegion() is right method here?

You want to check that the output is an argument here, isnt it?

haved: This seems correct to me.

An `rvsdg::Output` has an owner region iff it is a `RegionArgument`. Our condition for stopping is that:
 - we have a region we are forced to stay inside
 - the output is an argument of said region

jlm/rvsdg/Trace.cpp
```diff
@@ -286,6 +288,14 @@ traceOutputIntraProcedurally(Output & output)
   return tracer.trace(output);
 }
 
+Output &
+traceOutput(Output & output, rvsdg::Region * targetRegion)
```
haved: Here it is called `targetRegion`, we should stick to `withinRegion`.

jlm/rvsdg/Trace.hpp
```diff
@@ -115,6 +115,13 @@ class OutputTracer
   [[nodiscard]] Output &
   trace(Output & output);
 
+  /**
+   * Traces from the given \p output to find the source of the output's value.
+   * The optional parameter \p withinRegion prevents values from being traced out of the region.
+   */
+  Output &
```
haved: [[nodiscard]] on this one as well

haved (approved):

PR #1620, opened 2026-04-20
ScalarEvolution: Fix a logic error in handling dependencies with multiplication

andreaslhjulstad:
Logic was wrong. Some invalid cases, like a = 2 * a + 1 would be counted as valid, and some valid cases like b = b + 2 * a would be rejected. Only the variables that depend on itself through multiplication should be rejected as these result in geometric sequences. This is now fixed in this PR.


phate (commented):
Please add a test to ensure that this cannot happen again.

haved (approved):

PR #1621, opened 2026-04-20
CommonNodeElimination: add handling of identical gamma exit variables

### LINKED ISSUE #1563: CommonNodeElimination does not eliminate identical gamma exit variables
In a situation like the one pictured, CNE is currently not able to detect that the two gamma exit variables are congruent:

Image

Ideally this would be detected, such that the diverting pictured in red could be performed, making one of the exit variables dead.

In the picture the two exit variables take their result values from exactly the same arguments, but the pass should also handle cases where the exit variables take their value from different outputs, as long as they are congruent.

The proposed logic comes in addition to the existing logic that removes exit variables if they are invariant (all subregions have result origins that are congruent).

FYI @phate @caleridas

haved:
Fixes #1563

### DIFF
--- a/jlm/llvm/opt/CommonNodeElimination.cpp
+++ b/jlm/llvm/opt/CommonNodeElimination.cpp
@@ -11,9 +11,13 @@
 #include <jlm/rvsdg/MatchType.hpp>
 #include <jlm/rvsdg/theta.hpp>
 #include <jlm/rvsdg/traverser.hpp>
+#include <jlm/util/common.hpp>
+#include <jlm/util/Hash.hpp>
 #include <jlm/util/Statistics.hpp>
 #include <jlm/util/time.hpp>
+
 #include <map>
+#include <unordered_map>
 
 namespace jlm::llvm
 {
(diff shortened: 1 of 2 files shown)


jlm/llvm/opt/CommonNodeElimination.cpp
```diff
@@ -716,33 +720,204 @@ tryGetGammaExitVarCongruenceSet(
...
+ * @param second the second exit variable
+ * @param context the context of the common node elimination pass
+ * @return true if the exit variables are congruent
+ *
+ * @pre the exit variables belong to the same gamma node
+ */
+[[nodiscard]] static bool
+areGammaExitVariablesCongruent(
+    const rvsdg::GammaNode::ExitVar & first,
+    const rvsdg::GammaNode::ExitVar & second,
```
phate, line 761: Both exit variables should always be from the same gamma, right? Maybe you can assert that?

phate (dismissed):

haved:
@phate I added the assert, need a new review

phate (approved):

PR #1628, opened 2026-04-24
Add freeze operation

haved:
Also includes a test in the module converter (the frontend part). We do not currently have a nice file to place roundtrip conversion tests in, but that could be made in the future

### DIFF
--- a/jlm/llvm/frontend/LlvmModuleConversion.cpp
+++ b/jlm/llvm/frontend/LlvmModuleConversion.cpp
@@ -1579,6 +1579,17 @@ convert_insertelement_instruction(::llvm::Instruction * i, tacsvector_t & tacs, 
 }
 
 static const Variable *
+convert_freeze_instruction(::llvm::Instruction * i, tacsvector_t & tacs, Context & ctx)
+{
+  JLM_ASSERT(i->getOpcode() == ::llvm::Instruction::Freeze);
+
+  auto operand = ConvertValue(i->getOperand(0), tacs, ctx);
+  tacs.push_back(FreezeOperation::create(*operand));
+
+  return tacs.back()->result(0);
+}
+
+static const Variable *
 convert(::llvm::UnaryOperator * unaryOperator, tacsvector_t & threeAddressCodeVector, Context & ctx)
 {
   JLM_ASSERT(unaryOperator->getOpcode() == ::llvm::Instruction::FNeg);
@@ -1755,6 +1766,8 @@ convertInstruction(
     return convert<::llvm::ShuffleVectorInst>(instruction, threeAddressCodes, context);
   case ::llvm::Instruction::InsertElement:
     return convert_insertelement_instruction(instruction, threeAddressCodes, context);
+  case ::llvm::Instruction::Freeze:
+    return convert_freeze_instruction(instruction, threeAddressCodes, context);
   default:
     throw std::runtime_error(util::strfmt(instruction->getOpcodeName(), " is not supported."));
   }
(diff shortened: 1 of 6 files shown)


jlm/llvm/ir/operators/operators.hpp
```diff
@@ -1119,6 +1119,52 @@ class PoisonValueOperation final : public rvsdg::SimpleOperation
...
+ */
+class FreezeOperation final : public rvsdg::SimpleOperation
+{
+public:
+  ~FreezeOperation() noexcept override;
+
+  explicit FreezeOperation(std::shared_ptr<const jlm::rvsdg::Type> type)
+      : rvsdg::SimpleOperation({ type }, { type })
+  {
+    JLM_ASSERT(type->Kind() == rvsdg::TypeKind::Value);
```
phate: Hmm, please use explicit exceptions and type checking to ensure it is a value type.

jlm/llvm/ir/operators/operators.hpp
```diff
@@ -1119,6 +1119,52 @@ class PoisonValueOperation final : public rvsdg::SimpleOperation
...
+  copy() const override;
+
+  const jlm::rvsdg::Type &
+  getType() const noexcept
+  {
+    return *result(0).get();
+  }
+
+  static std::unique_ptr<llvm::ThreeAddressCode>
+  create(const Variable & operand)
```
phate: Name it `createTac()`

jlm/llvm/ir/operators/operators.hpp
```diff
@@ -1119,6 +1119,52 @@ class PoisonValueOperation final : public rvsdg::SimpleOperation
...
+  {
+    auto operation = std::make_unique<FreezeOperation>(operand.Type());
+    return ThreeAddressCode::create(std::move(operation), { &operand });
+  }
+
+  static jlm::rvsdg::Output &
+  create(jlm::rvsdg::Output & operand)
```
phate: Make it return a Node & and call it `createNode()`

jlm/llvm/ir/operators/operators.hpp
phate: See some other operations in this file that does this already.

jlm/llvm/ir/operators/operators.hpp
```diff
@@ -1119,6 +1119,52 @@ class PoisonValueOperation final : public rvsdg::SimpleOperation
   }
 };
 
+/** \brief FreezeOperation class
+ *
+ * This operator is the Jlm equivalent of LLVM's freeze operation.
+ * It converts undef and poison values into arbitrary, but fixed, values.
+ * For all other inputs it is a no-op.
+ */
+class FreezeOperation final : public rvsdg::SimpleOperation
```
phate: I think this should inherit from rvsdg::UnaryOperation

jlm/llvm/frontend/LlvmModuleConversion.cpp
```diff
@@ -1578,6 +1578,17 @@ convert_insertelement_instruction(::llvm::Instruction * i, tacsvector_t & tacs,
   return tacs.back()->result(0);
 }
 
+static const Variable *
+convert_freeze_instruction(::llvm::Instruction * i, tacsvector_t & tacs, Context & ctx)
```
phate:
```
static const Variable *
convertFreezeInstruction(::llvmFreezeInst * instruction, .....)
```

phate (changes requested):

haved:
@phate I implemented your comments

jlm/llvm/ir/operators/operators.cpp
```diff
@@ -3,11 +3,13 @@
  * See COPYING for terms of redistribution.
  */
 
+#include "jlm/rvsdg/unary.hpp"
```
phate: include <>

phate (dismissed):
One minor comment

phate (approved):

PR #1629, opened 2026-04-26
LLVM: add insertvalue instruction support

phate:


phate:
@haved I will provide a test asap.

jlm/llvm/ir/operators/AggregateOperations.hpp
```diff
@@ -0,0 +1,151 @@
...
+        new InsertValueOperation(aggregateType, valueType, std::move(indices)));
+  }
+
+  static rvsdg::SimpleNode &
+  createNode(
+      rvsdg::Output & aggregateOperand,
+      rvsdg::Output & valueOperand,
+      std::vector<unsigned> indices)
+  {
+    checkOperandTypes(aggregateOperand.Type(), valueOperand.Type(), indices);
```
haved: This check is also performed inside the call to `create` on the next line. Should we avoid calling it twice?

jlm/llvm/ir/operators/AggregateOperations.hpp
```diff
@@ -0,0 +1,151 @@
...
+        { &aggregateOperand, &valueOperand });
+  }
+
+  static std::unique_ptr<ThreeAddressCode>
+  createTac(
+      const Variable & aggregateOperand,
+      const Variable & valueOperand,
+      std::vector<unsigned> indices)
+  {
+    checkOperandTypes(aggregateOperand.Type(), valueOperand.Type(), indices);
```
haved: Same here

haved (commented):
This looks very nice, my only comment so far is that we perform the type checks twice @phate

jlm/llvm/ir/operators/AggregateOperations.hpp
phate: Great catch. Should be fixed now.

jlm/llvm/ir/operators/AggregateOperations.hpp
phate: See above.

phate:
@haved I added a test

haved (approved):

PR #1630, opened 2026-04-27
Add CallingConv support to functions and calls

haved:
Only a couple of the calling conventions from LLVM have been included:
 - The C calling convention (default)
 - `fastcc` (Not the same as `__fastcall`, which only applies to i386)
 - `coldcc`
 - `tailcc` (Enables tail call optimizations)

These alternative conventions can be used by LLVM when compiling static functions that are only accessible from within the module.

In order to support calling conventions throughout the pipeline, they have been added to:
 - `LlvmGraphImport` (for function declarations)
 - `FunctionNode` (In the InterProceduralGraph)
 - `LlvmLambdaOperation`
 - `CallOperation`

In order to make the code change cleaner, some of these classes now have helper creation functions that use the default calling convention and empty attribute sets. The `LlvmGraphImport` now has separate create functions for global variables and functions.


jlm/llvm/ir/CallingConv.hpp
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright 2026 Håvard Krogstie <krogstie.havard@gmail.com>
+ * See COPYING for terms of redistribution.
+ */
+
+#ifndef JLM_LLVM_IR_CALLCONV_HPP
+#define JLM_LLVM_IR_CALLCONV_HPP
```
phate: Could we spell the file name, guard etc. out, i.e., CallingConvention?

jlm/llvm/ir/operators/call.hpp
```diff
@@ -252,10 +254,12 @@ class CallOperation final : public jlm::rvsdg::SimpleOperation
 
   explicit CallOperation(
       std::shared_ptr<const rvsdg::FunctionType> functionType,
+      CallingConv callingConv,
```
phate: same here, spelling out

jlm/llvm/ir/operators/call.hpp
```diff
@@ -270,6 +274,12 @@ class CallOperation final : public jlm::rvsdg::SimpleOperation
     return FunctionType_;
   }
 
+  [[nodiscard]] CallingConv
+  getCallingConv() const noexcept
```
phate: spelling out

phate (dismissed):

phate:
@haved I have mostly issues with the naming, but I won't block it because of that.

haved:
@phate I changed it to "CallingConvention" in types, fields and filenames

phate (dismissed):

phate (approved):

PR #1632, opened 2026-04-28
Add missing operations to Andersen's set building phase

haved:


jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -971,6 +1016,41 @@ Andersen::AnalyzeConstantAggregateZero(const rvsdg::SimpleNode & node)
   (void)Set_->CreateRegisterPointerObject(output);
 }
 
+void
+Andersen::AnalyzeInsertValue(const rvsdg::SimpleNode & node)
+{
+  JLM_ASSERT(is<InsertValueOperation>(node.GetOperation()));
+
+  // The InsertValue instruction looks like
+  //     result = insertvalue(input, element, index)
```
phate: It should be indices...as it can be more than one

jlm/llvm/opt/alias-analyses/Andersen.cpp
```diff
@@ -971,6 +1016,41 @@ Andersen::AnalyzeConstantAggregateZero(const rvsdg::SimpleNode & node)
...
+    const auto elementPO = Set_->GetRegisterPointerObject(element);
+    const auto resultPO = Set_->CreateRegisterPointerObject(result);
+    Constraints_->AddConstraint(SupersetConstraint(resultPO, inputRegisterPO));
+    Constraints_->AddConstraint(SupersetConstraint(resultPO, elementPO));
+  }
+  else
+  {
+    // The operation does not add any additional pointees, so map the result to the input
+    Set_->MapRegisterToExistingPointerObject(result, inputRegisterPO);
+  }
```
phate, line 1053: I think this code does not handle the case where it can be more than one index, no?

haved:
@phate I didn't catch that insertvalue can take multiple indices, so I added that in the comment.
The code is still correct, however, as the entire aggregate is represented by a single pointer object

phate (approved):

PR #1633, opened 2026-04-28
Add missing comparison predicates to PtrCmpOperation

haved:
It previously did not support signed comparisons.

I ended up creating a central bijective map between the LLVM version and our own. The upside of this is that it keeps everything in one place. The downside is that the file that describes our own operation still `#include`s the LLVM header, while this ideally should only happen in the frontend and backend. Open to feedback.

PS:
I also had to fix the use of `std::move(functionType)` in `LlvmGraphImport::createFunctionImport`, as C++ might evaluate function arguments in arbitrary order.


phate (dismissed):

haved:
@phate I also had to do the MLIR conversion

phate (approved):

PR #1634, opened 2026-04-28
Store store reduction fix

haved:
The reduction had an assert that would trigger on alignment mismatch, but that is no reason to fail.

Instead I replaced it with an actual test that ensures that the size of the second store actually overrides the first store.

I also pass the address inputs through `traceOutput` just in case we can remove some more stores.

(As a side note, we could make a much greater effort to remove dead stores, but it does not fit that well in the reduction interface)

### DIFF
--- a/jlm/llvm/ir/operators/Store.cpp
+++ b/jlm/llvm/ir/operators/Store.cpp
@@ -8,6 +8,8 @@
 #include <jlm/llvm/ir/operators/MemoryStateOperations.hpp>
 #include <jlm/llvm/ir/operators/Store.hpp>
 #include <jlm/llvm/ir/Trace.hpp>
+#include <jlm/llvm/ir/types.hpp>
+#include <jlm/rvsdg/simple-node.hpp>
 #include <jlm/util/HashSet.hpp>
 
 namespace jlm::llvm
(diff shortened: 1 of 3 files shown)


jlm/llvm/ir/operators/Store.cpp
```diff
@@ -56,31 +58,38 @@ is_store_mux_reducible(const std::vector<jlm::rvsdg::Output *> & operands)
...
     const std::vector<jlm::rvsdg::Output *> & operands)
 {
   JLM_ASSERT(operands.size() > 2);
 
-  const auto storeNode = rvsdg::TryGetOwnerNode<rvsdg::SimpleNode>(*operands[2]);
-  if (!is<StoreNonVolatileOperation>(storeNode))
+  // Try tracing a memory state edge to a previous store
+  const auto [store1Node, store1Op] =
+      rvsdg::TryGetSimpleNodeAndOptionalOp<StoreNonVolatileOperation>(*operands[2]);
+  if (!store1Node || !store1Op)
```
phate: No need to check stor1Node here, store1Op is sufficient.

phate (dismissed):

haved:
@phate I fixed your comment, need a new review

phate (approved):

PR #1635, opened 2026-04-28
Add list of LLVM intrinsics to strip in the frontend

haved:


jlm/llvm/frontend/LlvmModuleConversion.cpp
```diff
@@ -1326,12 +1327,38 @@ createCall(
...
 }
 
+/**
+ * Checks if the intrinsic with the given ID should be ignored in the frontend.
+ * Calls to ignored intrinsics become no-ops, and declarations of ignored intrisincs are skipped.
+ * @param intrinsicId the id of the llvm intrinsic
+ * @return true if the intrinsic should be ignored, false otherwise
+ */
+static bool
+isIntrinsicIgnored(::llvm::Intrinsic::ID intrinsicId)
```
phate: How about: `ignoreIntrinsic()`?

phate (dismissed):

jlm/llvm/frontend/LlvmModuleConversion.cpp
haved: @phate I think that sounds more like an action and not a check. What about `shouldIgnoreIntrinsic()`?

phate: `shouldIgnoreIntrinsic()` is great.

phate (approved):

PR #1639, opened 2026-04-30
Use hashmap for SVF forwarding cache

caleridas:
Reduces the number of L3 cache misses caused from this location by 2/3 for large graphs.


haved (approved):
Nice find! I'm always annoyed at std::pair not having a hash implementation by default, so I went with the easy solution without considering the effect it would have

caleridas:
@haved For more context: did some analysis using linux perf on "what is really happening, why SVF is slow" -- and the answer is unfortunately a bit disappointing: primarily it is cache misses on the data structures comprising the "large" graph and all the auxiliary data structures

meaning: the way forward is to restructure to reduce pointer chasing (would esp. mean "dissolving" inputs / outputs, can talk about that in detail)

this change by itself will be "unimpressive", it may gain 2-5% or so -- the rest is pointer chasing in other places

PR #1642, opened 2026-05-01
LLVM: Remove jlm/llvm/ir/operators.hpp

phate:


haved (dismissed):
Nice cleanup

sjalander (dismissed):

jlm/hls/opt/cne.cpp
```diff
@@ -14,6 +13,9 @@
 #include <jlm/util/Statistics.hpp>
 #include <jlm/util/time.hpp>
 
+#include <jlm/rvsdg/delta.hpp>
+#include <jlm/rvsdg/lambda.hpp>
+#include <jlm/rvsdg/Phi.hpp>
```
haved: These are grouped with the library header

haved (dismissed):
@phate I noticed one grouping that was off

jlm/hls/opt/cne.cpp
phate: Fixed

haved (approved):

PR #1644, opened 2026-05-01
LLVM: add MemSetNonVolatileOperation

phate:


phate:
@haved I still need to add tests, but maybe you could already have a look at the implementations whether they make sense to you.

jlm/llvm/ir/operators/StdLibIntrinsicOperations.hpp
```diff
@@ -333,6 +333,228 @@ class MemCpyVolatileOperation final : public MemCpyOperation
...
+    types.insert(types.end(), numMemoryStates, MemoryStateType::Create());
+    return types;
+  }
+
+  static std::vector<std::shared_ptr<const rvsdg::Type>>
+  createResultTypes(size_t numMemoryStates)
+  {
+    // The memset() standard C library call has as return type void*, but LLVM models the
+    // function call nevertheless as:
+    // call void @llvm.memset.p0.p0.i64(ptr align 4 %7, ptr align 4 %8, i64 %9, i1 false)
```
haved: Is this comment correct? Memset doesn't have a source, it has a value. It doesn't read any memory, it only writes, right?

jlm/llvm/opt/LoadChainSeparation.cpp
```diff
@@ -428,6 +428,16 @@ LoadChainSeparation::traceModRefChains(rvsdg::Output & startOutput, ModRefChainS
                 currentOutput =
                     MemCpyOperation::mapMemoryStateOutputToInput(*currentOutput).origin();
               },
+              [&](const MemSetOperation &)
+              {
+                // FIXME: We really would like to know here which memory state belongs to the source
+                // and which to the dst address. This would allow us to be more precise in the
```
haved: Here as well. Memset doesn't have a source address, it just writes a byte value?

haved (commented):
It all looks good, I just saw some comments that gave me pause, as it looks like the comments are about memcpy instead @phate

jlm/llvm/opt/LoadChainSeparation.cpp
phate: Removed the comment. It was not true for memset. It was a copy/paste mistake.

jlm/llvm/ir/operators/StdLibIntrinsicOperations.hpp
phate: Fixed

phate:
@haved Addressed your comments and added some tests. Anything else you want me to add?

haved (approved):
Looks good @phate

PR #1645, opened 2026-05-02
LLVM: account for all memmove intrinsics in LlvmModuleConversion

phate:


phate:
@haved I ran the change through the jlm-benchmark suite. It produced numbers and nothing seem to have failed. So I guess it does not occur in our benchmarks.

haved (approved):
@phate LLVM has a bajillion intrinsics, so it makes sense that some of them never occur

phate:
@haved Ah well, but they seem to occur in the LLVM test suite. I will shelve this PR, as the info that it does not occur in the benchmarks is sufficient. No need to handle it then explicitly (for now).

PR #1646, opened 2026-05-06
Add number of alias queries to SVF statistics

haved:


jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -292,9 +298,9 @@ struct StoreValueOrigin
 class LoadTracingInfo
 {
 public:
-  LoadTracingInfo(rvsdg::SimpleNode & loadNode, OutputTracer & tracer)
+  LoadTracingInfo(rvsdg::SimpleNode & loadNode, StoreValueForwarding::Context & context)
```
phate: I  think this should be done differently:

1. Introduce a per Load counter to LoadTrancingInfo that counts the number of alias analysis queries
2. Add this number after all tracing of the load is done to the counter in the context

This avoids also that the context is handed into the LoadTracingInfo, and maybe also this friend stuff.

phate (changes requested):

jlm/llvm/opt/StoreValueForwarding.cpp
haved: Yes good point

phate (approved):

PR #1647, opened 2026-05-06
LLVM: Improve MatchOperation debug string

phate:


jlm/rvsdg/control.cpp
```diff
@@ -161,12 +161,21 @@ MatchOperation::reduce_operand(unop_reduction_path_t path, jlm::rvsdg::Output *
...
+  constexpr size_t maxAlternatives = 3;
+  for (auto [input, subregionId] : mapping_)
+  {
+    if (n >= maxAlternatives)
+      break;
+    oss << input << " -> " << subregionId << ", ";
+    n++;
+  }
+  if (nalternatives() >= maxAlternatives)
+    oss << "..., ";
```
haved: This ... could be printed inside the for loop when breaking, to avoid repeated comparisons

haved (commented):
One suggestion @phate

jlm/rvsdg/control.cpp
phate: Fixed.

haved (approved):

PR #1649, opened 2026-05-10
LLVM: add AggregateAllocaSplitting transformation

phate:


phate:
@haved This one is now ready for review.

jlm/llvm/opt/AggregateAllocaSplitting.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+#ifndef JLM_LLVM_OPT_AGGREGATEALLOCASPLITTING_HPP
+#define JLM_LLVM_OPT_AGGREGATEALLOCASPLITTING_HPP
+
+#include <jlm/rvsdg/Transformation.hpp>
+
+namespace jlm::llvm
+{
+
+class AggregateAllocaSplitting final : public rvsdg::Transformation
```
haved, line 22: A comment explaining what the class does would be nice

jlm/llvm/opt/AggregateAllocaSplitting.cpp
```diff
@@ -0,0 +1,573 @@
...
+  size_t numSplitAggregateAllocaNodes = 0;
+};
+
+struct AggregateAllocaSplitting::AllocaTraceInfo
+{
+  explicit AllocaTraceInfo(rvsdg::SimpleNode & allocaNode)
+      : allocaNode(&allocaNode)
+  {}
+
+  rvsdg::SimpleNode * allocaNode = nullptr;
```
haved: Since this value is assigned in the constructor I don't think it should have a default value. It should never actually be null?

jlm/llvm/opt/AggregateAllocaSplitting.cpp
```diff
@@ -0,0 +1,573 @@
...
+      dynamic_cast<const AllocaOperation *>(&allocaNode.GetOperation());
+  JLM_ASSERT(allocaOperation && isSplitableType(*allocaOperation->allocatedType()));
+
+  auto & address = AllocaOperation::getPointerOutput(allocaNode);
+
+  bool isSplitable = true;
+  AllocaTraceInfo allocaTraceInfo(allocaNode);
+
+  util::HashSet<rvsdg::Output *> visited;
+  util::HashSet<rvsdg::Output *> toVisit{ &address };
```
haved: The tovisit HashSet could instead be a simple queue, with the HashSet instead storing "visited or queued". I often call that hash set "seen", but it is not a great name.

jlm/llvm/opt/AggregateAllocaSplitting.cpp
```diff
@@ -0,0 +1,573 @@
...
+            {
+              return false;
+            },
+            [&]()
+            {
+              throw std::logic_error(util::strfmt(
+                  "Unhandled owner region node type: ",
+                  userRegion->node()->DebugString()));
+              // Silence compiler
+              return false;
```
haved, line 189: Do we really need to return even after throwing? How would control flow continue after the throw? If it complains then by all means, I just don't understand why the compiler would give such a false positive

jlm/llvm/opt/AggregateAllocaSplitting.cpp
```diff
@@ -0,0 +1,573 @@
...
+AggregateAllocaSplitting::checkGetElementPtrUsers(const rvsdg::SimpleNode & gepNode)
+{
+  [[maybe_unused]] auto gepOperation =
+      dynamic_cast<const GetElementPtrOperation *>(&gepNode.GetOperation());
+  auto & address = *gepNode.output(0);
+
+  bool hasOnlyLoadsAndStores = true;
+
+  util::HashSet<rvsdg::Output *> visited;
+  util::HashSet<rvsdg::Output *> toVisit{ &address };
```
haved: Same here, one hash set and one queue

jlm/llvm/opt/AggregateAllocaSplitting.cpp
```diff
@@ -0,0 +1,573 @@
...
+          auto & consumerRegion = *allocaConsumer->region();
+          // FIXME: Introduce convenient functions
+          [[maybe_unused]] auto index0 =
+              tryGetConstantSignedInteger(*allocaConsumer->input(1)->origin()).value();
+          const auto index1 =
+              tryGetConstantSignedInteger(*allocaConsumer->input(2)->origin()).value();
+          JLM_ASSERT(index0 == 0);
+
+          auto elementAlloca = elementAllocaNodes[index1];
+          auto & routedAddress = rvsdg::RouteToRegion(
```
haved, line 525: We do not memorize routing to regions here, so in the future it might be nice to do that to avoid giving extra work for CNE. Not a big issue now

haved (dismissed):
Looks good. I have some comments you can look at it you'd like. The code for finding all users of the alloca and all users of the GEP is very similar, so it makes me wish for a "reverse tracer" in the future

jlm/llvm/opt/AggregateAllocaSplitting.cpp
phate: Yes, the compiler needs the `return false` there, otherwise type inference is not working correctly and it thinks I put a void-returning lambda there, leading to a compilation error.

jlm/llvm/opt/AggregateAllocaSplitting.cpp
phate: Added a FIXME

jlm/llvm/opt/AggregateAllocaSplitting.hpp
phate: Done

jlm/llvm/opt/AggregateAllocaSplitting.cpp
phate: Removed

phate:
> The code for finding all users of the alloca and all users of the GEP is very similar, so it makes me wish for a "reverse > tracer" in the future

Yes, it is on my TODO list, but I was too lazy to factor it out for this PR now.

jlm/llvm/opt/AggregateAllocaSplitting.cpp
phate: I fail to understand how this efficiently ensures that values are not added multiple times to `toVisit`? I need to check that a value is not already in the `toVisit` before adding the value to it, as it could be that a value is not in `visited` yet, but still in `toVisit` when I try to add a second instance to `toVisit` again. This means I need to "iterate" through `toVisit`, which I cannot do efficiently when it is a list.

jlm/llvm/opt/AggregateAllocaSplitting.cpp
phate: see above

phate:
@haved I replied to all your comments.

jlm/llvm/opt/AggregateAllocaSplitting.cpp
phate: I changed it now. Please have a look whether it is like you thought.

jlm/llvm/opt/AggregateAllocaSplitting.cpp
phate: same as above.

haved (approved):

PR #1653, opened 2026-05-11
Dump json instead of dot when doing --dumpRvsdgDotGraphs

haved:
The latest improvements in the viewer only apply to json graphs, and the files are also smaller, so it makes sense to prefer json output.


phate:
@haved I think we can also update the flag name. While it is a public interface, we have nothing really depending on it (it is a debug flag after all). Maybe we should avoid the format name in the flag: "dumpRvsdgGraphs", just to make it a bit more future proof.

If you really want to be on the safe side, you can introduce another flag and mark this as deprecated, but I think that would be a bit overkill here.

phate (approved):

PR #1654, opened 2026-05-11
Use shared_ptr for AliasAnalysis references

haved:
We previously used references to pass the PointsToGraph to the PtGAA, and to pass multiple AAs into the chained AA.

This made it quite annoying to build arbitrary alias analysis chains, since you had to carefully manage the lifetimes of the individual analyses and the PointsToGraph.

This PR changes to `shared_ptr` to store the chained alias analyses and the PointsToGraph. These objects are created very rarely, so the overhead of using shared pointers should be non-existent.

EDIT: I also added the ability for the PtGAA to not crash when seeing unknown registers. This was not intended for this PR, but the change makes sense freestanding as well. (SVF creates new outputs during routing, which is why we need it)

### DIFF
--- a/jlm/llvm/opt/alias-analyses/PointsToGraph.hpp
+++ b/jlm/llvm/opt/alias-analyses/PointsToGraph.hpp
@@ -421,8 +421,8 @@ public:
    * if such a node exists. Otherwise nullopt is returned.
    * @see getNodeForRegister()
    */
-  std::optional<NodeIndex>
-  tryGetNodeForRegister(const rvsdg::Output & output) const
+  [[nodiscard]] std::optional<NodeIndex>
+  tryGetNodeForRegister(const rvsdg::Output & output) const noexcept
   {
     auto it = registerMap_.find(&output);
     if (it != registerMap_.end())
--- a/jlm/llvm/opt/alias-analyses/PointsToGraph.hpp
+++ b/jlm/llvm/opt/alias-analyses/PointsToGraph.hpp
@@ -16,6 +16,7 @@
 #include <jlm/util/Math.hpp>
 
 #include <memory>
+#include <optional>
 #include <string>
 #include <unordered_map>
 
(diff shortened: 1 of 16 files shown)


jlm/llvm/opt/alias-analyses/PointsToGraph.hpp
```diff
@@ -415,6 +416,20 @@ class PointsToGraph final
     return registerMap_.at(&output);
   }
 
+  /**
+   * Retrieves the index of the PointsToGraph node mapped to the given \p output,
+   * if such a node exists. Otherwise nullopt is returned.
+   * @see getNodeForRegister()
+   */
+  std::optional<NodeIndex>
+  tryGetNodeForRegister(const rvsdg::Output & output) const
```
phate: `noexcept` `[[nodiscard]`

jlm/llvm/opt/alias-analyses/PointsToGraphAliasAnalysisTests.cpp
```diff
@@ -3,6 +3,7 @@
  * See COPYING for terms of redistribution.
  */
 
+#include "jlm/llvm/opt/alias-analyses/PointsToGraph.hpp"
```
phate: <>

jlm/llvm/opt/alias-analyses/PointsToGraphAliasAnalysisTests.cpp
```diff
@@ -263,9 +264,9 @@ TEST(PointsToGraphAnalysisTests, TestPtGAliasAnalysis)
   // jlm::rvsdg::view(&rvsdg.graph().GetRootRegion(), stdout);
 
   Andersen andersen;
-  auto pointsToGraph = andersen.Analyze(rvsdg.module());
+  std::shared_ptr<PointsToGraph> pointsToGraph = andersen.Analyze(rvsdg.module());
```
phate: auto should be fine here, no?

jlm/llvm/opt/alias-analyses/PointsToGraphAliasAnalysisTests.cpp
```diff
@@ -453,9 +454,9 @@ TEST(PointsToGraphAnalysisTests, TestPtGAliasAnalysisOffsets)
...
```
phate: auto

phate (dismissed):

jlm/llvm/opt/alias-analyses/PointsToGraphAliasAnalysisTests.cpp
haved: We need a `std::move()` afterwards then, but that is fine

haved:
@phate fixed your comments

phate (approved):

PR #1665, opened 2026-05-19
AAG: add support for nested structs

phate:


phate:
This PR should not be merged before #1664

jlm/llvm/opt/AggregateAllocaSplitting.cpp
```diff
@@ -479,25 +472,62 @@ AggregateAllocaSplitting::findSplitableAllocaNodes(rvsdg::Region & region) const
   return traceInfo;
 }
 
+// FIXME: Do not use std::map, and rather use std::unordered_map
```
phate: Change to `std::unordered_map<>`

phate:
@haved This PR is ready for review.

jlm/llvm/opt/AggregateAllocaSplitting.cpp
```diff
@@ -224,11 +224,8 @@ AggregateAllocaSplitting::isSplitable(rvsdg::SimpleNode & allocaNode)
                   [&](const GetElementPtrOperation &)
                   {
                     JLM_ASSERT(userNode->input(0) == &user);
-                    if (const auto indicesOpt =
-                            GetElementPtrOperation::tryGetConstantIndices(simpleNode);
-                        !indicesOpt.has_value())
-                      return false;
-
+                    JLM_ASSERT(
+                        GetElementPtrOperation::tryGetConstantIndices(simpleNode).has_value());
```
haved, line 236: Why is this an assert? If the GEP has non-constant indices that just means we can not replace the alloca.

jlm/llvm/opt/AggregateAllocaSplitting.cpp
haved, line 125: We should also check the "count" input to the alloca. If the count is not 1 we should not try to separate, no?

jlm/llvm/opt/AggregateAllocaSplitting.cpp
```diff
@@ -508,14 +553,13 @@ AggregateAllocaSplitting::splitAllocaNode(const AllocaTraceInfo & allocaTraceInf
...
         [&](const GetElementPtrOperation &)
         {
-          JLM_ASSERT(GetElementPtrOperation::numIndices(*allocaConsumer) == 2);
+          JLM_ASSERT(GetElementPtrOperation::numIndices(*allocaConsumer) >= 2);
           auto & consumerRegion = *allocaConsumer->region();
-
           const auto indices =
               GetElementPtrOperation::tryGetConstantIndices(*allocaConsumer).value();
-          JLM_ASSERT(indices.size() == 2 && indices[0] == 0);
+          JLM_ASSERT(indices[0] == 0);
```
haved, line 568: How can you be sure about this? We never checked anywhere that the count is equal to 1, did we? Do we want to support cases where the count is != 1?

haved (changes requested):
Looks good, but we need to decide how we handle allocas with count != 1. I think the easiest is to just not touch them, but we need to perform the check when deciding.

jlm/llvm/opt/AggregateAllocaSplitting.cpp
phate: Yes, you are right. I added the check and a test for it.

jlm/llvm/opt/AggregateAllocaSplitting.cpp
phate: We are only handling struct types right now. As far as I understand it, it should never occur that we have a getelementptr into a struct that has non-constant indices.

jlm/llvm/opt/AggregateAllocaSplitting.cpp
phate: No, we do not want to support cases where count != 1 for now. I added the check and a FIXME to tackle this in the future.

phate:

jlm/llvm/opt/AggregateAllocaSplitting.cpp
haved: Ah yes, that is true, good point

haved (approved):

PR #1666, opened 2026-05-20
Ignore warning that appear in LLVM 18 library headers

haved:
Later versions of clang++ have a warning about methods in `final` classes being `virtual`.

This is a very reasonable warning, but some of the headers in LLVM18 have them.
Since we have `-Werror`, this is preventing me from building.


haved:
Okay clang++ does not like disabling warning that it does not know about, so I will not push this to master for now

PR #1674, opened 2026-06-05
region predicate tracing

caleridas:
Add RegionPredicateTrace which allows to:
- compute the predicate assignments effective from a region
- compute the predicate assignments required to enter region
- determine whether two regions can be visited simultaneously

This allows to make path-aware value analysis in the RVSDG.


jlm/rvsdg/RegionPredicateTrace.hpp
```diff
@@ -0,0 +1,174 @@
+/*
+ * Copyright 2026 Helge Bahmann <hcb@chaoticmind.net>
+ * See COPYING for terms of redistribution.
+ */
+
+#ifndef JLM_RVSDG_REGIONTRACE_HPP
```
phate: The guard is not aligned with the file name.

jlm/rvsdg/RegionPredicateTraceTests.cpp
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright 2026 Helge Bahmann <hcb@chaoticmind.net>
```
phate, line 2: The file name misses an "s" at the end. It should be ...Tests

jlm/rvsdg/RegionPredicateTrace.hpp
```diff
@@ -0,0 +1,174 @@
...
+ * Describes which values a predicate can take.
+ */
+class PredicateValueRange
+{
+public:
+  /**
+   * \brief Constructs empty value range (unsatisfiable predicate range).
+   */
+  static inline PredicateValueRange
+  Empty(const ControlType & type)
```
phate: Can prefix the static creators with create... to align with the rest of the code base.

phate (commented):
Methods should start with a small letter.

haved:
@caleridas I find the docstrings for the individual methods in the header a bit hard to understand. I'm not quite able to follow which regions are which, and if the inputs and origins are supposed to be inside the region, in a parent region or a sibling region. Could you include some examples in the docstrings?

caleridas:
> @caleridas I find the docstrings for the individual methods in the header a bit hard to understand. I'm not quite able to follow which regions are which, and if the inputs and origins are supposed to be inside the region, in a parent region or a sibling region. Could you include some examples in the docstrings?

@haved the intent is that "fromRegion" is the "upper" one (where the _origin_ of the value that you want to check), and "toRegion" is the "lower" one (this is the place where you want to check whether e.g. you can forward a load to)

That said, thinking about this closer, this distinction may not be terribly sensible, I could quite easily make this symmetric, so this basically answers the question "can these two distinct regions _ever_ be visited on the same pass?", and that is logically not even dependent on which one is the upper/lower one. I would just need to check both directions behind the scenes which "costs a bit more", but may in the end be utterly irrelevant.

If I change the interface that way (i.e. "check two regions for mutual flow exclusion"), is this then better comprehensible? (Maybe it would fit your need even better?) As said it "costs" a bit more which TBH I do not care too much about, but maybe you want to make a call here.

haved:
@caleridas looking back at the docstrings now, I understand them much better, so I don't think my comment from yesterday matters much. I prefer explicit functions, so forcing the caller to distinguish between the upstream and downstream region is fine by me. 

The only thing I think the comment on the last function could make clearer is what it actually tests. What is the implication that holds if the function returns true or false? The docstring says true means reachable, but does false mean target region is guaranteed to be unreachable from originRegion? I'm struggling to formulate precise implications that I can then do modus tollens with in my head

caleridas:
> @caleridas looking back at the docstrings now, I understand them much better, so I don't think my comment from yesterday matters much. I prefer explicit functions, so forcing the caller to distinguish between the upstream and downstream region is fine by me.
> 
> The only thing I think the comment on the last function could make clearer is what it actually tests. What is the implication that holds if the function returns true or false? The docstring says true means reachable, but does false mean target region is guaranteed to be unreachable from originRegion? I'm struggling to formulate precise implications that I can then do modus tollens with in my head

ah thanks for the feedback -- yes, I can clarify that to make it unabiguous -- will do!

jlm/rvsdg/RegionPredicateTrace.cpp
```diff
@@ -0,0 +1,269 @@
...
+  {
+    return MatchTypeWithDefault(
+        *node,
+        [&](const rvsdg::GammaNode & node) -> PredicateValueRange
+        {
+          auto argVar = node.MapBranchArgument(*origin);
+
+          if (auto entry = std::get_if<GammaNode::EntryVar>(&argVar))
+          {
+            return ComputeAndRecord(regionPredRange, *entry->input, type);
```
phate, line 200: You only care about the input for the argument here, and not all other arguments. Instead of computing the entire entry variable, you could just use `mapBranchArgumentToInput()`, no?

jlm/rvsdg/RegionPredicateTrace.cpp
```diff
@@ -0,0 +1,269 @@
...
+          }
+          else
+          {
+            return PredicateValueRange::Unknown(type);
+          }
+        },
+        [&](rvsdg::ThetaNode & node) -> PredicateValueRange
+        {
+          auto loopVar = node.MapPreLoopVar(*origin);
+          if (loopVar.post->origin() == loopVar.pre)
```
phate, line 210: `ThetaLoopVarIsInvariant()`

jlm/rvsdg/RegionPredicateTrace.hpp
caleridas: fixed

jlm/rvsdg/RegionPredicateTraceTests.cpp
caleridas: renamed

jlm/rvsdg/RegionPredicateTrace.hpp
caleridas: renamed

jlm/rvsdg/RegionPredicateTrace.cpp
caleridas: maybe, but I would for now just prefer to keep it "canonical" rather than using special-cased functions

jlm/rvsdg/RegionPredicateTrace.cpp
caleridas: I still need the full loop var description to trace out if needed -- since I have it I can also check directly here.

haved (approved):
Thank you for the extensive docstring, it helped a lot

PR #1676, opened 2026-06-08
handle sext, zext and trunc in tryGetConstantSignedInteger

haved:
As discovered when looking at StoreValueForwarding, the `tryGetSignedIntegerConstant` function would return `std::nullopt` as soon as a constant was passed through a `SExt` or `ZExt`. Ideally these cases should be handled by reductions, but it was relatively common for these situations to only become apparent during the middle of a SVF pass, so it makes sense to handle them immediately, as it is relatively cheap to do so.

As a part of this, I also did some cleanup of the operations:
SExt and ZExt were in different files (`sext.hpp` and `operations.hpp`), and their implementation differed in subtle ways, so I moved them to a shared `Bitcast.[ch]pp` along with Trunc and Bitcast, and cleaned up the static creation helper methods.

Only SExt had reductions defined, but most of these reductions would be incorrect as soon as any overflow gets involved. Doing these reductions properly requires putting flags on operations that guarantee no overflow / no information loss, so I removed the reductions for now.


jlm/llvm/backend/CastingTests.cpp
```diff
@@ -8,9 +8,9 @@
 #include <jlm/llvm/backend/IpGraphToLlvmConverter.hpp>
 #include <jlm/llvm/backend/RvsdgToIpGraphConverter.hpp>
 #include <jlm/llvm/frontend/LlvmModuleConversion.hpp>
+#include <jlm/llvm/ir/operators/Bitcast.hpp>
```
phate: Could we rename the files `ConversionOperations.[hpp/cpp]` and `ConversionOperationTests.cpp`? The operations in this file are in the `Conversion Operations` section of the LLVM reference manual: https://llvm.org/docs/LangRef.html#conversion-operations

jlm/llvm/ir/operators/Bitcast.hpp
```diff
@@ -0,0 +1,292 @@
+/*
+ * Copyright 2017 Nico Reißmann <nico.reissmann@gmail.com>
+ * See COPYING for terms of redistribution.
+ */
+
+#ifndef JLM_LLVM_IR_OPERATORS_SEXT_HPP
+#define JLM_LLVM_IR_OPERATORS_SEXT_HPP
```
phate: The header guard needs to be changed as well.

jlm/llvm/ir/operators/BitcastTests.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <jlm/llvm/ir/operators/Bitcast.hpp>
+#include <jlm/llvm/ir/operators/operators.hpp>
+#include <jlm/rvsdg/bitstring/arithmetic.hpp>
+#include <jlm/rvsdg/bitstring/constant.hpp>
+#include <jlm/rvsdg/bitstring/value-representation.hpp>
+#include <jlm/rvsdg/NodeNormalization.hpp>
+#include <jlm/rvsdg/simple-node.hpp>
+#include <jlm/rvsdg/view.hpp>
+
+TEST(BitcastTests, testSextZextConstantReductions)
```
phate: This file name along with the test names would also need to be adjusted.

phate (commented):
Could we separate the moving and adjusting of the conversion operations from extending the getelementptr computation in different PRs?

haved:
@phate this PR is ready now

jlm/llvm/ir/Trace.cpp
```diff
@@ -88,6 +90,60 @@ tryGetConstantSignedInteger(const rvsdg::Output & output)
...
+      return std::nullopt;
+
+    // When doing sign extensions, we only need to care about the size of the input type
+    const auto inputType = sextOp->argument(0);
+    const auto inputBits = util::assertedCast<const rvsdg::BitType>(inputType.get())->nbits();
+    JLM_ASSERT(inputBits <= 64);
+    const auto extendBits = 64 - inputBits;
+
+    // Shift signed value left and right again to sign extend
+    return (static_cast<int64_t>(*inputValue) << extendBits) >> extendBits;
```
phate: This feels like it show go somewhere else. `utils/Math.hpp` maybe?

jlm/llvm/ir/Trace.cpp
```diff
@@ -88,6 +90,60 @@ tryGetConstantSignedInteger(const rvsdg::Output & output)
...
+      return std::nullopt;
+
+    // When doing zero extensions, we only need to care about the size of the input type
+    const auto inputType = zextOp->argument(0);
+    const auto inputBits = util::assertedCast<const rvsdg::BitType>(inputType.get())->nbits();
+    JLM_ASSERT(inputBits <= 64);
+    const auto extendBits = 64 - inputBits;
+
+    // Shift unsigned value left and right again to zero extend
+    return (static_cast<uint64_t>(*inputValue) << extendBits) >> extendBits;
```
phate: same here

jlm/llvm/ir/Trace.cpp
```diff
@@ -88,6 +90,60 @@ tryGetConstantSignedInteger(const rvsdg::Output & output)
...
+      return std::nullopt;
+
+    // When truncating, we keep only the desired output bits and sign extend the rest
+    const auto outputType = truncOp->result(0);
+    const auto outputBits = util::assertedCast<const rvsdg::BitType>(outputType.get())->nbits();
+    JLM_ASSERT(outputBits <= 64);
+    const auto eraseBits = 64 - outputBits;
+
+    // Shift signed value left and right again to clear the top eraseBits with sign bits
+    return (static_cast<int64_t>(*inputValue) << eraseBits) >> eraseBits;
```
phate: same here

jlm/llvm/ir/Trace.cpp
phate: or maybe on the respective Operation class?

jlm/llvm/ir/TraceTests.cpp
```diff
@@ -133,3 +134,71 @@ TEST(TraceTests, testGetConstantSignedInteger)
   EXPECT_EQ(tryGetConstantSignedInteger(*matchOutput), std::nullopt);
   EXPECT_EQ(tryGetConstantSignedInteger(*lambdaOutput), std::nullopt);
 }
+
```
phate, line 137: Maybe one more test that ensures you recursively call tryGetConstantSignedInteger()? IN other words, the conversion operation below a gamma and the constant above the gamma, or something like that.

jlm/llvm/ir/Trace.cpp
haved: Moved it to math, and also used the much cleaner `sextOp->nsrcbits()` instead of typecasting BitTypes

haved:
@phate I moved the logic to `Math.hpp` and added more tests

jlm/util/Math.hpp
```diff
@@ -115,6 +118,50 @@ BitWidthOfEnum(T endValue)
   return BitsRequiredToRepresent(static_cast<UnderlyingT>(endValue));
 }
 
+/**
+ * Takes a 64-bit integer \p value and keeps the \p inputBit least significant bits.
```
phate: inputBit does not exist as parameter

jlm/util/Math.hpp
```diff
@@ -115,6 +118,50 @@ BitWidthOfEnum(T endValue)
...
+{
+  JLM_ASSERT(keepBits <= 64);
+  const auto extendBits = 64 - keepBits;
+
+  // Shift signed value left and right again to sign extend
+  return (value << extendBits) >> extendBits;
+}
+
+/**
+ * Takes a 64-bit integer \p value and keeps the \p inputBit least significant bits.
```
phate: same here

phate (dismissed):
Two minor comments. Doxygen should actually fail because of it.

phate:
Hmm doxygen does not fail. I am actually wondering why.

haved:
@phate I fixed it, weird that Doxygen didn't fail

phate (approved):

PR #1678, opened 2026-06-09
SVF: forward loads with no memory states

phate:
This adds support for forwarding loads with no memory state that can be traced to a delta node containing an integer constant.

### DIFF
--- a/jlm/llvm/opt/StoreValueForwarding.cpp
+++ b/jlm/llvm/opt/StoreValueForwarding.cpp
@@ -854,12 +854,11 @@ StoreValueForwarding::processLoadWithoutMemoryStates(rvsdg::SimpleNode & loadNod
 
   context_->statistics.startTracing();
   const auto tracedDelta = traceLoadWithoutMemoryStates(loadNode);
+  context_->statistics.stopTracing();
   if (!tracedDelta.has_value())
   {
-    context_->statistics.stopTracing();
     return;
   }
-  context_->statistics.stopTracing();
 
   context_->statistics.startForwarding();
   forwardLoadWithoutMemoryStates(loadNode, tracedDelta.value());
@@ -873,7 +872,7 @@ StoreValueForwarding::traceLoadWithoutMemoryStates(const rvsdg::SimpleNode & loa
   JLM_ASSERT(LoadOperation::numMemoryStates(loadNode) == 0);
 
   auto & loadAddress = *LoadOperation::AddressInput(loadNode).origin();
-  const auto & tracedAddress = TracePointerOriginPrecise(loadAddress);
+  const auto tracedAddress = TracePointerOriginPrecise(loadAddress);
   if (!tracedAddress.Offset.has_value())
   {
     return std::nullopt;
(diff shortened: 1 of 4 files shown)


jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -797,34 +808,142 @@ class LoadTracingInfo
...
+}
+
+std::optional<StoreValueForwarding::TracedDelta>
+StoreValueForwarding::traceLoadWithoutMemoryStates(const rvsdg::SimpleNode & loadNode)
+{
+  JLM_ASSERT(is<LoadNonVolatileOperation>(&loadNode));
+  JLM_ASSERT(LoadOperation::numMemoryStates(loadNode) == 0);
+
+  auto & loadAddress = *LoadOperation::AddressInput(loadNode).origin();
+  const auto & tracedAddress = TracePointerOriginPrecise(loadAddress);
```
haved: Why is  this a reference? `TracePointerOriginPrecise` returns by value, so I don't know how this ends up working in practice

jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -797,34 +808,142 @@ class LoadTracingInfo
...
 
+void
+StoreValueForwarding::processLoadWithoutMemoryStates(rvsdg::SimpleNode & loadNode)
+{
+  JLM_ASSERT(is<LoadNonVolatileOperation>(&loadNode));
+  JLM_ASSERT(LoadOperation::numMemoryStates(loadNode) == 0);
+
+  context_->statistics.startTracing();
+  const auto tracedDelta = traceLoadWithoutMemoryStates(loadNode);
+  if (!tracedDelta.has_value())
```
haved, line 858: If you move the `has_value()` check below the call to `statistics.stopTracing()`, you only need one call

jlm/llvm/opt/StoreValueForwarding.hpp
```diff
@@ -72,7 +72,59 @@ class StoreValueForwarding final : public rvsdg::Transformation
...
-  processLoadNode(rvsdg::SimpleNode & loadNode);
+  processLoad(rvsdg::SimpleNode & loadNode);
+
+  /**
+   * Process a \ref LoadNonVolatileOperation node with no memory states during traversal
+   *
+   * @param loadNode The \ref LoadNonVolatileOperation node to handle
+   */
+  void
+  processLoadWithMemoryStates(rvsdg::SimpleNode & loadNode);
```
haved: This is the only function in the "load with memory states" category to be above the "load without memory states" functions. All the others are below. Could we move it down beneath `forwardLoadWithoutMemoryStates` both in the header and in the .cpp file?

haved (dismissed):
Looks good, I just have some nitpick suggestions @phate

jlm/llvm/opt/StoreValueForwarding.cpp
phate: Changed

jlm/llvm/opt/StoreValueForwarding.cpp
phate: Done

jlm/llvm/opt/StoreValueForwarding.hpp
phate: Done

phate:
@haved I addressed your comments

phate:
@haved I also mapped out all the operations that we need to implement in order to get full coverage for all loads that can be traced to a delta. Please have a look.

haved (approved):

PR #1679, opened 2026-06-09
LLVM: add creators to IntegerOperation classes

phate:


phate:
@haved Let's merge this one before #1678

haved (approved):

PR #1680, opened 2026-06-09
Add depth info to regions, and use it for isAncestor

haved:
Having depth info in necessary for figuring out routing between sibling regions in a reasonable way.

Along with adding a depth field I added comments to the fields in the `Region` class and docstrings to the getters.

The `Region::isAncestor` function is updated to make use of the depth info.

I also removed the unused `Region*` parameter from the `Region` constructor.

### DIFF
--- a/jlm/rvsdg/region.cpp
+++ b/jlm/rvsdg/region.cpp
@@ -4,6 +4,7 @@
  * See COPYING for terms of redistribution.
  */
 
+#include <filesystem>
 #include <jlm/rvsdg/DotWriter.hpp>
 #include <jlm/rvsdg/graph.hpp>
 #include <jlm/rvsdg/structural-node.hpp>
@@ -144,23 +145,25 @@ Region::~Region() noexcept
   }
 }
 
-Region::Region(Region *, Graph * graph)
-    : id_(graph->generateRegionId()),
-      index_(0),
-      graph_(graph),
-      nextNodeId_(0),
+Region::Region(Graph * graph)
+    : graph_(graph),
+      id_(graph->generateRegionId()),
+      depth_(0),
       node_(nullptr),
+      index_(0),
+      nextNodeId_(0),
       numTopNodes_(0),
       numBottomNodes_(0),
       numNodes_(0)
 {}
 
 Region::Region(StructuralNode * node, const size_t index)
-    : id_(node->graph()->generateRegionId()),
-      index_(index),
-      graph_(node->graph()),
-      nextNodeId_(0),
+    : graph_(node->graph()),
+      id_(node->graph()->generateRegionId()),
+      depth_(node->region()->getDepth() + 1),
       node_(node),
+      index_(index),
+      nextNodeId_(0),
       numTopNodes_(0),
       numBottomNodes_(0),
       numNodes_(0)
@@ -470,17 +473,19 @@ Region::NumRegions(const rvsdg::Region & region) noexcept
 bool
 Region::isAncestor(const Region & region, const Region & ancestor) noexcept
 {
-  auto current = &region;
+  const auto ancestorDepth = ancestor.getDepth();
+  // If region starts at the same level or higher than ancestor, it can not be an ancestor
+  if (region.getDepth() <= ancestorDepth)
+    return false;
 
-  while (!current->IsRootRegion())
+  // Follow the parent chain until the ancestor depth is reached
+  auto current = &region;
+  do
   {
     current = current->node()->region();
+  } while (current->getDepth() > ancestorDepth);
 
-    if (current == &ancestor)
-      return true;
-  }
-
-  return false;
+  return current == &ancestor;
 }
 
 std::string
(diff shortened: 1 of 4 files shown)


jlm/rvsdg/region.hpp
```diff
@@ -790,11 +819,20 @@ class Region
...
+  // The depth of the region. The root region has depth 0, all others have parent region depth + 1
+  size_t depth_;
+
+  // The structural node this region belong to, or nullptr if it is the root region
+  rvsdg::StructuralNode * node_;
+  // The index of the region, if it is a subregion of a structural node
   size_t index_;
-  Graph * graph_;
+
+  // The ID that will be given to the next node created in this region
```
phate: Any particular reason why you do not use doxygen comments?

jlm/rvsdg/region.hpp
```diff
@@ -265,6 +280,38 @@ class Region
...
+  inline rvsdg::StructuralNode *
+  node() const noexcept
+  {
+    return node_;
+  }
+
+  /**
+   * @return the index of the region within the surrounding structural node.
+   */
+  size_t
```
phate: `[[nodiscard]`

jlm/rvsdg/region.hpp
```diff
@@ -265,6 +280,38 @@ class Region
...
+  getDepth() const noexcept
+  {
+    return depth_;
+  }
+
+  /**
+   * @return the structual node this region is a subregion of.
+   * If the region is the root region, nullptr is returned instead.
+   */
+  inline rvsdg::StructuralNode *
```
phate: 1. no `inline`
2. `[[nodiscard]`

jlm/rvsdg/region.hpp
```diff
@@ -265,6 +280,38 @@ class Region
...
   }
 
+  /**
+   * Gets the depth of the region.
+   * Region depth is defined as 0 for the root region,
+   * while all other regions have a depth equal to the depth of their parent region + 1.
+   *
+   * @return the depth of the region
+   */
+  size_t
```
phate: `[[nodiscard]]`

jlm/rvsdg/region.cpp
```diff
@@ -470,17 +472,19 @@ Region::NumRegions(const rvsdg::Region & region) noexcept
 bool
 Region::isAncestor(const Region & region, const Region & ancestor) noexcept
```
phate, line 473: Do we have a unit test on this method? If not, would you mind adding one?

jlm/rvsdg/region.hpp
haved: Just because they take up more space, and there were some // comments there already

jlm/rvsdg/region.cpp
haved: There was one already, but I expanded it now

phate (approved):

phate (changes requested):
Blocking it, since I want something to be clarified before: What happens if a structural node is moved out of a structural. This PR has nowhere code for updating the depth of a region.

phate:
I put @caleridas on this PR as well, as the non-updating of region depth almost slipped through.

haved:
@phate @caleridas  there is (currently) no way of moving regions out of structural nodes. You have to create a new region and copy over nodes. To make this clearer I marked all the fields in the `Region` class `const`.

jlm/rvsdg/region.hpp
haved: I changet them to multiline doxygen comments

caleridas:
Is numerical region depth strictly needed, or is it just convenience / speed up? I would generally be cautious materialising "recoverable" information -- maybe it is doable to cache region relation computation in separate data structure instead?

phate (approved):

haved:
@caleridas I would like to have the region depth easily at hand, since otherwise finding the least common ancestor between two regions becomes a pain, needing unbounded amounts of memory.

I prefer to not have to maintain an auxiliary structure, since I then have to pass it around to all the different places I may want to do some routing.

Since the current API does not even allow moving regions, maintaining depth is easy, and even if we did allow moving regions in the future, I still think it would be worth it to keep the depth pre-computed, as we traverse the region tree a lot more than we modify it.

caleridas:
> @caleridas I would like to have the region depth easily at hand, since otherwise finding the least common ancestor between two regions becomes a pain, needing unbounded amounts of memory.
> 
> I prefer to not have to maintain an auxiliary structure, since I then have to pass it around to all the different places I may want to do some routing.
> 
> Since the current API does not even allow moving regions, maintaining depth is easy, and even if we did allow moving regions in the future, I still think it would be worth it to keep the depth pre-computed, as we traverse the region tree a lot more than we modify it.

I would still caution as it "breaks" locality. I would long-term definitely like to be able construct graphs "bottom-up", i.e. create subregion, then "wrap" nodes around etc. That is definitely at odds.

PR #1689, opened 2026-06-16
LLVM: add GetElementPtrOperation::Constant::getOffetInBytes()

phate:


phate:
@haved I know that this method is currently a copy of the `GetElementPtrOperation::CalculateOffset()`, but the plan is to make this one obsolete and replace it with the following pattern:

```
auto constant = GetElementPtrOperation::tryGetAsConstant(...);
auto offset = constant.offsetInBytes();
```

haved (approved):

PR #1693, opened 2026-06-17
Llvm: Fix miscompilation of theta nodes

### LINKED ISSUE #1691: Miscompile of tail controlled loops
Based on the miscompilation issue we have with picojpeg in embench.

Given a C program like:
```c
static int c = 0;
int get_next() {
    return c++;
}

int main () {
  int status;

  while (1) {
    status = get_next();

    if (status == 5)
      break;
  }

  return status;
}
```
we would expect the main method to always return 5. However, after round-tripping the LLVM IR through jlm-opt (without any passes):
```sh
clang-18 file.c -Os -Xclang -disable-llvm-passes -S -emit-llvm -o file.ll
jlm-opt file.ll -o file-jlm-opt.ll --dumpRvsdgGraphs -s statistics
clang-18 file-jlm-opt.ll -o file-binary
./file-binary
echo $?  # Prints 0
```
We get return code 0, due to the fact that the while loop stops after the first iteration.

The LLVM IR produced by clang (with some mem2reg and manual cleanup) looks like:
```llvm
  br label %1
1:                                                ; preds = %0, %1
  %3 = call i32 @get_next() #2
  %4 = icmp eq i32 %3, 5
  br i1 %4, label %5, label %1

5:                                                ; preds = %2
  ret i32 %3
}
```

The jlm frontend converts this to the following RVSDG:
Image

In the RVSDG, the match node controlling the theta predicate gives the opposite control signal from the one we expect. (When status == 5 it keeps looping, otherwise terminate. We expect it to stop looping iff status==5)

phate:
We experienced a miscompilation if the repetition edge of a tail-controlled loop had index 0 in the control flow graph. This collided with the expectation of the RVSDG's theta node that the repetition happens on control value 1.

Closes #1691

### DIFF
--- a/jlm/llvm/frontend/ControlFlowRestructuring.cpp
+++ b/jlm/llvm/frontend/ControlFlowRestructuring.cpp
@@ -240,6 +240,60 @@ RestructureControlFlow(
     ControlFlowGraphNode &,
     std::vector<TailControlledLoop> &);
 
+/**
+ * The RVSDG theta node expects that the repetition of its body always happens on control value 1.
+ * Thus, we need to ensure that the repetition edge of the tail-controlled
+ * loop has index 1 in order to avoid miscompilations.
+ *
+ * @param sccStructure The \ref StronglyConnectedComponentStructure that represents the
+ * tail-controlled loop.
+ */
+static void
+adjustLoopRepetitionEdge(const StronglyConnectedComponentStructure & sccStructure)
+{
+  JLM_ASSERT(sccStructure.IsTailControlledLoop());
+
+  const auto repetitionEdge = *sccStructure.RepetitionEdges().begin();
+  if (repetitionEdge->index() == 1)
+  {
+    // Nothing needs to be done. The repetition edge has already index 1.
+    return;
+  }
+
+  // The repetition edge has index 0.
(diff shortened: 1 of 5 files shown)


jlm/llvm/frontend/ControlFlowRestructuring.cpp
```diff
@@ -240,6 +240,60 @@ RestructureControlFlow(
...
+
+  repetitionEdge->divert(exitEdgeSink);
+  exitEdge->divert(repetitionEdgeSink);
+
+  // Branch and match operations are always created together in the \ref LlvmModuleConversion
+  // Thus, we would expect them to also be together here. Just to be sure, we plastered the path
+  // with asserts.
+  auto & threeAddressCodes = exitNode->tacs();
+  JLM_ASSERT(is<BranchOperation>(threeAddressCodes.last()));
+  auto matchTac = *std::next(threeAddressCodes.rbegin(), 1);
```
haved, line 279: Also assert that the branch uses the match as its operand

phate: Done

haved (approved):

PR #1695, opened 2026-06-18
SVF: Replace loads of floatin point type from aggregate zero-initialized deltas

haved:
Some of the benchmark programs threw exceptions during compilation due to missing this


phate:
@haved Only tangible to this PR: Why did the CI not yell at us for this? Which file in the benchmarks did it affect? Is it any of the benchmarks that are compiled as part of the CI? If so, it should have triggered as these are compile errors, no?

haved:
@phate there were still some other situations that occurred in other files, such as a Fixed vector of two bit64 values, or an IntToPtr in the delta initializer. Instead of implementing anything more in this PR I just added if-cases with FIXME comments for them

phate (approved):

PR #1696, opened 2026-06-18
Make benchmark CI job fail when the benchmarking fails

haved:
This also updates the commit such that we get validation of the compiled benchmarks as well, for those that have a validation script (currently only embench and a tiny list script for emacs)

### DIFF
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -38,14 +38,15 @@ jobs:
             cactuBSSN,
             x264,
             blender,
+            imagick,
             nab,
             xz,
             emacs,
             ghostscript,
             gdb,
             sendmail,
-            embench,
             polybench,
+            embench,
           ]
     runs-on: ubuntu-24.04
     steps:
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -33,6 +33,7 @@ jobs:
       matrix:
         benchmark:
           [
+          # Programs from SPEC CPU2017 (redist version)
             perlbench,
             gcc,
             cactuBSSN,
@@ -41,12 +42,16 @@ jobs:
             imagick,
             nab,
             xz,
-            emacs,
-            ghostscript,
-            gdb,
-            sendmail,
-            polybench,
+
+          # Other benchmark suites
             embench,
+            polybench,
+
+          # Open source programs
+            emacs,
+            gdb,
+            ghostscript,
+            sendmail,
           ]
     runs-on: ubuntu-24.04
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/Benchmarks.yml
```diff
@@ -33,14 +33,19 @@ jobs:
...
         benchmark:
           [
-            polybench,
             perlbench,
             gcc,
             cactuBSSN,
             x264,
             blender,
             nab,
             xz,
```
phate, line 44: Are those all spec benchmarks we are looking at, or should there be more? I have the feeling that some are missing. Could you double check, please?

scripts/run-benchmarks.sh
```diff
@@ -20,8 +20,8 @@ else
 fi
 
 APT_INSTALL_DEPS=false
-CLEAN=false
-BENCHMARK=""
+CLEAN_RUNS=false
+RUN_OPTIONS=""
```
phate: Why did you call it run_options? I find it rather misleading. What was wrong with benchmarks?

haved: changed it back to use separate variables for filters and other extra options

.github/workflows/Benchmarks.yml
haved: Good point, imagick was missing from both the help text and the CI list

.github/workflows/Benchmarks.yml
```diff
@@ -33,14 +33,20 @@ jobs:
...
             blender,
+            imagick,
             nab,
             xz,
+            emacs,
+            ghostscript,
+            gdb,
+            sendmail,
+            polybench,
+            embench,
```
phate: Nitpick: Alphabetically?

scripts/run-benchmarks.sh
```diff
@@ -20,8 +20,12 @@ else
...
 
 APT_INSTALL_DEPS=false
-CLEAN=false
-BENCHMARK=""
+CLEAN_RUNS=false
+
+# Used for filtering to only run certain benchmarks
+BENCHMARKS=""
+# Other options to pass to the ./run.sh script
+EXTRA_RUN_OPTIONS=""
```
phate: EXTRA_RUN_OPTIONS is not mentioned in the help text.

phate: Is that deliberate?

.github/workflows/Benchmarks.yml
haved: I grouped them in the run.sh help text. The SPEC ones are together, sorted by their number. Then the other benchmark suites, then the open source programs, sorted alphabetically. I now do the same grouping in the workflow file, with comments for each group

scripts/run-benchmarks.sh
haved: It is only used for the --ci flag only right now, so I'll change it

haved:
@phate at last, they are all green. None of the benchmarks are slower than the 20 minute llvm test suite

phate (approved):

PR #1699, opened 2026-06-20
Add RegionTreeObserver class

phate:


phate:
@caleridas I still need to add tests, but I wanted your opinion first before I continue to implement them. I decided to rather implement a RegionTreeObserver than a GraphObserver as this struck me as more flexible.

haved:
@phate should structural nodes created after the observer also have their subregions observed?

phate:
@haved Yeah, definitely. Also, nodes that are removed should the observer be destroyed.....yeah big oversight. Great catch!

PR #1701, opened 2026-06-20
Fix non-determinism in JSON generation

sjalander:
Depending on platform (Linux or MAC) the generated JSON is ordered differently, which breaks the unit tests. This fix sorts the generated JSON such that the order becomes deterministic.

### DIFF
--- a/jlm/rvsdg/region.cpp
+++ b/jlm/rvsdg/region.cpp
@@ -515,13 +515,15 @@ Region::toJson(
 
   if (annotationMap.HasAnnotations(&region))
   {
-    auto annotations = const_cast<std::vector<util::Annotation> &>(
-        annotationMap.GetAnnotations(&region));
+    auto annotations =
+        const_cast<std::vector<util::Annotation> &>(annotationMap.GetAnnotations(&region));
     std::sort(
         annotations.begin(),
         annotations.end(),
         [](const util::Annotation & a, const util::Annotation & b)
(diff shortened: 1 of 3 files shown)


jlm/rvsdg/region.cpp
```diff
@@ -514,8 +515,18 @@ Region::toJson(
 
   if (annotationMap.HasAnnotations(&region))
   {
+    auto annotations =
+        const_cast<std::vector<util::Annotation> &>(annotationMap.GetAnnotations(&region));
```
phate: Please do not use const_cast<>() here. Assign to a new annotation map and sort that one.

jlm/rvsdg/region.cpp
```diff
@@ -568,8 +579,18 @@ Region::toJson(
   {
     stream << ",";
 
+    auto annotations =
+        const_cast<std::vector<util::Annotation> &>(annotationMap.GetAnnotations(&structuralNode));
```
phate: same here

jlm/rvsdg/region.cpp
sjalander: Not using const_cast<>()

jlm/rvsdg/region.cpp
sjalander: Not using const_cast<>()

phate (approved):

PR #1704, opened 2026-06-21
SVF: add load forwarding for ConstantDataArray operations

phate:


phate:
@haved The PR is finally ready for review.

phate:
@haved If you want me to move the value-representation change to its own PR, just let me know.

phate:
@haved Also, if I remember correctly, this case was the crucial one for forwarding loads and was accounting for the significant majority of the cases. So getting some before/after numbers would be nice.

jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -934,9 +966,48 @@ StoreValueForwarding::forwardLoadWithoutMemoryStates(
...
+                  rvsdg::TryGetOwnerNode<rvsdg::SimpleNode>(*node->input(n)->origin());
+              const auto constantOp =
+                  util::assertedCast<const IntegerConstantOperation>(&elementNode->GetOperation());
+              constants.push_back(constantOp->Representation());
+            }
+            JLM_ASSERT(constants.size() == numRequiredConstants);
+
+            auto & constantNode = IntegerConstantOperation::Create(
+                *loadNode.region(),
+                rvsdg::BitValueRepresentation::create(constants));
```
haved, line 1007: This function makes an assumption about endianness, which is likely to always be little, but we should at least include a comment containing the word Endianness, either here or in the BitValueRepresentation create function, so that we easily find every location where endianness matters

haved (dismissed):
Looks good, just one comment about endianness assumptions @phate

jlm/llvm/opt/StoreValueForwarding.cpp
phate: @haved I added a FIXME

haved (approved):

PR #1706, opened 2026-06-21
hls: Intitial R-HLS to FIRRTL conversion tests

sjalander:


phate (commented):
Could we get the fixes separated from the unit test implementations?

I mean, there is several orthogonal things going on here:
1. Fixing of library order
2. Fixing of memory leaks
3. Unit tests

They can all be separated, no? If not 1 from 2, then at least 1/2 from 3

sjalander:
Well, it is the tests that exposed the other issues, so they are what validate that the other changes work.

phate (dismissed):

sjalander:
@phate I had to change the IntToPtrOperation name, so I need a new review.

phate (approved):

PR #1712, opened 2026-06-24
hls: More extensive GammaConversion tests

sjalander:


jlm/hls/backend/rvsdg2rhls/GammaConversionTests.cpp
```diff
@@ -0,0 +1,318 @@
...
+#include <jlm/rvsdg/TestType.hpp>
+#include <jlm/rvsdg/traverser.hpp>
+
+using namespace jlm::rvsdg;
+using namespace jlm::hls;
+using namespace jlm::llvm;
+using namespace jlm::util;
+
+namespace
+{
```
phate: The tests live alongside the source code. So they should preferably be in jlm::hls as well (I know that this is not done consequently so far)

phate: Just throw everything in jlm::hls

phate (dismissed):

jlm/hls/backend/rvsdg2rhls/GammaConversionTests.cpp
sjalander: Done

phate: I think you misunderstood: The entire file can be put in the jlm::hls namespace, and you can remove the using namespace jlm::hls at the beginning then.

phate (dismissed):

jlm/hls/backend/rvsdg2rhls/GammaConversionTests.cpp
phate: It is a minor thing though. Your choice.

sjalander: I put everything in jlm::hls

phate (approved):

PR #1713, opened 2026-06-28
make Trace's TraceAllPointerOrigins correctly handle theta pre arguments

haved:
We were missing handling of pointers that originate from loop variable pre arguments. This caused the LocalAliasAnalysis to give incorrect responses in some cases, but only in the aggressive case.

This PR adds tests to both the tracer and the LocalAliasAnalysis. Both tests fail without the changes. The change does not make the LocalAliasAnalysis noticably "worse" in terms of removed loads.


phate (dismissed):

haved:
@phate I fixed the doxygen complaints, may I get a new review, please?

phate (approved):

PR #1715, opened 2026-07-05
LLVM: Add constant folding for integer comparison operations

phate:


jlm/llvm/ir/operators/IntegerOperations.cpp
```diff
@@ -4,10 +4,136 @@
...
+  const auto & tracedOperand2 = llvm::traceOutput(operand2);
+  auto [c2Node, c2Operation] =
+      rvsdg::TryGetSimpleNodeAndOptionalOp<IntegerConstantOperation>(tracedOperand2);
+  if (!c2Operation)
+    return std::nullopt;
+
+  auto & c1Representation = c1Operation->Representation();
+  auto & c2Representation = c2Operation->Representation();
+
+  auto fold = getBinaryConstantFolder<TBinOp>();
```
haved: Is there a reason why we have a higher order function returning a function that we then immediately invoke, instead of passing the arguments directly to the function and returning the value directly?

jlm/llvm/ir/operators/IntegerOperations.cpp
```diff
@@ -4,10 +4,136 @@
...
+    return [](const IntegerValueRepresentation & r1, const IntegerValueRepresentation & r2)
+    {
+      return IntegerValueRepresentation::repeat(1, r1.ugt(r2));
+    };
+  }
+  else if constexpr (std::is_same_v<TBinOp, IntegerUleOperation>)
+  {
+    return [](const IntegerValueRepresentation & r1, const IntegerValueRepresentation & r2)
+    {
+      return IntegerValueRepresentation::repeat(1, r1.ule(r2));
```
haved: Using `repeat` was a bit confusing to me at first, could we have a `singleBitValue` constructor, maybe?

haved (dismissed):
Looks good, I just have one question about the use of a higher order function, as I don't see why it is needed currently @phate

jlm/llvm/ir/operators/IntegerOperations.cpp
phate: Done.

jlm/llvm/ir/operators/IntegerOperations.cpp
phate: Done.

phate:

haved (dismissed):

sjalander:
Some impressive reductions in cycle count :)

phate:
@sjalander This makes me question a bit the usefulness of some of those tests actually. I expected reductions, but some of them feel too high.

phate:
@haved I updated the branch and the golden cycles. Please have a look again and approve if everything looks all right.

haved (approved):

PR #1716, opened 2026-07-08
LLVM: Add constant folding for match operation

phate:


haved (approved):
Looks good, just some hls tests are failing @phate

Edit: I see you updated cycles

PR #1719, opened 2026-07-10
LLVM: Add constant folding for ZExtOperation

phate:


haved (dismissed):

haved:
@phate looks like this PR had a slight effect on a hls cycle count

haved (approved):

PR #1730, opened 2026-07-13
SVF: add load forwarding for ConstantStruct and ConstantArray operations

phate:


jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -1096,6 +1096,242 @@ getConstantDataArrayElementIndex(
...
+          return &output;
+        },
+        [&](const ConstantArrayOperation & constantArrayOperation)
+        {
+          const auto arrayType = constantArrayOperation.type();
+          const auto elementSizeInBytes = GetTypeAllocSize(*arrayType->GetElementType());
+
+          const auto index = elementOffsetInBytes / elementSizeInBytes;
+          return getDeltaElement(
+              elementOffsetInBytes % elementSizeInBytes,
```
phate: This is wrong

phate:
@haved I would like to merge this branch as it is, and then fix the leftover FIXMEs in a follow up branches. However, before doing so, I would like to see that we are not regressing (too much) in terms of forward loads and eliminated stores. Would you be so kind and get me some numbers for a before/after comparison for this PR?

haved (approved):

PR #1738, opened 2026-07-15
RVSDG: update distribution reduction of BitSliceOperation

phate:


jlm/rvsdg/bitstring/slice.hpp
```diff
@@ -73,6 +73,25 @@ class BitSliceOperation final : public UnaryOperation
...
 
+  /**
+   * Distribute a \ref BitSliceOperation node over a \ref BitConcatOperation node:
+   *
+   * c = BitConcatOperation x1[3] x2[3] x3[3]
+   * s = BitSliceOperation[1:4] c
+   * =>
+   * s1 = BitSliceOperation[1:2]
+   * s2 = BitSliceOperation[0:1]
+   * s = BitConcatOperation s1 s2
```
haved, line 84: This example is giving me pause. Should the "after" part refer to `x1` and `x2`, and should the starting concatenation say `[0:2]` instead of `[3]`?

phate: @haved I updated the example. It should make sense now.

haved (approved):

PR #1740, opened 2026-07-16
LLVM: add PtrCmpOperation reduction against nullptr

phate:


jlm/llvm/ir/operators/OperatorsTests.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include <jlm/llvm/ir/operators/alloca.hpp>
+#include <jlm/llvm/ir/operators/IntegerOperations.hpp>
+#include <jlm/llvm/ir/operators/operators.hpp>
+#include <jlm/rvsdg/NodeNormalization.hpp>
+
+namespace jlm::llvm
+{
+
+TEST(PtrCmpOperationTests, testNormalizeNullPointerComparison)
```
phate, line 20: Extend test

jlm/llvm/ir/operators/operators.cpp
```diff
@@ -269,6 +272,86 @@ PtrCmpOperation::reduce_operand_pair(
...
+  {
+    return true;
+  }
+
+  if (rvsdg::TryGetOwnerNode<rvsdg::DeltaNode>(output))
+  {
+    return true;
+  }
+
+  if (const auto region = rvsdg::TryGetOwnerRegion(output); region && region->IsRootRegion())
```
haved: We could also check if it is an instance of `rvsdg::GraphImport`

jlm/llvm/ir/operators/operators.cpp
```diff
@@ -269,6 +272,86 @@ PtrCmpOperation::reduce_operand_pair(
...
+      (isOutputOf<ConstantPointerNullOperation>(tracedOperand1) && isAllocationSide(tracedOperand2))
+      || (isOutputOf<ConstantPointerNullOperation>(tracedOperand2)
+          && isAllocationSide(tracedOperand1));
+  if (hasRequiredOperands)
+  {
+    auto & region = *operands[0]->region();
+    switch (ptrCmpOperation.predicate())
+    {
+    case ICmpPredicate::Eq:
+      return std::vector<rvsdg::Output *>{
```
haved: there is also `rvsdg::outputs()` that takes a node and returns a vector of all its outputs

jlm/llvm/ir/operators/operators.cpp
```diff
@@ -269,6 +272,86 @@ PtrCmpOperation::reduce_operand_pair(
...
+        IntegerConstantOperation::Create(region, 1, 0).output(0)
+      };
+    case ICmpPredicate::Ne:
+      return std::vector<rvsdg::Output *>{
+        IntegerConstantOperation::Create(region, 1, 1).output(0)
+      };
+    default:
+      return std::nullopt;
```
haved: Maybe just go with UNREACHABLE here, since any other predicates would have already caused an early return further up?

haved (commented):
Looks good, just some comments for your consideration

jlm/llvm/ir/operators/operators.cpp
phate: Done

jlm/llvm/ir/operators/operators.cpp
phate: Used.

jlm/llvm/ir/operators/operators.cpp
phate: Change it to logic_error

haved (approved):

PR #1766, opened 2026-07-28
Add simple caching to RegionPredicateTrace

haved:
Simple caching that stores the computed ranges for inputs to avoid re-traversing. The cache is not stored in the class between method calls, so it does not need to be invalidated.

Without this caching some of the files in the benchmark suite took tens of minutes. With it the slowest file is 6 minutes end-to-end (region predication gets added to SVF in the next PR)


phate (approved):
Looks good to me. It makes sense that it does not need to be invalidated. What I am wondering though. Would it help even more if we would make it as part of the class and could use it for multiple queries? We would need to deal with invalidation then.

PR #1769, opened 2026-07-29
LLVM: Fix and improve IOBarrier normalization

phate:
This PR fixes the problem that the number of loaded bytes myst be within the allocation site and adds more possible allocation sites.


phate:
@haved This PR is ready for review now.

haved (approved):

PR #1772, opened 2026-07-30
Add region predication checking to SVF

haved:
The logic is based on remembering whether or not we have followed any back-edges while tracing for origins of the loaded value. As long as we have never taken a back-edge, we can use the region predication check whenever we enter a gamma subregion. If the check tells us there is no way for the gamma subregion to be the origin, we skip it.

When we route values afterwards, we have to use the value origins found with loopBackEdgeTaken=true **if it exists**, but if it does not  exist it means tracing was able to finish without ever taking a back-edge and then going into the region / through the node. This lets us use the more precise value origin traced under the assumption that no back-edges have been followed.


jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -832,7 +1002,11 @@ StoreValueForwarding::processLoadWithMemoryStates(rvsdg::SimpleNode & loadNode)
   JLM_ASSERT(LoadOperation::numMemoryStates(loadNode) != 0);
 
   context_->statistics.startTracing();
-  LoadTracingInfo loadTracingInfo(loadNode, context_->outputTracer, context_->aliasAnalysis);
+  LoadTracingInfo loadTracingInfo(
+      loadNode,
+      context_->outputTracer,
+      context_->aliasAnalysis,
+      context_->regionPredicateTrace);
```
phate, line 1011: As far as I can see it, the region predication is only used for load nodes with states. It should also be used for loads without states IMO. That is not something for this PR, but we should put it on the TODO list.

jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -428,9 +442,12 @@ class LoadTracingInfo
     // Perform tracing from each memory state input to find exactly what store it leads to
     for (auto & memoryStateInput : LoadOperation::MemoryStateInputs(loadNode))
     {
-      // If the memory state input can not be traced back to store nodes,
-      // or different memory state inputs lead to different store nodes, we give up
-      auto lastValueOrigin = getLastValueOriginBeforeInput(memoryStateInput);
+      // Tracing starts at the load, so no loop back-edges have been taken yet
+      auto lastValueOrigin = getLastValueOriginBeforeInput(memoryStateInput, false);
+
+      // If the memory state input can not be traced back to value origins,
```
phate: cannot

jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -450,6 +468,57 @@ class LoadTracingInfo
     return true;
   }
 
+  /**
+   * After tracing has finished, provides the last value origin before the given \p node.
+   * @param node the node to trace backwards from
+   * @param loopBackEdgeMaybeTaken if true, the result of tracing with loopBackEdgeTaken=true
+   *        is attempted first, which is the most conservative traced value origin.
+   *        If \p node was never rached via any back-edges, the function falls back to false.
```
phate: reached

phate: I do not understand: "I node was never reached via any back-edges, the function falls back to false". What do you mean with "falls back to false".

Does it only mean that if you do not hand in true for the \p loopBackEdgeMaybeTaken parameter, then it assumes false? If so, it is a bool which has only two choices.

jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -450,6 +468,57 @@ class LoadTracingInfo
...
+    return std::nullopt;
+  }
+
+  /**
+   * After tracing has finished, provides the last value origin before the end of a \p region.
+   * Note that this value origin may be outside the region, if the region does not clobber.
+   * @param region the region whose results are traced backwards from
+   * @param loopBackEdgeMaybeTaken if true, the result of tracing with loopBackEdgeTaken=true
+   *        is attempted first, which is the most conservative traced value origin.
+   *        If \p region was never rached via any back-edges, the function falls back to false.
```
phate: reached

jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -1152,16 +1326,17 @@ StoreValueForwarding::getValueOriginOutput(
...
-      std::vector<rvsdg::Output *> lastStorePerSubregion;
+      std::vector<rvsdg::Output *> lastValueOriginPerSubregion;
       for (auto & subregion : gammaNode->Subregions())
       {
-        JLM_ASSERT(tracingInfo.lastValueOriginInRegion.count(&subregion));
-        auto lastValueOrigin = tracingInfo.lastValueOriginInRegion[&subregion];
-        auto & valueOriginOutput = getValueOriginOutput(lastValueOrigin, subregion, tracingInfo);
-        lastStorePerSubregion.push_back(&valueOriginOutput);
+        // If the gamma subregion has ever been traced with loopBackEdgeTaken = true, use it
+        auto lastValueOrigin = tracingInfo.getLastValueOriginInRegion(subregion, true);
```
phate, line 1340: Why do you hand automatically a true for the loop back-edge in here? It could be that the gamma subregion was traced with false, no?

jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -1176,40 +1351,29 @@ StoreValueForwarding::getValueOriginOutput(
...
+        JLM_ASSERT(lastValueOrigin->isKnown());
         auto & outerRegion = *thetaNode->region();
-        initialValue =
-            &getValueOriginOutput(lastStoreBeforeTheta->second, outerRegion, tracingInfo);
+        initialValue = &getValueOriginOutput(*lastValueOrigin, outerRegion, tracingInfo);
       }
       else
       {
-        // the loop variable is only used as a theta output, so give it undef as input
+        // Tracing never left reached the loop entry, so the value must be defined inside the loop.
```
phate: left reached

jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -1176,40 +1351,29 @@ StoreValueForwarding::getValueOriginOutput(
...
+        // The created loop variable can therefore take undef as its input.
+        // Since tracing never reached the top of the loop, the loop post was never added to the
+        // list of back-edges, and possibly never traced with loopBackEdgeTaken=true.
```
phate: Possibly, or definitely? The logic suggests definitely, the comment says possibly?

jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -1251,9 +1415,11 @@ StoreValueForwarding::connectUnroutedLoopPosts(LoadTracingInfo & tracingInfo)
     auto post = tracingInfo.unroutedLoopVarPosts.front();
     tracingInfo.unroutedLoopVarPosts.pop();
 
-    auto lastStore = tracingInfo.lastValueOriginInRegion.find(post->region());
-    JLM_ASSERT(lastStore != tracingInfo.lastValueOriginInRegion.end());
-    auto & origin = getValueOriginOutput(lastStore->second, *post->region(), tracingInfo);
+    // If the theta post has ever been traced with loopBackEdgeTaken=true, we use that
+    // otherwise fall back to the value origin reached with no back-edges taken.
+    auto lastValueOrigin = tracingInfo.getLastValueOriginInRegion(*post->region(), true);
```
phate, line 1430: The same here, the comment suggest loop back-edge is false, but then you hand in true and never try with false. It confuses me.

phate (dismissed):

jlm/llvm/opt/StoreValueForwarding.cpp
haved: Not quite sure how we would go about that in SVF. Partial redundancy elimination could benefit from predicate tracing, and that would help stateless loads

jlm/llvm/opt/StoreValueForwarding.cpp
haved: I rewrote the docstring param as so:
```cpp
/**
 * [...]
 * @param loopBackEdgeMaybeTaken if true, the function uses the most conservative value origin
 *        for the node, which was found without making any assumptions about loop back-edges.
 *        If false, or if tracing never reached the node after following a back-edge,
 *        the value origin found under the assumption that no back-edges have been taken is used.
```

jlm/llvm/opt/StoreValueForwarding.cpp
haved: Made the same change as above

jlm/llvm/opt/StoreValueForwarding.cpp
haved: I expanded the comment:
```cpp
// We only create one gamma exit variable for each load,
// so if tracing ever reached the gamma after following a back-edge,
// we can not use value origins traced under the assumption that no back-edges were taken.
// If the gamma output was never reached after tracing through a back-edge,
// the getter function will fall back to using value origins traced under the assumption
// that no back-edges have been followed, which is then a correct assumption.
```

jlm/llvm/opt/StoreValueForwarding.cpp
haved: While this comment is correct, it just confuses right here. Moved to `connectUnroutedLoopPosts`

jlm/llvm/opt/StoreValueForwarding.cpp
phate: Better

jlm/llvm/opt/StoreValueForwarding.cpp
haved: Updated:
```cpp
// We only create one loop variable per theta,
// so if tracing ever entered the theta after following a back-edge,
// we conservatively use the value origin found with loopBackEdgeTaken=true.
// If the theta subregion was never traced after following a back-edge,
// it falls back to using the value origin found assuming no back-edges have been followed.
```

phate (approved):

PR #1774, opened 2026-07-31
fix operator== in GEP to compare types correctly

haved:


jlm/llvm/ir/operators/GetElementPtr.cpp
phate, line 34: This entire thing is still not correct. You have the same problem for the argument types. They should also be compared by reference. Also, please add a unit test on it to ensure this never happens again.

phate (changes requested):

phate (approved):

PR #1778, opened 2026-07-31
Make node hoisting skip control constants

haved:
The existing code already tried to skip hosting them, but at a previous point in time we stopped treating control constants as state edges, so this check no longer does anything.

While everything still works with control constants hoisted, it becomes much harder to read the graphs, with no benefit.

### DIFF
--- a/jlm/llvm/opt/PushTests.cpp
+++ b/jlm/llvm/opt/PushTests.cpp
@@ -459,7 +459,8 @@ TEST(NodeHoistingTests, controlConstants)
   // Control constants must stay in their original regions and not be hoisted.
   // Bit constants (Int32) should be hoisted to the lambda subregion.
 
-  // Lambda subregion: theta node + gamma1 + gamma2 + CTL(0) (loop var entry) + two hoisted Int32 constants
+  // Lambda subregion: theta node + gamma1 + gamma2 + CTL(0) (loop var entry) + two hoisted Int32
+  // constants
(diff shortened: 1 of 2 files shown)


phate (changes requested):
I would love to see a unit test that documents/tests this behavior.

haved:
@phate added unit test

jlm/llvm/opt/PushTests.cpp
```diff
@@ -361,3 +363,126 @@ TEST(NodeHoistingTests, statefulOperations)
...
+  auto gamma1Exit = gamma1.AddExitVar({ &gamma1Ctrl1, &gamma1Ctrl0 });
+
+  // gamma2: takes gamma1's exit (control) as its predicate
+  auto & gamma2 = GammaNode::Create(*gamma1Exit.output, 2, {});
+
+  // gamma2 exit variable takes integer constants
+  auto & gamma2Int3 =
+      BitConstantOperation::create(*gamma2.subregion(0), BitValueRepresentation(32, 3));
+  auto & gamma2Int7 =
+      BitConstantOperation::create(*gamma2.subregion(1), BitValueRepresentation(32, 7));
```
phate: IntegerConstantOperation

phate (dismissed):

jlm/llvm/opt/PushTests.cpp
haved: fixed

haved:
@fixed your comment, need a review

phate (approved):

PR #1779, opened 2026-07-31
rhls-dne: remove unreachable check

caleridas:
rhls-dne IsDead argument check never triggers in practice. Even if it would trigger, the following code path would immediately cause UB (unconditionally deferences first user, which by definition cannot exist if the argument is dead).


caleridas:
To make progress on dependent type representation and (finally!)
publish the coq bindings, I need to address inputs, outputs & types
in a specific way. The present structure of hls loopnode is a
show-stopper for this at present, so I will try to address it -- effectively,
I need to remove all custom-typed RegionArgument / RegionResult
for good, and finally. Since the precise usage of the hls structures
is a little bit involved, I will try to do it piecemeal and carefully --
please check the individual bits.

phate (approved):

PR #1793, opened 2026-08-10
RVSDG: add node id to debug string of nodes

phate:


haved:
@phate some tests relied on the old printing

phate:
Creates too many problems.

PR #1796, opened 2026-08-11
LLVM: Overhaul store-store reduction

phate:


jlm/llvm/ir/operators/Store.cpp
```diff
@@ -218,14 +168,48 @@ StoreNonVolatileOperation::NormalizeStoreMux(
...
 std::optional<std::vector<rvsdg::Output *>>
-StoreNonVolatileOperation::NormalizeStoreStore(
-    const StoreNonVolatileOperation & operation,
+StoreNonVolatileOperation::normalizeStoreStore(
+    const StoreNonVolatileOperation & store2Op,
     const std::vector<rvsdg::Output *> & operands)
 {
-  if (is_store_store_reducible(operation, operands))
-    return perform_store_store_reduction(operation, operands);
+  JLM_ASSERT(operands.size() > 2);
```
haved, line 182: I'm a bit surprised that this assert never triggers. It is possible for a program to contain a store to a NULL pointer, as long as control flow never actually reaches the store. Could we abort early instead if the store doesn't have any state edges?

haved (dismissed):
Looks good, I have one comment @phate

jlm/llvm/ir/operators/Store.cpp
phate: Yeah, makes sense. I will fix it.

phate: I added an early return. I also wanted to add a test for it, but since the store node does not have any state, it means also that it does not have any output that I can connect to such that it survives the region prune. This might also be the reason why we do not see it in practice, as it is simply pruned away because it must be a dead operation.

phate:
@haved I addressed your comment.

haved (approved):

PR #1797, opened 2026-08-12
LLVM: Fix store-store reduction

phate:
It might happen that the upper store in the store-store chain has more memory states than the lower store even though they both load from the same address. This might occur if encoding was (at the time) unable to determine the minimal set of states for the upper store. The old code simply assumed that both stores had the same number of memory states, which lead to an assert being triggered when the reduction was applied. An example RVSDG looks as follows:

image

Both stores ultimately load from the same alloca node, but this can only be determined after some node reductions were applied.

### DIFF
--- a/jlm/llvm/ir/operators/Store.cpp
+++ b/jlm/llvm/ir/operators/Store.cpp
@@ -196,14 +196,22 @@ StoreNonVolatileOperation::normalizeStoreStore(
     return std::nullopt;
 
   // Check that all memory state inputs originate from store1 AND have no other users
+  std::vector<rvsdg::Output *> newMemoryStates;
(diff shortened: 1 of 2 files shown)


phate:
@haved Looking a second time at the graph, I am wondering whether SVF is able to forward the value of the store node to the load. Both of them go to the same address, but have different number of memory states.

haved:
As long as all memory states from a load reach the same store, SVF can do forwarding.

In this exact program it looks to me as if the loaded address is taken from a GEP with a non-constant offset, while the store targets the alloca directly. I think that might be disqualifying them from AA.

haved (approved):

PR #1803, opened 2026-08-12
SVF: add support for truncating integers

phate:


jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -1157,7 +1157,23 @@ copyDeltaElement(
...
+            return *copiedOutput;
+          }
+
+          if (loadBitType->nbits() < copiedBitType->nbits())
+          {
+            return *TruncOperation::createNode(*copiedOutput, loadedType).output(0);
+          }
+
+          // FIXME: In this case, we would need to concat multiple integers.
+          return *copiedOutput;
```
haved, line 1176: The fixme implies we give up on forwarding, but you are still returning the copied output? Has it already been determined previously that we are not forwarding from a smaller type? Should we have an UNREACHABLE here instead then?

haved (commented):
@phate I have one comment I don't understand

jlm/llvm/opt/StoreValueForwarding.cpp
phate: No, this is correct, but the only reason why this currently works is because there is another test in the caller that checks whether the type is different, and if so, gives up. I need to return something in this function without failing.

haved (approved):

PR #1810, opened 2026-08-14
LLVM: add support for llvm.assume and llvm.expect intrinsics

phate:


haved:
@phate since the expect intrinsic requires actually being converted perhaps it would be nicer to not have it be "ignored", by `shouldIgnoreIntrinsic`. That way we can avoid the doubled up switch case. Otherwise we should at least have an assert that the intrinsic is not supposed to be ignored before creating a call for it

phate:
@haved We are not converting the expect intrinsic. We get the value of its first operand and return that (which the intrinsic returns). I see no value to carry this intrinsic as node in the RVSDG around as long as we do not utilize it.

haved (approved):

PR #1831, opened 2026-08-16
LLVM: add support for llvm.ssub.with.overflow intrinsic

phate:


jlm/llvm/frontend/LlvmModuleConversion.cpp
```diff
@@ -1412,6 +1412,19 @@ converSAddWithOverflowIntrinsic(
   return tacs.back()->result(0);
 }
 
+static const Variable *
+converSSubWithOverflowIntrinsic(
```
haved: Same typo here. Is it being copied from somewhere?

haved (commented):
@phate same typo again. Is it copied from somewhere?

jlm/llvm/frontend/LlvmModuleConversion.cpp
phate: Fixed all occurences of this typo (I hope)

haved (dismissed):

haved (approved):

PR #1837, opened 2026-08-17
LLVM: add support for llvm.ptrmask intrinsic

phate:


jlm/llvm/ir/operators/GeneralIntrinsicOperations.cpp
```diff
@@ -29,4 +29,56 @@ IsConstantOperation::copy() const
...
+    {
+      throw std::runtime_error(util::strfmt(
+          "PtrMaskOperation::checkOperandTypes: Expected mask type to be a "
+          "vector of integers of size ",
+          vectorType->size(),
+          "."));
+    }
+  }
+
+  throw std::runtime_error(
```
haved, line 80: This always throws, there is no else, nor early return in the function

haved (commented):
The check types always throws @phate

jlm/llvm/ir/operators/GeneralIntrinsicOperations.cpp
phate: Yeah, it is early. Fixed.

haved (approved):

PR #1840, opened 2026-08-18
HLS: Transformation for dumping graph in different formats

sjalander:
A new DumpTransformation is introduced that can be configured to use existing infrastructure to dump an RVSGD graph in different formats.


phate (changes requested):
Can separate the LLVM parts of this PR from the HLS parts. Let's get the LLVM parts in first, then we can deal with the HLS parts, please.

PR #1841, opened 2026-08-19
LLVM: add IOBarrierElimination pass

phate:


jlm/llvm/opt/IOBarrierElimination.cpp
```diff
@@ -0,0 +1,239 @@
...
+            return;
+
+          const auto & barredAddressOperand =
+              *IOBarrierOperation::BarredInput(*ioBarrierNode).origin();
+          auto sizeOpt = context_->isDereferenceable(barredAddressOperand);
+          const auto sizeInBytes = GetTypeStoreSize(*loadOperation.GetLoadedType());
+          if (!sizeOpt.has_value() || sizeOpt.value() < sizeInBytes)
+            return;
+
+          removeIOBarrierNode(*ioBarrierNode);
```
haved: It might not be correct to remove the entire IOBarrier if it has multiple users. We should divert the one load we know we can redirect.

haved (commented):
Looks good, I have one comment @phate

jlm/llvm/opt/IOBarrierElimination.cpp
phate: Great catch! I added a unit test for this case and redesigned the sweep stage.

haved (approved):

PR #1842, opened 2026-08-19
LLVM: Adds RenderAnnotatedTree method for printing statistics

sjalander:
Makes it possible to print tree information other than as a transfromation.

### DIFF
--- a/jlm/llvm/opt/RvsdgTreePrinter.cpp
+++ b/jlm/llvm/opt/RvsdgTreePrinter.cpp
@@ -17,6 +17,49 @@
 namespace jlm::llvm
 {
 
+std::string
+RvsdgTreePrinter::RenderAnnotatedTree(const rvsdg::Graph & graph, const rvsdg::Region & region)
+{
+  util::AnnotationMap annotationMap;
+
+  auto matchAll = [](const rvsdg::Node &)
+  {
+    return true;
+  };
+
+  AnnotateDebugIds(graph, annotationMap);
+  AnnotateNumNodes(graph, matchAll, "NumRvsdgNodes", annotationMap);
+
+  auto isAggAlloca = [](const rvsdg::Node & node) -> bool
+  {
+    const auto * op = dynamic_cast<const AllocaOperation *>(&node.GetOperation());
+    return op != nullptr && IsAggregateType(*op->allocatedType());
+  };
+  AnnotateNumNodes(graph, isAggAlloca, "NumAggregateAllocaNodes", annotationMap);
+
+  auto isAlloca = [](const rvsdg::Node & node) -> bool
+  {
+    return rvsdg::is<AllocaOperation>(&node);
+  };
+  AnnotateNumNodes(graph, isAlloca, "NumAllocaNodes", annotationMap);
+
+  auto isLoad = [](const rvsdg::Node & node) -> bool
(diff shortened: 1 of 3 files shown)


jlm/llvm/opt/RvsdgTreePrinter.cpp
```diff
@@ -17,6 +17,49 @@
...
+
+  auto isStore = [](const rvsdg::Node & node) -> bool
+  {
+    return rvsdg::is<StoreOperation>(&node);
+  };
+  AnnotateNumNodes(graph, isStore, "NumStoreNodes", annotationMap);
+
+  AnnotateNumMemoryStateInputsOutputs(graph, annotationMap);
+
+  return rvsdg::Region::ToTree(region, annotationMap);
```
phate, line 60: All these is* lambdas are duplicated from the code that is already there. Refactor them in individual static methods in this file and use the respective version at both locations, so we do not have so much code duplication.

jlm/llvm/opt/RvsdgTreePrinter.hpp
```diff
@@ -129,6 +129,23 @@ class RvsdgTreePrinter final : public rvsdg::Transformation
...
   Run(rvsdg::RvsdgModule & rvsdgModule, util::StatisticsCollector & statisticsCollector) override;
 
+public:
+  /**
+   * Render an annotated tree string for the given \p region with all available annotations.
+   *
+   * This convenience method applies every annotation type to the RVSDG and produces a
+   * human-readable indented tree output via \ref rvsdg::Region::ToTree. It encapsulates the full
+   * annotation-compute-and-render pipeline so callers do not need to invoke individual annotator
+   * functions or handle the intermediate \ref AnnotationMap.
```
phate, line 139: The second sentence can be dropped. Does not make sense as part of the documentation I think.

jlm/llvm/opt/RvsdgTreePrinter.hpp
```diff
@@ -129,6 +129,23 @@ class RvsdgTreePrinter final : public rvsdg::Transformation
...
+   *
+   * @param graph The RVSDG graph (required because annotation computation walks the entire region
+   *              tree).
+   * @param region The region whose annotated tree to render.
+   * @return A string containing the fully annotated tree.
+   */
+  [[nodiscard]] static std::string
+  RenderAnnotatedTree(const rvsdg::Graph & graph, const rvsdg::Region & region);
```
phate, line 147: The graph parameter makes no sense here. The proper solution is to refactor the AnnotateNum*() methods to take a region instead of the graph. The only thing these methods do is to take the root region from the graph. Thus, by changing their parameter to a region and handing in the root region to them, you can simply drop this entire graph parameter here and you get a clean interface.

jlm/llvm/opt/RvsdgTreePrinterTests.cpp
```diff
@@ -273,3 +275,137 @@ TEST(RvsdgTreePrinterTests, printAllocaNodes)
 
   EXPECT_EQ(tree, expectedTree);
 }
+
+/* ------------------------------------------------------------------------ */
```
phate, line 279: Remove

phate (changes requested):

PR #1862, opened 2026-09-02
SVF: fix bug in predicated forwarding

phate:


haved:
@phate closing this as we found the actual cause

PR #1859, opened 2026-08-31
IVR: Add control constant redirection

phate:

Issue #1764, opened 2026-07-28
Predicate Tracing can yield incorrect results


haved:
This was found by reducing a miscompilation of embench's picojpeg. The graph we end up with looks like this:
<img width="1048" height="772" alt="Image" src="https://github.com/user-attachments/assets/2f8bf006-3ddc-4244-8142-ff2c018eaf3e" />

The control constants were originally inside the inner gamma, but were pushed put by the `NodePush` pass.

When we query the region predication tracer about the highlighted origin region and target region, it responds `false`, which is wrong. The class correctly sees that the predicate to the lower gamma must be 1, and goes through the graph to see what values the inputs that may end up as the predicate may have. I have written the set of possible values in red for each input.

The class also creates a set of possible values assigned to the predicate **per region**. This per region set is not the union of all the inputs in the region, or the region results, but rather the first input in the region that finishes its set. The set for the `origin region` ends up being {0}, since the first input that finishes its set of possible values is the leftmost input to the inner gamma.

The fix for this is to take the union of all inputs (or all region results) in the region when determining what values the region can provide for the predicate.

----

While thinking about the fix for the above case, I started to consider cases like the one shown here:
<img width="486" height="701" alt="Image" src="https://github.com/user-attachments/assets/f8e2d98c-001a-4709-98ea-9a55643e04a4" />

When asked about the highlighted target region and origin region, the predicate tracer is going to respond `false`. It sees that the predicate of the lowest gamma must be 1. When tracing the possible origins of the predicate, tracing will reach the origin region, in which the only possible value is 0. However, if the middle gamma evaluates the rightmost subregion, it does not matter that the origin region produces a 0. It is still possible for control flow to go from the origin to the target region.

Before we consider doing lots of funky logic for handling thins, I am wondering how common situations like the one above really are. For example, if it is rare for a gamma to take control values as entry variables, we can work around the above issue by simply never tracing out of gamma nodes.

Whenever I have seen control typed values being routed into gammas, it has been because of NodePushOut hoisting control constants. If we tell NodePushOut to not move control constants, how many situations are left where control inputs are routed into gammas as entry variables?

I'll make a PR for fixing the first part of this issue, and then see if anything is still going wrong.

FYI @caleridas @phate

haved:
[...]

PR #1868, opened 2026-09-04
Write alternative region predication checker with fix + theta tracing

haved:
The fix is demonstrated in the added test, where a commented out line at the bottom shows how the old predicate tracer will answer incorrectly. The issue is the same as reported in issue #1764, we were just unsure if the conditions needed to trigger the issue could ever occur in practice, until @phate's IVR improvements in PR #1859 triggered it.

The fix itself separates tracing for finding definite values from tracing for finding sets of possible values. The implementation also differs in what it caches, to support theta predicate tracing. When traversing up the region hierarchy from the origin region and leaving a theta node, we know that the theta's predicate must be 0 in order to leave. This gives more precise tracing when the target region is outside a theta.

The new tracer does not register any observers to handle automatic cache invalidation. I did this because I realized there were anyways issues with the existing observers, where they would not react to regions being deleted, which means new regions can re-use their address and pick up stale cache values. Instead of adding even more region observers, I elected to not support automatic invalidation. SVF does not affect control flow, and if some pass really needs it later, I think it in any case makes more sense to have a general `GraphObserver` instead of adding individual region observers.

With the alternative tracer used in SVF, we get 1774 fewer loads than on master, across 438 files. 18 files do worse (by 1-7 loads), which is as far as I can tell due to the fix, so the loads might not have been legal to remove in the first place.

FYI @caleridas @phate

### DIFF
--- a/jlm/rvsdg/RegionPredicateTrace.hpp
+++ b/jlm/rvsdg/RegionPredicateTrace.hpp
@@ -9,6 +9,7 @@
 #include <jlm/rvsdg/control.hpp>
 #include <jlm/rvsdg/graph.hpp>
 #include <jlm/rvsdg/node.hpp>
+#include <jlm/util/HashSet.hpp>
 
 #include <unordered_map>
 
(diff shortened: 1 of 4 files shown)


jlm/rvsdg/RegionPredicateTrace.hpp
```diff
@@ -235,6 +236,101 @@ class RegionPredicateTrace
...
+
+  /**
+   * Determines if it is possible for control flow to go from the given
+   * \p originRegion to the given \p targetRegion,
+   * without following any back-edges in any theta node that contains the \p originRegion.
+   * No assumptions are made about theta nodes surrounding the \p targetRegion,
+   * or control flow entering theta nodes between the two regions.
+   */
+  [[nodiscard]] bool
+  canRegionReachRegion(Region & originRegion, Region & targetRegion);
```
phate: I really do not like the name....what about `isReachableFrom()`?

phate (commented):
I really have not much to say about it. The implementation looks clean, very readable. The biggest "problem" I have is with a name, otherwise all good for me.

haved:
@phate I did the renaming

caleridas:
My only comment is why "Alternative..." -- maybe just replace the existing one, and be done with it? I do not think there is too much benefit in having two tracers, the old one should go away. (Unless your intent is to do this in a follow-up PR, fine with that).

caleridas (approved):

PR #1873, opened 2026-09-09
LLVM: add debug information to emitted LLVM IR

phate:


jlm/llvm/backend/IpGraphToLlvmConverter.cpp
```diff
@@ -74,6 +91,18 @@ class IpGraphToLlvmConverter::Context final
...
   }
 
+  [[nodiscard]] ::llvm::DIBuilder &
+  getDIBuilder() noexcept
+  {
+    return diBuilder_;
+  }
+
+  [[nodiscard]] ::llvm::DIFile *
+  di_file() const noexcept
```
phate: Change

haved (dismissed):

phate:
@haved @sjalander Can we disable the mergesort_decouple tests. I pushed now 3-4 golden updates in a row and the cycles ping forth and back for these benchmarks. The cycle output is clearly indeterministic and something is just horribly off there.

sjalander:
It would be good to document the problem. We could comment them out in usr/hls-test-suite/Makefile.sub and maybe add a note that references this specific PR, as it shows the problematic behavior.

PR #1874, opened 2026-09-09
LLVM: add RvsdgNodeLocation support to ThreeAddressCode class

phate:


jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
```diff
@@ -264,10 +264,12 @@ RvsdgToIpGraphConverter::ConvertSimpleNode(const rvsdg::SimpleNode & simpleNode)
     operands.push_back(Context_->GetVariable(simpleNode.input(n)->origin()));
 
   auto operation = simpleNode.GetOperation().copy();
-  Context_->GetLastProcessedBasicBlock()->append_last(ThreeAddressCode::create(
+  auto tac = ThreeAddressCode::create(
       std::unique_ptr<rvsdg::SimpleOperation>(
           util::assertedCast<rvsdg::SimpleOperation>(operation.release())),
-      operands));
+      operands);
+  tac->setRvsdgNodeLocation({ simpleNode.region()->getRegionId(), simpleNode.GetNodeId() });
```
haved: Since all the calls to this function all take the pattern`({ node.region()->getRegionId(), node.GetNodeId() })`, should we have an overload of the function that just takes a Node& and extracts this information itself?

haved (dismissed):
One suggestion @phate, otherwise looks good

jlm/llvm/backend/RvsdgToIpGraphConverter.cpp
phate: Great catch. I changed the interface of the method. No need to offer two if only one is used.

haved (approved):

PR #1876, opened 2026-09-10
LLVM: extend PtrCmpOperation::normalizeNullPointerComparison()

phate:
1. Fixes a bug in the normalization where we stopped tracing intraprocedural. This was not sufficient as we would also like to catch global allocation sites (i.e. delta nodes)
2. Extends the normalization for comparisons between ConstantPointerNullOperations


jlm/llvm/ir/operators/operators.cpp
```diff
@@ -328,6 +328,21 @@ PtrCmpOperation::normalizeNullPointerComparison(
   auto & tracedOperand1 = rvsdg::traceOutputIntraProcedurally(*operands[0]);
   auto & tracedOperand2 = rvsdg::traceOutputIntraProcedurally(*operands[1]);
 
+  if (isOutputOf<ConstantPointerNullOperation>(tracedOperand1)
```
phate, line 331: Blocking

haved (approved):

PR #1877, opened 2026-09-10
add mayEnterSubregion to trace functions

haved:
internally creates there different policies for how to handle structural nodes. This PR only adds the ability to enter theta subregions, leaving gamma nodes for later. It also cleans up the caching of invariant subregion outputs a bit, and explains exactly what is cached.

 I decided against giving `mayEnterSubregion` parameter a default, because I think all users should make a conscious decision. Since tracing into subregions can return outputs inside sibling regions, all users send in false for now.

The only users of the tracer that sets `mayEnterSubregion` to `true` in this PR is the `tryGetConstantSignedInteger`.

Tests are added for both tracing into a theta, and for getting an Integer constant out of a theta.

This PR does not lead to any fewer loads.


jlm/rvsdg/Trace.hpp
```diff
@@ -27,38 +25,81 @@ class ThetaNode;
...
+    // If so, tracing continues from the corresponding input to the structural node.
+    traceThroughTriviallyInvariant = 0,
+
+    // Performs the same check as above, but tries harder to determine invariance.
+    // Tracing continues inside the subregion, to check if it can reach a subregion argument.
+    // If all subregions reach the same argument, the structural node is invariant,
+    // and tracing continue from the input of the structural node.
+    traceThroughIfDetectedInvariant = 1,
+
+    // Enables tracing ending up inside subregions of structural nodes.
```
phate: "Enables tracing into subregions of structural nodes."

jlm/rvsdg/Trace.hpp
```diff
@@ -86,26 +127,50 @@ class OutputTracer
...
+  isInvarianceCachingEnabled() const noexcept
   {
-    return isInterprocedural_;
+    return enableInvarianceCaching_;
   }
 
   /**
-   * Enables or disables interprocedural tracing.
-   * @see isInterprocedural()
+   * Enables or disables caching the fact that structural outputs are invariant.
```
phate: "Enables or disables invariant structural output caching."

jlm/rvsdg/Trace.hpp
```diff
@@ -176,39 +234,44 @@ class OutputTracer
   traceStep(Output & output, const rvsdg::Region * withinRegion);
 
   /**
-   * Inserts a traced value into the tracing cache.
+   * Inserts the given \p structuralOutput in the invariance cache.
+   * The output needs to be a copy of one of the inputs of the structural node,
+   * which is given as \p structualInput.
```
phate: I fail to understand this sentence: "a copy of one of the inputs of the structural node"?

jlm/rvsdg/Trace.hpp
```diff
@@ -176,39 +234,44 @@ class OutputTracer
...
    *
-   * @param output The output that was traced.
-   * @param traceResult The input at which the tracing arrived.
-   * @return The origin of \p traceResult for convenience.
+   * @param structuralOutput The structural output that was traced.
+   * @param structuralInput The corresponding structural input, or nullptr if it isn't invariant.
```
phate: structuralInput can never by a nullptr. It is a reference.

jlm/rvsdg/Trace.hpp
```diff
@@ -176,39 +234,44 @@ class OutputTracer
...
-  insertInCache(const Output & output, Input * traceResult);
+  insertInInvarianceCache(const Output & structuralOutput, Input & structuralInput);
 
   /**
-   * Loops up the tracing result for \p output.
+   * Performs a lookup in the invariance cache.
    *
-   * @param output The output that is traced.
-   * @return The output at which tracing arrived, otherwise std::nullopt.
+   * @param structuralOutput the output looked up in the cache.
```
phate: the output to look up in the cache.

jlm/rvsdg/Trace.hpp
```diff
@@ -176,39 +234,44 @@ class OutputTracer
...
+   * @return the corresponding structural input the output is a copy of,
```
phate: I fail to understand this copy principle you are using here.

phate: I guess you mean: "the corresponding structural input that represents the same value as output" or something like this?

jlm/rvsdg/Trace.hpp
```diff
@@ -176,39 +234,44 @@ class OutputTracer
...
 
   // When true, tracing is allowed to continue outside of lambda nodes.
   // When false, tracing will stop at the lambda's context arguments.
   bool isInterprocedural_ = true;
 
-  // When true, tracing is allowed to cache traced values.
-  bool enableCaching_;
-  std::unordered_map<const Output *, Input *> traceCache_{};
+  // When true, tracing can go from the output of a Phi node into its subregion.
+  // When false, tracing will stop at the Phi output.
```
phate: at the output of a \ref PhiNode.

jlm/rvsdg/Trace.hpp
```diff
@@ -176,39 +234,44 @@ class OutputTracer
...
+  bool enterPhiNodes_ = true;
+
+  // When true, the tracer can cache the fact that outputs of structural nodes are invariant.
+  // Enabling caching also means you must take care to manually invalidate the cache.
```
phate: Enabling caching means that the user of the tracer is responsible for cache invalidation.

jlm/rvsdg/Trace.cpp
```diff
@@ -71,16 +70,19 @@ OutputTracer::tryTraceThroughGamma(GammaNode & gammaNode, Output & output)
   {
     auto tracedInner = branchResult->origin();
 
-    // If deep tracing is enabled, make a greater effort to trace up to a region argument
-    if (traceThroughStrucutalNodes_)
+    if (structuralNodePolicy_ >= StructuralNodePolicy::traceThroughIfDetectedInvariant)
```
phate: Uff, this smells like future bugs. You create a dependency on the ordering in the enum here. Either you very explicitly say in the enum that the ordering is important and should not be changed there, or what I prefer, you make this more explicit here to avoid this dependency.

jlm/rvsdg/Trace.cpp
```diff
@@ -93,29 +95,29 @@ OutputTracer::tryTraceThroughGamma(GammaNode & gammaNode, Output & output)
...
   }
 
   const auto loopVar = thetaNode.MapOutputLoopVar(output);
 
   auto tracedInner = loopVar.post->origin();
 
-  // If deep tracing is enabled, make a greater effort in tracing up to a region argument
-  if (traceThroughStrucutalNodes_)
+  // If invariance detection is enabled, perform tracing inside the subregion
+  if (structuralNodePolicy_ >= StructuralNodePolicy::traceThroughIfDetectedInvariant)
```
phate: same here

jlm/rvsdg/Trace.cpp
```diff
@@ -134,11 +136,19 @@ OutputTracer::tryTraceThroughTheta(ThetaNode & thetaNode, Output & output)
...
     {
-      return insertInCache(output, originLoopVar.input);
+      return insertInInvarianceCache(output, *originLoopVar.input);
     }
   }
 
-  return insertInCache(output, nullptr);
+  // If we are allowed to return outputs from inside the subregion,
+  // return the result from tracing inside the subregion
+  if (structuralNodePolicy_ >= StructuralNodePolicy::traceIntoSubregions)
```
phate: same here

jlm/llvm/opt/StoreValueForwarding.cpp
```diff
@@ -191,11 +191,16 @@ class StoreValueForwarding::Statistics final : public util::Statistics
...
   explicit Context(aa::AliasAnalysis & aliasAnalysis, Statistics & statistics) noexcept
-      : outputTracer(true),
+      : outputTracer(),
         aliasAnalysis(aliasAnalysis),
         statistics(statistics)
   {
-    outputTracer.setTraceThroughStructuralNodes(true);
+    // It is safe to do invariance caching.
+    //  - outputs that are invariant in a structural node are never changed inside the node.
+    //  - structural node inputs and outputs are never deleted during the duration of this pass.
```
phate: removed during the duration of this pass

jlm/rvsdg/Trace.hpp
haved: Changed it to: @return the corresponding structural input the invariant output gets its value from.

jlm/rvsdg/Trace.cpp
haved: I clarified the comment on the enum, and added helper getters:   `bool structuralNodePolicyAllowsDeepInvarianceChecking()` and `bool structuralNodePolicyAllowsTracingIntoSubregions()`. It is a bit verbose, but makes it very clear what is being checked

jlm/rvsdg/Trace.cpp
haved: see above

jlm/rvsdg/Trace.cpp
haved: see above

jlm/rvsdg/Trace.hpp
```diff
@@ -27,38 +25,109 @@ class ThetaNode;
...
+  /**
+   * Checks if the current \ref StructuralNodePolicy allows checking for invariant outputs
+   * of structural nodes by performing tracing inside the subregion(s) of the node.
    *
-   * @param enableCaching Determines whether the output tracer should cache traced results.
+   * @return true if the policy allows deep invariance checking, false otherwise
    */
-  explicit OutputTracer(bool enableCaching) noexcept;
+  [[nodiscard]] bool
+  structuralNodePolicyAllowsDeepInvarianceChecking() const noexcept
```
phate: How about: isDeepInvarianceCheckingEnabled()

jlm/rvsdg/Trace.hpp
```diff
@@ -27,38 +25,109 @@ class ThetaNode;
...
-   * @return true if tracing though the subregions of structural nodes is enabled.
+   * Checks if the current \ref StructuralNodePolicy allows tracing to enter subregions,
+   * and returning value origins that are inside subregions or sibling regions of the region
+   * where tracing started.
+   *
+   * @return true if the policy allows tracing to enter subregions, false otherwise.
    */
   [[nodiscard]] bool
-  isTracingThroughStructuralNodes() const noexcept
+  structuralNodePolicyAllowsTracingIntoSubregions() const noexcept
```
phate: isTraceingIntoSubregionsEnabled()

phate (dismissed):

phate (approved):

haved:
@phate took your suggestions

PR #1879, opened 2026-09-11
LLVM: add constant folding for FBinaryOperation

phate:


jlm/llvm/ir/operators/OperatorsTests.cpp
```diff
@@ -305,4 +305,79 @@ TEST(FCmpOperationTests, testFoldConstants)
...
+      break;
+    default:
+      FAIL() << "Unhandled fpop in test";
+    }
+
+    EXPECT_TRUE(constantOperation->constant().bitwiseIsEqual(expected))
+        << "fpop " << static_cast<int>(op);
+  };
+
+  expectFoldedTo(fpop::add, c1.output(0), c2.output(0));
```
haved: Why is the expected result not sent in as a parameter to this function? That would obviate the switch inside the function matching on operation

haved (dismissed):
One suggestion for the test, and his CI fails @phate

jlm/llvm/ir/operators/OperatorsTests.cpp
phate: Fixed.

haved (dismissed):

haved (dismissed):

haved (approved):

PR #1886, opened 2026-09-13
MLIR: Handle context variables for delta

sjalander:


jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -1012,6 +1012,12 @@ MlirToJlmConverter::ConvertOperation(
             mlirDeltaNode.getConstant(),
             4)); // FIXME: the MLIR delta node does not support the alignment attribute
 
+    // Add context variables from inputs to delta's subregion
+    for (size_t i = 0; i < inputs.size(); ++i)
```
phate: for (auto & input : inputs) does not work?

phate (dismissed):

jlm/mlir/frontend/MlirToJlmConverter.cpp
sjalander: Changed

phate (approved):

PR #1891, opened 2026-09-14
MLIR: Handle linkage for LambdaNode in frontend

sjalander:


haved (dismissed):

jlm/mlir/frontend/MlirToJlmConverter.cpp
```diff
@@ -1213,14 +1213,18 @@ MlirToJlmConverter::ConvertLambda(
...
-  // The linkage should be part of the MLIR attributes so it can be extracted here
+  // Get the linkage attribute from the MLIR LambdaNode
+  auto linkage = llvm::Linkage::externalLinkage; // Default to external linkage
+  auto linkageAttribute = mlirOperation.getAttr(::llvm::StringRef("linkage"));
+  if (linkageAttribute != nullptr)
+  {
+    auto linkageStr = ::mlir::cast<::mlir::StringAttr>(linkageAttribute);
+    linkage = llvm::linkageFromString(linkageStr.str());
+  }
+
```
phate, line 1224: I do not like the default here. The linkage attribute should always be present, otherwise it is an error, no?

sjalander: @phate 
The LlvmLambdaOperation does not enforce that linkage is included when created.
Should we instead have a check if it is a nullptr and then create a LlvmLambdaOperation that doesn't have linkage information?
Or should the LlvmLambdaOperation be required to provide linkage information?

phate: Looking at the constructor of the LlvmLambdaOperation, it enforces that the linkage information must be present: https://github.com/phate/jlm/blob/efa222d0665245cec59e96fbb820fecb8d53678d/jlm/llvm/ir/operators/lambda.hpp#L34-L39

sjalander: My fault, it was the MLIR tests that didn't include the linkage.
The code now asserts if there is no linkage information.

phate (approved):

PR #1892, opened 2026-09-14
Add region predication checking to the OutputTracer

haved:
Makes us of the AlternativeRegionPredicationTracer to skip regions during tracing when the region provably is not the origin of the output being traced.

See the two added tests for examples. The design for only doing predication checks when we know that no back-edges have been taken is very similar to the one used in SVF, but I used an enum with two states instead of a boolean to make it clearer.

In this PR, the region predication checking is only enabled whenever the helper functions are given mayEnterSubregion=true. Later PRs will enable it in more places, which has a nice effect on number of loads removed. This PR does not affect number of loads much.


jlm/rvsdg/Trace.hpp
```diff
@@ -154,6 +155,18 @@ class OutputTracer
...
   }
 
+  [[nodiscard]] bool
+  isRegionPredicateCheckingEnabled() const noexcept
+  {
+    return regionPredicateChecking_;
+  }
+
+  void
+  setRegionPredicateCheckingEnabled(bool value) noexcept
```
phate, line 165: enableRegionPredicateChecking()

jlm/rvsdg/Trace.hpp
```diff
@@ -217,61 +230,108 @@ class OutputTracer
...
+  {
+    // Tracing has gone from the starting output to the current output without
+    // following any back-edges around the current output.
+    // Theta nodes between the current output and the starting output do not matter.
+    NotTakenAroundCurrentOutput,
+
+    // While tracing from the starting output to the current output,
+    // the tracer may have followed a back-edge going around the current output.
+    // This prevents the use of the region predication checker.
+    PossiblyTakenAroundCurrentOutput,
```
phate: Could we not call them:
1. NoBackEdgeTaken
2. PossiblyBackEdgeTaken

I am not sure what exactly you even mean by "not taken around current output", even with me having more context than other people and the comment to it.

jlm/rvsdg/Trace.hpp
```diff
@@ -295,10 +355,22 @@ class OutputTracer
   // When false, tracing will stop at the output of the phi node.
   bool enterPhiNodes_ = true;
 
+  // When true, gamma subregions are ignored when it is impossible for control flow to go
+  // from the gamma subregion to the region containing the output tracing started from
+  bool regionPredicateChecking_ = false;
```
phate: `enableRegionPredicateChecking_`

jlm/rvsdg/TraceTests.cpp
```diff
@@ -473,3 +475,213 @@ TEST(TraceTests, ThetaCachingTest)
...
+  auto & gammaCtrl1 = ControlConstantOperation::createTrue(*gammaNode.subregion(1));
+  auto & gammaInt5 = BitConstantOperation::create(*gammaNode.subregion(1), { 32, 5 });
+
+  auto gammaCtrlExit = gammaNode.AddExitVar({ &gammaCtrl0, &gammaCtrl1 });
+  auto gammaExit1 =
+      gammaNode.AddExitVar({ gammaEntry1.branchArgument[0], gammaEntry1.branchArgument[1] });
+  auto gammaExit2 = gammaNode.AddExitVar({ gammaEntry2.branchArgument[0], &gammaInt5 });
+  auto gammaExit3 = gammaNode.AddExitVar({ &gammaInt4, gammaEntry3.branchArgument[1] });
+
+  // Diver theta results
```
phate: Divert theta results

jlm/rvsdg/Trace.cpp
```diff
@@ -24,15 +24,38 @@ OutputTracer::trace(Output & output)
 }
 
 Output &
-OutputTracer::trace(Output & output, const rvsdg::Region * withinRegion)
+OutputTracer::trace(Output & output, const Region * withinRegion)
+{
+  // PERF: Querying the region predication checker class causes repeated re-traversal of the same
```
phate: Add a FIXME here so that it is findable again. We do not really have precedence for the keyword PERF

jlm/rvsdg/Trace.hpp
haved: Are you sure? There is already a pattern of setters starting with set.

haved:
@phate fixed your comments, except for the one with the setter since it goes against the established pattern in the class

phate (approved):

PR #1898, opened 2026-09-17
Add region predication to output tracer

haved:
Also changes the return type of the traceStep functions to let the function declare that the returned output is done, and should not be traced any more.

This PR was previously merged and then reverted because the previous version would re-try tracing for as long as the output changed. This could cause exponential re-trying in the case of deeply nested gammas. When a gamma was found where only one subregion could provide the value, it would fully trace the origin within the subregion, and then fully trace it once again. With multiple tracing loops inside one another this explodes.


jlm/rvsdg/Trace.hpp
```diff
@@ -217,61 +230,170 @@ class OutputTracer
...
+  [[nodiscard]] TraceStepResult
+  traceThetaArgument(ThetaNode & thetaNode, Output & output);
 
   /**
    * The innermost body of the tracing loop. Should trace at least one step, if possible.
    * If it is not possible to trace further, the same output is returned.
    * @param output the output to trace from.
+   * @param backEdgeState enum describing the path taken from the starting output to \p output.
    * @param withinRegion if not nullptr, tracing stops if it reaches an argument of the region.
    * @return the result of tracing from the given output, if possible. Otherwise, \p output.
```
phate: The documentation does not fully fit here any longer.

jlm/rvsdg/Trace.cpp
```diff
@@ -24,19 +24,47 @@ OutputTracer::trace(Output & output)
...
+    const auto traceStepResult = traceStep(*head, backEdgeState, withinRegion);
+
+    // If the tracing step is final, we are done
+    if (traceStepResult.isFinalResult())
+      return traceStepResult.getOutput();
+
+    // If the tracing step is not final, it must have made progress
+    JLM_ASSERT(&traceStepResult.getOutput() != head);
+
+    // Otherwise continue tracing
```
phate: The otherwise is wrong here.....I think you can just remove this comment

phate (dismissed):

haved:
Tended to your comments @phate

phate (approved):

